PR #66, opened 2011-10-10
Improvement on filters

snoack:
I have added the template filters, listed below and have done some improvements on existing filters.
- min
- max
- filter
- filterfalse
- map
- unique

You find further details including examples in the docstring of those filter functions.


snoack:
I have submitted new pull requests for adding the `unique` filter (#469) and the `min`/`max` filters (#475).

PR #84, opened 2012-01-05
Support included blocks override

paradoxxxzero:
This commit fix inheritance with included blocks.
Consider these templates:

**A.jinja2**

```
I am A
{% include 'I.jinja2' %}
```

**I.jinja2**

```
Who are you ?
{% block whoami %}
    I am I
{% endblock %}
```

**B.jinja2**

```
{% extends 'A.jinja2' %}
{% block whoami %}
    I am B
{% endblock %}
```

As of now, rendering **B.jinja2** would output:

```
I am A
Who are you ?
I am I
```

This patch allow B to override I block, the output will be instead:

```
I am A
Who are you ?
I am B
```

### DIFF
--- a/jinja2/compiler.py
+++ b/jinja2/compiler.py
@@ -950,9 +950,16 @@ class CodeGenerator(NodeVisitor):
             self.indent()
 
         if node.with_context:
+            self.writeline('include_context = template.new_context('
+                           'context.parent, True, locals())')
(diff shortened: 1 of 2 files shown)


SuryaSankar:
Will this feature be revived? Would be very happy to see this. The lack of this feature is the only issue I have had so far with Jinja2.

PR #229, opened 2013-05-25
Possible improvements to extension docs

plaes:
I tried making the extension docs a bit clearer.


davidism:
Thanks for the interest, but I don't feel that this changes the clarity of those sections.

PR #231, opened 2013-05-26
Clarified and extended the docs for the `{select|reject}[attr]` filters.

homeworkprod:


luzpaz:
bump

jeffwidman:
I'm not seeing a typo that needs fixing? If it's still there, can you rebase?

homeworkprod:
I've corrected four instances of "appying" to "applying".

However, that typo has been fixed by c251d8a, 21 months after my original PR.

My additional note (or something similar) to each of the four filters on how they behave if no test is given has not been added in the meantime, though.

Should I rebase?

jeffwidman:
> My additional note (or something similar) to each of the four filters on how they behave if no test is given 

Yeah, I think this is worth including. Rebase appreciated.

homeworkprod:
I have rebased. Please take a look.

I think the clarification of the docstring regarding the test being applied to *each* object of the sequence instead of "the object" prevents confusion, so I kept it.

PR #233, opened 2013-05-26
fix some stuff discovered by PyCharm

ThomasWaldmann:
also add more python classifiers to setup.py

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -147,14 +147,20 @@ work in production environments::
 Credit for this snippet goes to `Thomas Johansson
 <http://stackoverflow.com/questions/3086091/debug-jinja2-in-google-app-engine/3694434#3694434>`_
 
-Why is there no Python 2.3 support?
------------------------------------
+Why is there no Python 2.3/2.4/2.5/3.1/3.2 support?
+-------------------------------------------------------
 
 Python 2.3 is missing a lot of features that are used heavily in Jinja2.  This
 decision was made as with the upcoming Python 2.6 and 3.0 versions it becomes
 harder to maintain the code for older Python versions.  If you really need
 Python 2.3 support you either have to use `Jinja 1`_ or other templating
 engines that still support 2.3.
+
+Python 2.4/2.5/3.1/3.2 support was removed when we switched to supporting
+Python 2 and 3 by the same sourcecode (without using 2to3). It was required to
+drop support because only Python 2.6/2.7 and >=3.3 support byte and unicode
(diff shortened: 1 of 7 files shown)


ThomasWaldmann:
ping

docs/faq.rst
```diff
@@ -147,8 +147,8 @@ work in production environments::
 Credit for this snippet goes to `Thomas Johansson
 <http://stackoverflow.com/questions/3086091/debug-jinja2-in-google-app-engine/3694434#3694434>`_
 
-Why is there no Python 2.3 support?
------------------------------------
+Why is there no Python 2.3/2.4/2.5/3.1/3.2 support?
+-------------------------------------------------------
```
untitaker: Is this really necessary? How about "Why is there no support for Python X.X?"

docs/faq.rst
```diff
@@ -156,6 +156,12 @@ harder to maintain the code for older Python versions.  If you really need
 Python 2.3 support you either have to use `Jinja 1`_ or other templating
 engines that still support 2.3.
 
+Python 2.4/2.5/3.1/3.2 support was removed when we switched to supporting
+Python 2 and 3 by the same sourcecode (without using 2to3). It was required to
+drop support because only Python 2.6/2.7 and >=3.3 support byte and unicode
+literals in an easy and pretty way. If you really need support for older
```
untitaker: That phrasing is actually incorrect... every Python version has some form of byte and unicode literals, but in the versions supported by Jinja, the syntax overlaps.

untitaker:
I am typing this from mobile, so I don't really know why you e.g. removed some filters. Will take a closer look this evening.

jinja2/filters.py
```diff
@@ -952,8 +952,6 @@ def _select_or_reject(args, kwargs, modfunc, lookup_attr):
     'reverse':              do_reverse,
     'center':               do_center,
     'indent':               do_indent,
-    'title':                do_title,
-    'capitalize':           do_capitalize,
```
untitaker: I see, could you sort this dict to detect similar mistakes?

docs/faq.rst
ThomasWaldmann: ok, guess we need "in a common and pretty way".

jinja2/filters.py
ThomasWaldmann: how about just taking this PR as is and then doing minor additional fixes on your own?

untitaker: Right, ok.

untitaker:
I've applied the most important stuff to master.

PR #234, opened 2013-05-26
Fix "PrefixLoader does not pass globals to load"

MorrisJobke:
Fix mitsuhiko/jinja2#225


untitaker:
Fixed in master, apparently.

Issue #258, opened 2013-07-31
Command line client?


tebeka:
I wrote a little utility to expand Jinja templates from the command line - https://bitbucket.org/tebeka/pythonwise/src/tip/jj2.py?at=default.

Maybe we can incorporate it to Jinja? Either as a tool or as `__main__.py` module?

cp10000:
I agree that there should be a command-line interface to use Jinja2.   The command-line should be done in a way that exposes all the features (no loss of features).   Today (2015), there are more reasons than ever to have a command-line interface.

luzpaz:
:+1: for CLI

jeffwidman:
I'm not sure I see enough value here to add this to the main repo versus just letting you maintain it in your own repo. It feels like a helper script that won't get used a lot, and those who really want it are somewhat likely to google and stumble across your repo.

> Flask has several options to load data into a config file (from_envvar, from_object, ...). I think it's a good idea to move these functions to a new package and use them in a CLI tool to load template variables.

Maybe I'm misunderstanding, but didn't we reject a similar idea over on Flask? https://github.com/pallets/flask/issues/1706

kblomqvist:
I have made this kind of CLI tool, https://github.com/kblomqvist/yasha. I know there are others as well, but not as complete I think.

davidism:
#129 is closed as well.  I'd rather go the route that @mitsuhiko suggests and @kblomqvist took: build the tool you need separately.  There's so much in Jinja that any official cli would either be massive or would be underpowered.  Instead, focus on the exact use case you need and build it with Jinja and Click.

PR #291, opened 2014-01-10
Add jinja2 command line script

tebeka:
See [issue 258](https://github.com/mitsuhiko/jinja2/issues/258)

I decided to go with the simple approach for now - just command line variables.


jinja2/__main__.py
```diff
@@ -0,0 +1,58 @@
...
+    $
+'''
+
+
+from . import Environment, StrictUndefined
+
+
+def parse_value(value):
+    prefix = 'py:'
+    if value[:3] != prefix:
```
mallegonian: Use `if value.startswith(prefix):`?

mallegonian:
I'd also like to be able to read variables from the environment, for use like eRuby. I had written my own before thinking to search here: https://gist.github.com/mallegonian/09d7e005ddbc225852fc
This allows for easier usage from other scripts, eg:
```
export THING_A="beep"
export THING_B="boop"
python -m jinja2 /etc/braap.jinja >/etc/braap.conf
```

jinja2/__main__.py
```diff
@@ -0,0 +1,58 @@
...
+
+    args = parser.parse_args(argv[1:])
+
+    tvars = parse_cmdline_vars(args.var or [])
+
+    # Fail on undefined
+    env = Environment(undefined=StrictUndefined)
+    template = env.from_string(args.template.read())
+
+    args.output.write(template.render(tvars))
```
snoack: I think you should use `template.generate()` here, writing chunk by chunk to the file. This will improve performance and keeps memory usage low in case the output is huge.

jinja2/__main__.py
```diff
@@ -0,0 +1,58 @@
...
+        help='template variables (in X=Y format, prefix with py: to eval)')
+    parser.add_argument('--output', '-o', help='output file',
+                        type=FileType('w'), nargs='?', default=sys.stdout)
+
+    args = parser.parse_args(argv[1:])
+
+    tvars = parse_cmdline_vars(args.var or [])
+
+    # Fail on undefined
+    env = Environment(undefined=StrictUndefined)
```
snoack: Why do you enforce this non-standard behavior? If anything this should be an option.

setup.py
```diff
@@ -80,12 +84,15 @@
...
+    install_requires=install_requires,
     extras_require={'i18n': ['Babel>=0.8']},
     test_suite='jinja2.testsuite.suite',
     include_package_data=True,
     entry_points="""
     [babel.extractors]
     jinja2 = jinja2.ext:babel_extract[i18n]
+
+    [console_scripts]
+    jinja2 = jinja2.__main__:main
```
snoack: I think this should be a standalone script rather than a `__main__` module.

jinja2/testsuite/script.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def test_script(self):
+        for case in self.iter_cases():
+            expected = case['out'].strip()
+            out, rval = self.run_script(case['in'], case['args'])
+            assert rval == 0, 'script failed'
+            assert out.strip() == expected, 'bad output'
+
+
+def suite():
```
snoack: This should probably be integrated in the pytest suite.

auvipy:
@tebeka are you still working on it?

tebeka:
@auvipy Nope :)

davidism:
https://github.com/pallets/jinja/issues/258#issuecomment-239675504

PR #315, opened 2014-03-30
Fix loop context length calculation for iterators. Fixes #244

### LINKED ISSUE #244: Loop properties revindex and revindex0 not behaving correctly in Jinja2 2.7
In Jinja2 2.7 both `loop.revindex` and `loop.revindex0` give incorrect values.

`loop.revindex` should be 1-based, but instead is 0-based.
`loop.revindex0` should be 0-based, but instead is -1-based (it has value -1 in last loop iteration).

lazka:
This resulted in wrong values for revindex and revindex0
while looping over an iterator.

### DIFF
--- a/jinja2/runtime.py
+++ b/jinja2/runtime.py
@@ -339,10 +339,11 @@ class LoopContext(object):
             # if was not possible to get the length of the iterator when
             # the loop context was created (ie: iterating over a generator)
             # we have to convert the iterable into a sequence and use the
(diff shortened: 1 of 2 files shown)


serialine:
```
-            self._length = len(iterable) + self.index0 + 1
+            iterations_done = self.index0 + 2
```
Is +2 right?
has problem with `loop.index`, it returns +1

PR #343, opened 2014-06-15
Custom Filters example inserts escaped <br>s

joernhees:
The example nl2br custom filter inserts escaped `&lt;br&gt;` <br> tags as the `Markup.replace` method seems to pass `str`/`unicode` arguments through `escape`.


untitaker:
Good catch!

Issue #204, opened 2013-05-01
line_comment_prefix lines get rendered as blank lines


aspiers:
If I set

```
line_comment_prefix="##"
```

in my environment, and then include in my template

```
## start
## empty line follows:
## 
## end
```

Then it renders to 4 blank lines, which is clearly silly.  It should not render any of them.

mvolfik:
I'd say this could be the desired behavior. HOWEVER, they are also kept when `trim_blocks` and `lstrip_blocks` are true, which they probably shouldn't.

Interestingly though, they are correctly deleted when marked as `##- comment`, but the line is kept when there already were some statements on the line, so the logic remove-if-only-whitespace-before-else-keep as discussed in that PR is already there.

```python3
import jinja2

e = jinja2.Environment(line_comment_prefix="##")
e_trim = jinja2.Environment(
    line_comment_prefix="#", lstrip_blocks=True, trim_blocks=True
)

print("== Comment line is correctly kept")
print(e.from_string("markup\n## comment\nmarkup").render())

print("== Comment line is incorrectly kept")
print(e_trim.from_string("markup\n## comment\nmarkup").render())

print("== Comment line is correctly removed")
print(e.from_string("markup\n##- comment\nmarkup").render())

print("== Comment line is correctly kept")
print(e.from_string("markup\nmarkup ##- comment\nmarkup").render())
```

mvolfik:
I looked into how the `markup ##- comment` vs `##- comment` works: in `Lexer.tokeniter`, in `markup\n##- commment`, the regex match is `markup\n##-`. The `-` is detected, and the markup part is `rstrip`ped.

In the `markup\nmarkup ##- comment` case, the same is performed, but there is, obviously, no newline to `rstrip`.

This also leads to the following bug:

```python3
import jinja2
e = jinja2.Environment(line_comment_prefix="##")
print(e.from_string("markup\n\n\n\n\n\n##- comment\nmarkup").render() == "markup\nmarkup")
# True
```

I don't think there's a very simple solution to this.

davidism:
See https://github.com/pallets/jinja/issues/408#issuecomment-556501327 and https://github.com/pallets/jinja/pull/1109 (closed but relevant) for more newline issues.

mvolfik:
Yes, during debugging I noticed that jinja parses across multiple newlines at once. I don't think this is an issue per se, but it just needs to take care when stripping and decide when it should strip newlines and how many.

[...]

PR #346, opened 2014-07-03
Fix for #204 - line_comment_prefix

ieckert:
Fix for this issue: https://github.com/mitsuhiko/jinja2/issues/204

This fix cascaded into a few other tests when it comes to whitespace. 

For `test_line_syntax_priority`, since this line was in the code:

```
# XXX: why is the whitespace there in front of the newline?
```

I feel that it is alright for me to change that test.

I'm not sure about `test_lstrip_angle_bracket` and `test_lstrip_angle_bracket_compact`. On line 505:

```
${item} ## the rest of the stuff
```

It seems like there could be a whitespace before the comment, but I'm not sure.

Let me know if I should change anything!

### DIFF
--- a/jinja2/lexer.py
+++ b/jinja2/lexer.py
@@ -530,7 +530,7 @@ class Lexer(object):
             ] + tag_rules,
             # line comments
             TOKEN_LINECOMMENT_BEGIN: [
-                (c(r'(.*?)()(?=\n|$)'), (TOKEN_LINECOMMENT,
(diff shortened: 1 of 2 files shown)


job:
Can this be merged?

davidism:
Thanks for looking at this, sorry it got left for so long. This causes line comments to collapse multiple lines into one, as shown by the tests where you had to remove `\n` from the expected renders. I think what needs to happen instead is that a line comment should only consume whitespace if it's on a line by itself, so I'm going to close this in favor of figuring that out.

PR #354, opened 2014-07-23
Fix cycler documentation to match the behaviour

georgthegreat:
Rename next() method to __next__(), since it is the common behaviour.

AFAIK, there is no way to call python next() from jinja template.

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -1244,10 +1244,10 @@ The following functions are available in the global scope by default:
         {% set row_class = cycler('odd', 'even') %}
         <ul class="browser">
         {% for folder in folders %}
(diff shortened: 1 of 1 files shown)


jmagnusson:
I get the following in py3.4 when I use `.next()` on the cycler object. I didn't have it in py2.7.

`UndefinedError: 'jinja2.utils.Cycler object' has no attribute 'next'`

Are you having this problem as well?

Changing to `.__next__()` fixes it.

untitaker:
This is the wrong fix for a valid problem. Changing it to `__next__` is ugly I think.

davidism:
Fixed in #525

PR #356, opened 2014-07-31
Update utils.py

ankitjain87:
HTML escaping target would be nice?


untitaker:
Probably... Could you add tests?

PR #375, opened 2014-09-28
Windows returns EINVAL on bogus filenames

zart:
Related pull of pyramid_jinja2: https://github.com/Pylons/pyramid_jinja2/pull/94


untitaker:
:+1:

PR #387, opened 2014-12-10
Dict: Mapping iterable object

dkrnl:
usage example:
{{ dict(("a", 1), ("b", 2)) }} --> {'a': 1, 'b': 2}


ThiefMaster:
I think it woud make much more sense to expose the `dict` constructor itself.

untitaker:
Already fixed by only using `dict`. Thanks though!

PR #388, opened 2014-12-11
Fix incorrect truncate filter docs

olegpesok:
Examples contrary to behavior and description of killwords argument.


untitaker:
Thanks!

PR #395, opened 2015-01-08
Updated template documentation examples

esonderegger:
Fixed html tags in base template example. Since I was removing the
xmlns="http://www.w3.org/1999/xhtml" tag, it seemed like a good time to
update the examples to the html5 doctype.

(in response to issue: https://github.com/mitsuhiko/jinja2/issues/394)


untitaker:
Since you explicitly asked for feedback: Your PR seems fine to me :+1:

esonderegger:
Thanks!

untitaker:
Thanks again!

PR #400, opened 2015-01-23
Replace lambda for 'dict' in with dict itself

scavpy:
lambda **kw: kw is not equivalent to the dict constructor. It is much less useful.
In particular it doesn't accept a sequence of pairs.
Why not put dict itself into the DEFAULT_NAMESPACE?
Principle of least surprise, etc.


untitaker:
Seems sensible. Thanks!

PR #402, opened 2015-02-05
Change attr filter example to match description

quodlibetor:
`foo["bar"]` is an item lookup, so saying that items are not looked
up with that as an example is confusing.


untitaker:
Good catch, thanks!

PR #403, opened 2015-02-05
trim obsolete sentence re: i18n extension

whit537:
This sentence seems like a leftover from when i18n was the only extension available, eh?


untitaker:
Yup, thanks!

whit537:
:)

PR #428, opened 2015-03-22
Change py2 print statements to py3 in example test_ files

lkarthee:
Travis builds fail while running in py3 env because the print statements are in py2 style.


untitaker:
I applied your commit as https://github.com/mitsuhiko/jinja2/commit/29b08c85a206199094070205383e44e690ec44cf, though Travis doesn't run those anymore (it doesn't make sense, they just print to stdout).

lkarthee:
thank you

PR #429, opened 2015-03-23
firstof function

axil:
To simplify transition of django templates to jinja2:

{% firstof obj1 obj2 obj3 %} -> {{ firstof(obj1, obj2, obj3) }}

instead of currently working

{{ [obj1, obj2, obj3]|select|first }}

Or at least maybe mention this in the documentation somewhere for it doesn't look obvious at first glance.

PS Yes, I've seen https://github.com/mitsuhiko/jinja2/pull/42

### DIFF
--- a/jinja2/defaults.py
+++ b/jinja2/defaults.py
@@ -9,7 +9,7 @@
     :license: BSD, see LICENSE for more details.
 """
 from jinja2._compat import range_type
-from jinja2.utils import generate_lorem_ipsum, Cycler, Joiner
+from jinja2.utils import generate_lorem_ipsum, Cycler, Joiner, firstof
 
 
 # defaults for the parser / lexer
@@ -35,7 +35,8 @@ DEFAULT_NAMESPACE = {
     'dict':         dict,
     'lipsum':       generate_lorem_ipsum,
     'cycler':       Cycler,
-    'joiner':       Joiner
+    'joiner':       Joiner,
+    'firstof':      firstof,
 }
 
 
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -666,8 +666,10 @@ def do_round(value, precision=0, method='common'):
 
 
 @environmentfilter
-def do_groupby(environment, value, attribute):
+def do_groupby(environment, value, attribute, sort=True):
     """Group a sequence of objects by a common attribute.
+
+    Use `sort=False` if the objects are already sorted accordingly.
 
     If you for example have a list of dicts or objects that represent persons
     with `gender`, `first_name` and `last_name` attributes and you want to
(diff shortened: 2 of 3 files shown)


untitaker:
>PS Yes, I've seen #42

And what is your response to the comment there?

axil:
My response is that {{ firstof(obj1, obj2, obj3) }} is more readable and straightforward than {{ [obj1, obj2, obj3]|select|first }}.

In my opinion, templates are just about that. Otherwise we would still use c with sprintf and %s notation :)

It's completely upto the author to decide. If he still considers it unnecessary, I'd suggest at least creating somewhere in the docs a table for translation common django tags into the jinja tags.

snoack:
I still disagree. It's anything but obvious that `firstof(..)` will return the first argument that evaluates to `True`. Therefore `[..]|select|first` is actually more readable IMO, since readability isn't about the length of code, but about whether it's understandable what it does. Beside being more in line with best practices (and in particular the UNIX philosophy) of having each piece of code doing only one thing.

And frankly, I cannot think of many use cases where a `firstof` function would be useful anyway. If you are migrating from django chances are that you used `{% firstof %}` to work around django's limitation, while a different data structure would actually make more sense anyway.

I understand that you don't want, and probably shouldn't, change the logic and data structures while migrating your templates to jinja2. But it doesn't matter whether to replace django's `{% firstof %}` tag with `{{ [..]|select|first }}` or `{{ firstof(..) }}` as long as you have to change the code anyway. And when you prefer the "django way", I wonder why you switch to jinja2 in the first place.

untitaker:
I agree with Sebastian. A migration like this is reasonably easy to do with sed, or you register this filter yourself.

On 4 August 2015 20:15:59 CEST, Sebastian Noack <notifications@github.com> wrote:
>I still disagree. It's anything but obvious that `firstof(..)` will
>return the first argument that evaluates to `True`. Therefore
>`[..]|select|first` is actually more readable IMO, since readability
>isn't about the length of code, but about whether it's understandable
>what it does. Beside being more in line with best practices (and in
>particular the UNIX philosophy) of having each piece of code doing only
>one thing.
>
>And frankly, I cannot think of many use cases where a `firstof`
>function would be useful anyway. If you are migrating from django
>chances are that you used `{% firstof %}` to work around django's
>limitation, while a different data structure would actually make more
>sense anyway.
>
>I understand that you don't want, and probably shouldn't, change the
>logic and data structures while migrating your templates to jinja2. But
>it doesn't matter whether to replace django's `{% firstof %}` tag with
>`{{ [..]|select|first }}` or `{{ firstof(..) }}` as long as you have to
>change the code anyway. And when you prefer the "django way", I wonder
>why you switch to jinja2 in the first place.
>
>---
>https://github.com/mitsuhiko/jinja2/pull/429#issuecomment-127697774

-- 
Sent from my phone. Please excuse my brevity.

StriveForBest:
Having `firstof` in jinja2 would be extremely helpful 👍

ThiefMaster:
doesn't `{{ foo or bar or baz }}` do what you want?

StriveForBest:
`{{ foo.bar.baz }}` raises `AttributeError` if `foo` doesn't have `bar` or `bar.baz` attribute @ThiefMaster

ThiefMaster:
But that's not what this suggestion is about. It's about picking the first non-falsy value, not going along a chain of attributes.

StriveForBest:
@ThiefMaster  you are correct, this is probably a separate request. Django's `firstof` has chained getattr with fallback built-in and I, personally, find it very handy.

PR #430, opened 2015-03-24
unsorted groupby

axil:
Django groupby tag cannot currently be translated to jinja2.

This commit makes it possible without breaking backwards compatilibity.

In Django ordering is done by `group._meta.ordering` field(s). Currently performed `sorting` by `group` does no sorting at all since `django.db.Model` doesn't have `__lt__` method.

```python
class Group(object):
    def __init__(self, v):
        self.v = v

#    def __lt__(self, other):
#        return self.v < other.v

class A(object):
    def __init__(self, n, v):
        self.name = n
        self.group = Group(v)
    
    def __repr__(self):
        return 'A(%s, %d)' % (self.name, self.group.v)


q = (
    A('a', 5),
    A('b', 2),
    A('c', 5),
    A('d', 1),
    A('e', 2)
)

print q
print sorted(q, key=lambda x:x.group)
```

before uncommenting `__lt__()`:
(A(a, 5), A(b, 2), A(c, 5), A(d, 1), A(e, 2))
[A(a, 5), A(b, 2), A(c, 5), A(d, 1), A(e, 2)]

after uncommenting `__lt__()`:
(A(a, 5), A(b, 2), A(c, 5), A(d, 1), A(e, 2))
[A(d, 1), A(b, 2), A(e, 2), A(a, 5), A(c, 5)]

`{{ q|groupby('group') }}` wont sort at all
`{{ q|groupby('group.v') }}` will sort but returns senseless group.v as a grouper
`{{ q|sort('group.v')|groupby('group', sort=False) }}` works as expected

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -666,8 +666,10 @@ def do_round(value, precision=0, method='common'):
 
 
 @environmentfilter
-def do_groupby(environment, value, attribute):
+def do_groupby(environment, value, attribute, sort=True):
     """Group a sequence of objects by a common attribute.
+
+    Use `sort=False` if the objects are already sorted accordingly.
 
     If you for example have a list of dicts or objects that represent persons
     with `gender`, `first_name` and `last_name` attributes and you want to
@@ -705,7 +707,11 @@ def do_groupby(environment, value, attribute):
        attribute of another attribute.
     """
     expr = make_attrgetter(environment, attribute)
-    return sorted(map(_GroupTuple, groupby(sorted(value, key=expr), expr)))
+    if sort:
+        return sorted(map(_GroupTuple, groupby(sorted(value, key=expr), expr)))
+    else:
+        return map(_GroupTuple, groupby(value, expr))
+
 
 
 class _GroupTuple(tuple):


untitaker:
FWIW you can add any Python function to the template globals. This new parameter seems harmless IMO, but changing the current behavior has been historically rejected (IKR, this one doesn't change the defaults)

axil:
Another agrument is that both `itertools.groupby` and django's `regroup` keep ordering intact, so this issue is not actual for them.

snoack:
Please note that calling `itertools.groupby()` on an unordered sequence will only group adjacent items with the same key. I couldn't come up with a use case where this behavior would be intended/useful. AFAIK, the only reason Python's `itertools` module doesn't automatically sort the input, is performance when grouping huge lists that are already ordered. I don't think this is an issue in templates usually. On the other hand only grouping adjacent items, would be rather confusing for many template authors. I know, this PR wouldn't change the default behavior, but it might be still a confusing and rather useless feature IMO, unless I missed the point.

jinja2/filters.py
```diff
@@ -705,7 +707,11 @@ def do_groupby(environment, value, attribute):
        attribute of another attribute.
     """
     expr = make_attrgetter(environment, attribute)
-    return sorted(map(_GroupTuple, groupby(sorted(value, key=expr), expr)))
+    if sort:
+        return sorted(map(_GroupTuple, groupby(sorted(value, key=expr), expr)))
```
d1ffuz0r: not sure about memory consumption. Will `sorted` every time create a copy of value?

snoack: Yes, `sorted()` creates a new list, the items aren't copied however but passed by reference. But indeed the outer sort seems to be unneeded. Anyway, that logic doesn't change with this pull request. I will file a new pull request to get rid of the redundant sort.

jeffwidman:
https://github.com/pallets/jinja/pull/430#issuecomment-127755500 makes sense to me.

PR #433, opened 2015-03-26
Fix capturing references to external objects

beloborodov:
_tb was not deleted in except block left in module scope and captured references to external objects by holding on _tb.tb_frame.


untitaker:
Fixed by moving the whole type finding into a temporary function.

PR #439, opened 2015-04-07
Make title filter capitalize words inside parens and other brackets

jgmize:


jgmize:
Updated to simplify the code.

davidism:
What about the other brackets? `{ [ <`

jgmize:
> What about the other brackets? { [ <

Good point, @davidism. I've updated the PR to handle those brackets as well, along with some performance optimizations.

jeffwidman:
Thanks!

PR #440, opened 2015-04-08
clean up do_int function

kelvintaywl:
This PR shortens the `do_int` function to its essential.


untitaker:
I think this is a performance optimization.

kelvintaywl:
@untitaker i'm not sure how the reason given meant that this PR should be closed. but... no problem

untitaker:
I think the original code tried to avoid a `float` call by first trying a simple `int(value)`.

PR #442, opened 2015-04-21
Add more information to preprocessor, parser and lexer in babel extraction

jespino:
The babel extractor don't pass the filename information to the preprocessor, parser and lexer, this is an error because some extensions use the filename to decide to preprocess or not the code (like pyjade).

This change (IMHO) doesn't break anything, and allow to integrate babel with jinja + extensions easily.


jeffwidman:
LGTM

jespino:
After almost 3 years, and after almost 2 years without feedback, I close the PR.

PR #443, opened 2015-04-24
Return only unique values in the output of the sort filter

steelman:
Alternatively a separate `uniq` filter may be developed.

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -224,7 +224,7 @@ def do_dictsort(value, case_sensitive=False, by='key'):
 
 @environmentfilter
 def do_sort(environment, value, reverse=False, case_sensitive=False,
-            attribute=None):
+            attribute=None, uniq=False):
     """Sort an iterable.  Per default it sorts ascending, if you pass it
     true as first argument it will reverse the sorting.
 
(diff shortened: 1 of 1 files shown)


untitaker:
* This should probably document that the items must be hashable.
* I'm not sure there's a good enough usecase for this. What about `{{ set(x)|sort }}`?
* It'd probably make sense to implement this as separate filter.

steelman:
1. Indeed, I didn't check the [details](https://docs.python.org/2/library/stdtypes.html#set-types-set-frozenset).

2. I haven't seen this anywhere in the [documentation](http://jinja.pocoo.org/docs/dev/templates/) and I am new to Jinja enough not to know which parts of it are common with Python.

3. I thought about it. However, I prefered sorted output anyway so I came up with something simmilar to `sort -u`.

untitaker:
Re 2: I'm not so sure anymore if that builtin is actually available in Jinja. I think it's not.
Re 3: On the CLI there's also `uniq` available.

FWIW you can add your own filters and globals to Jinja, and since Python doesn't have a comparable function, I am not sure if this might not be out of scope -- even if only slightly. You're the first one to request this, so I'd like to just keep this open and see how many people `+1` this.

snoack:
I'd agree that a built-in way to eliminate duplicates from a sequence would be desirable. In fact, I ended up implementing a `unique` filter myself for a couple of projects, and already asked to have it added to upstream jinja2, 4 years ago as part of #66, see snoack/jinja2@2d9a896. But I think it should be a separate filter for larger flexibility. So one can eliminate duplicates from a sequence while preserving order.

snoack:
I have just rebased and improved my old patch, and submitted a new pull request, see #469.

jeffwidman:
I'm a fan of adding a filter like `unique` or `set`, but like @untitaker I'd rather see it as a separate filter rather than as a param on `sort`.

PR #446, opened 2015-04-29
Add the "base" parameter when calling the int filter

bentimms:
The default is base 10.


bentimms:
Should I close and resubmit this pull request?

PR #450, opened 2015-05-09
FIX DOC-EXAMPLE: cache_extension.py

jenisys:
No longer working with current code base since the TokenStream.next() method was removed.


untitaker:
Thanks!

PR #454, opened 2015-05-17
FIX: Filter title, capitalize is registered twice in FILTERS

jenisys:
Both filters appear twice in FILTERS registration (second time in reverse order).
That was probably a merge-problem.


untitaker:
Thanks!

PR #455, opened 2015-05-29
Fix typo

merwok:
Probably a mistake caused by “bekommen” :‑)

### DIFF
--- a/jinja2/loaders.py
+++ b/jinja2/loaders.py
@@ -291,7 +291,7 @@ class DictLoader(BaseLoader):
 
 class FunctionLoader(BaseLoader):
     """A loader that is passed a function which does the loading.  The
-    function receives the name of the template passed and has to return either
+    function receives the name of the template and has to return either
     an unicode string with the template source, a tuple in the form ``(source,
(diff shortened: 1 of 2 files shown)


jinja2/loaders.py
```diff
@@ -291,7 +291,7 @@ def list_templates(self):
 
 class FunctionLoader(BaseLoader):
     """A loader that is passed a function which does the loading.  The
-    function becomes the name of the template passed and has to return either
+    function receives the name of the template passed and has to return either
```
untitaker: "recieves... passed" still sounds kind of weird.

merwok: Right.  Fixed.

untitaker:
Thanks!

merwok:
Thanks!  I see “This pull request is closed, but the merwok:patch-1 branch has unmerged commits.”  I guess I can ignore that?  I see both of my changes in your commit.

untitaker:
I've melded your commits into a single one under your name, so the ones in your branch are technically unmerged. You can ignore that msg by github

PR #456, opened 2015-06-11
Adding tojson to the built-in filters

inhumantsar:
Test included. Sorry about the newline in README.rst.

There was a concern that this might not be appropriate for the built-ins. Obviously, I'd argue that it is. Many major projects outside Pocoo (Ansible and Django for example) have written their own filters to do the same job. In my case, our company uses Jinja as a standalone library for our automation tooling and don't get the extras included with applications/frameworks that build on top of Jinja. Since so many devs are doing the same work in their downstream projects to make up for the lack in Jinja, it seems appropriate to simply provide that functionality here.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,3 +47,4 @@ Builds
 | ``2.7-maintenance`` | .. image:: https://travis-ci.org/mitsuhiko/jinja2.svg?branch=2.7-maintenance |
 |                     |     :target: https://travis-ci.org/mitsuhiko/jinja2                          |
 +---------------------+------------------------------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


snoack:
Sorry, I wasn't aware of this pull request when I created #468, also see my [comment here](https://github.com/mitsuhiko/jinja2/issues/458#issuecomment-127684750).

I called the filter simply `json` however. This better fits the naming convention of filters like `int` and `list`. Also I pass additionally keywords arguments through to `json.dumps`.

jeffwidman:
Closing in favor of #468

Issue #466, opened 2015-07-30
Regression: int filter always returns 0 in 2.8, returned result of __int__() in 2.7.3


gsmethells:
I am seeing a regression in my test suite since Jinja2 2.8 came out. I have in my template:

id="scope-stream-{{stream|int}}"

the stream instance has a ```__int__()``` dunder method. 

In Jinja 2.7.3 my template renders with:

id="scope-stream-2"

In Jinja 2.8 my template renders with:

id="scope-stream-0"

snoack:
Reproduced ..and fixed. That was because `int()` always fails with a `TypeError` when an explicit base is specified, but the value isn't a string. Hence this regression where introduced by adding the optional `base` argument to the `int` filter.

PR #467, opened 2015-08-04
Fixed int() filter for non-string objects #466

snoack:
See #466

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -518,9 +518,12 @@ def do_int(value, default=0, base=10):
     can also override the default base (10) in the second
     parameter, which handles input with prefixes such as
     0b, 0o and 0x for bases 2, 8 and 16 respectively.
+    The base is ignored for decimal numbers and non-string values.
     """
     try:
-        return int(value, base)
+        if isinstance(value, string_types):
+            return int(value, base)
+        return int(value)
     except (TypeError, ValueError):
         # this quirk is necessary so that "42.23"|int gives 42.
         try:
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -133,11 +133,16 @@ class TestFilter():
                        'foo bar foo bar\n  foo bar foo bar')
 
     def test_int(self, env):
+        class IntIsh(object):
+            def __int__(self):
+                return 42
+
         tmpl = env.from_string('{{ "42"|int }}|{{ "ajsghasjgd"|int }}|'
                                '{{ "32.32"|int }}|{{ "0x4d32"|int(0, 16) }}|'
-                               '{{ "011"|int(0, 8)}}|{{ "0x33FU"|int(0, 16) }}')
-        out = tmpl.render()
-        assert out == '42|0|32|19762|9|0'
+                               '{{ "011"|int(0, 8)}}|{{ "0x33FU"|int(0, 16) }}|'
+                               '{{ obj|int }}')
+        out = tmpl.render(obj=IntIsh())
+        assert out == '42|0|32|19762|9|0|42'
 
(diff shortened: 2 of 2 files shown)


jinja2/filters.py
```diff
@@ -518,9 +518,17 @@ def do_int(value, default=0, base=10):
...
+        try:
+            return int(value, base)
+        except ValueError:
+            return default
+
     try:
-        return int(value, base)
+        return int(value)
     except (TypeError, ValueError):
         # this quirk is necessary so that "42.23"|int gives 42.
```
untitaker: With your fix, this quirk may get skipped when required. I prefer a solution where logic is not duplicated.

jinja2/filters.py
```diff
@@ -518,9 +518,17 @@ def do_int(value, default=0, base=10):
...
     parameter, which handles input with prefixes such as
     0b, 0o and 0x for bases 2, 8 and 16 respectively.
+    The base is ignored for non-string values.
     """
+    # only string types support int() with explicit base
+    if isinstance(value, string_types) and base != 10:
+        try:
+            return int(value, base)
+        except ValueError:
+            return default
```
untitaker: `{{ "42.3"|int(0, 16) }}` -- this used to return `42`.

snoack: I realized this, and considered it a bug as well. Treating strings with representing a floating point number as decimal, when hexadecimal was specified, while respecting the base when the string represents an integer, is certainly wrong.

You could either say that floating point semantics aren't specified for non-decimal numbers. Then my code would behave correctly and as expected by returning the default.

Or one could insist that floating points should be supported consistently for all bases. This could be implemented like below. Let me know what you think.

```python
try:
        if isinstance(value, string_types):
            return [int(x, base) for x in value.split('.', 1)][0]
        else:
            try:
                return int(value)
            except (TypeError, ValueError):
                return int(float(value))
    except (TypeError, ValueError):
        return default
```

untitaker: I kind of understand that rationale but changing the API is inappropriate for a bugfix, so I'd like to keep the previous behavior (`{{ "42.3" | int(0, 16) }}` is `42`). Feel free to open an issue about this though.

untitaker: I'm not opposed to changing behavior, but it should happen in a noticeable way (unlike the regression itself), e.g. raise an exception if a base is given and the value is unparseable with `int`.

snoack: On first though this is inconsistent, as we return the `default` in all other scenarios where the input cannot be parsed. But on the other hand there doesn't seem to be one particular expected way to handle explicit bases in these corner cases. So I'm fine with not catching the exception there. I've updated the PR.

untitaker:
I need to think about this more, I'm still not sure what the API should look like. Perhaps @mitsuhiko should decide on this.

snoack:
I updated the pull request, to only pass the base to `int()` if we have string, ignoring the base otherwise. As opposed to my initial patch, I use the fallback, casting the value to `float` and that result to `int`, regardless of the base, now. Sot his only fixes the regression, without changing any other behavior.

The inconsistency when parsing floating points with non-decimal bases, isn't probably too important. On second thought, string representations of floating point numbers are also called decimal numbers, implying a base of 10. So ignoring the base for these would be similar to ignoring the base for non-string values. So I simply documented this behavior respectively, to avoid confusion. If we still want to change that behavior, we can do so later with a separate patch.

jeffwidman:
LGTM

untitaker:
Release is blocked on https://github.com/pallets/meta/issues/8, feel free to ping me about additional things in Flask/Jinja/Werkzeug that need attention.

Issue #458, opened 2015-06-28
add filter `tojson`


schettino72:
It would be very useful to add builtin filter to convert values to json.

Actually there is already an implementation in flask http://flask.pocoo.org/docs/0.10/templating/#standard-filters

If you guys agree I could work on a pull-request for that.

tima:
Ansible also implements a to_json and a to_nice_json (pretty JSON) filter to Jinja2. See around line 64 here: https://github.com/ansible/ansible/blob/e07dde1a3ce88f4ba21f779f4c2cb2d78397f34b/lib/ansible/plugins/filter/core.py#L64

snoack:
My pull request adds a filter called `json` which effectively just calls `json.dumps()`. I think it make sense to have this in upstream jinja2, as every other project using jinja2 I've seen adds a filter like this.

I went for `json`rather than `tojson`, to match the naming convention of other filters like `list` and `int`.

Additional, keyword arguments are passed through to `json.dumps()`. However, I made `ensure_ascii` default to `False`. So pretty printing is as simple as `{{ obj|json(indent=4) }}`

thatarchguy:
+1 pretty print

kiddick:
Bump
Why this is still open?

moigagoo:
What's missing to merge the PR?

Bedotech:
+1

jeffwidman:
@snoack has merge rights, not sure if he's holding off for some reason...

PR #468, opened 2015-08-04
Added json filter #458

snoack:
See #458

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -914,6 +914,14 @@ def do_rejectattr(*args, **kwargs):
     """
     return _select_or_reject(args, kwargs, lambda x: not x, True)
 
+def do_json(value, **kwargs):
+    """Serializes an object as JSON. Optionally given keyword arguments
+    are passed to json.dumps(), ensure_ascii however defaults to False.
+    """
+    import json
+    kwargs.setdefault('ensure_ascii', False)
+    return json.dumps(value, **kwargs)
+
 
 def _select_or_reject(args, kwargs, modfunc, lookup_attr):
     context = args[0]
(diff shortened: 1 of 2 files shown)


jeffwidman:
LGTM. While I'm not sure this is useful for Flask (I'd suggest `marshmallow` instead), I do think it makes sense for the Ansible project and probably a few others. 

Thanks for adding tests. Be good to add a note to the docs (including an example of how to pretty print the json) as well as the changelog. However, might wait for another maintainer or two to chime in whether they're on board with adding this filter. They may have objections that I haven't thought about.

sciyoshi:
:+1: for the feature. One note is that any built-in JSON filter should properly handle escaping characters like `<`, `>`, and `&` so that it can safely be used inside `<script>` tags. (See Flask's code here: https://github.com/pallets/flask/blob/master/flask/json.py#L161)

jinja2/filters.py
```diff
@@ -914,6 +914,14 @@ def do_rejectattr(*args, **kwargs):
     """
     return _select_or_reject(args, kwargs, lambda x: not x, True)
 
+def do_json(value, **kwargs):
+    """Serializes an object as JSON. Optionally given keyword arguments
+    are passed to json.dumps(), ensure_ascii however defaults to False.
+    """
+    import json
```
ThiefMaster: Not a big fan of this. Now you have a `json` filter that always uses the stdlib `json` module with the default settings and absolutely no way (except defining your own filter) to override this.

ThiefMaster:
Besides `json` vs `tojson` (the latter is already used e.g. in Flask) I think it would also make this an `environmentfilter` and allow the environment to specify another json encoder.

jinja2/filters.py
```diff
@@ -914,6 +914,14 @@ def do_rejectattr(*args, **kwargs):
...
+    kwargs.setdefault('ensure_ascii', False)
```
ThiefMaster: What's the advantage of disabling `ensure_ascii`? In Flask it defaults to `True`...

PR #469, opened 2015-08-04
Added unique filter

snoack:
4 years ago I suggested to add a `unique` filter, along with some other new filters as part of #66. That pull request has been closed, after some parts of it got implemented independently. However, I just saw a similar request (#443) for excluding duplicates using filters. So I rebased my old patch, added tests and did some logical improvements, that it hopefully get merged now.

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -51,21 +51,24 @@ def environmentfilter(f):
     return f
 
 
-def make_attrgetter(environment, attribute):
+def make_attrgetter(environment, attribute, lowercase=False):
     """Returns a callable that looks up the given attribute from a
     passed object with the rules of the environment.  Dots are allowed
     to access attributes of attributes.  Integer parts in paths are
     looked up as integers.
     """
-    if not isinstance(attribute, string_types) \
-       or ('.' not in attribute and not attribute.isdigit()):
-        return lambda x: environment.getitem(x, attribute)
-    attribute = attribute.split('.')
+    if attribute is None:
+        attribute = []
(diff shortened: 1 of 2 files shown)


jettero:
why is this not getting merged?

jinja2/filters.py
```diff
@@ -51,21 +51,24 @@ def environmentfilter(f):
     return f
 
 
-def make_attrgetter(environment, attribute):
+def make_attrgetter(environment, attribute, lowercase=False):
```
ThiefMaster: Does this really fit in this function? I'd rather have the option to pass an arbitrary `postprocess` callable that'll be applied before returning the attribute's value.

davidism: `sort` and `dictsort` use the name `case_sensitive`.

jinja2/filters.py
```diff
@@ -251,18 +254,51 @@ def do_sort(environment, value, reverse=False, case_sensitive=False,
...
+    getter = make_attrgetter(environment, attribute, not case_sensitive)
+
+    seen = set()
+    rv = []
+
+    for item in value:
+        key = getter(item)
+        if key not in seen:
+            seen.add(key)
+            rv.append(item)
```
davidism: This should `yield item` instead of building the full list ahead of time.

davidism:
Needs a changelog entry.

Talkless:
@snoack Are you yet interested in fixing this PR?

davidism:
Continued in #735

Issue #298, opened 2014-02-10
Overriding blueprint template does not take effect until restart


adrianlopezroche:
Overriding a blueprint's template does not take effect without restarting the development server.

Follow these steps to reproduce:
- First, add and register the following blueprint without creating any templates. It should produce a "Template Not Found" error:

```
from flask import Blueprint, render_template

blueprint = Blueprint(
    'blueprint', 
    __name__, 
    template_folder='blueprint/templates')

@blueprint.route('/testoverride')
def TestOverride():
    return render_template('test.html')
```
- Next, add a new template called "test.html" under the blueprint's template directory and refresh the page. It should find the template without restarting the server. 
- Now add a second template having the same filename but different content under the application's template directory and refresh the page. The page's content will not change without restarting the server. 
- Restart the server and refresh the page. Confirm the content has now changed.

PR #470, opened 2015-08-05
Use partial(next, ...) instead get_next()

snoack:
Since jinja2 doesn't support any version older than Python 2.6 anymore, we can simply use the `next` built-in function rather than abstracting `it.__next__` (Python 3) and `it.next` (Python 2) ourselves.

### DIFF
--- a/jinja2/environment.py
+++ b/jinja2/environment.py
@@ -10,6 +10,7 @@
 """
 import os
 import sys
+from functools import reduce, partial
 from jinja2 import nodes
 from jinja2.defaults import BLOCK_START_STRING, \
      BLOCK_END_STRING, VARIABLE_START_STRING, VARIABLE_END_STRING, \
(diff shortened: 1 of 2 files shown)


jinja2/environment.py
```diff
@@ -10,6 +10,7 @@
 """
 import os
 import sys
+from functools import reduce, partial
```
snoack: Note that I moved this line while changing it above the `from jinja2 ..` imports for compliance with PEP-8, where corelib imports should go first.

jeffwidman:
LGTM

snoack:
Since I'm not a committer, and you seem to approve this change, mind merging it? (Same for my other PRs where you gave LGTM)

jeffwidman:
I'm the least-experienced maintainer on the team, so if I review a PR and think it looks good but that there's a chance I might be unaware of why something is the way it is, I just give a LGTM.

So far that's worked pretty well, sometimes it gets merged directly by one of the other maintainers, sometimes someone chimes in with more background on why something is a particular way. But generally those PRs get a second look. 

BTW, I know many of these PRs/issues just sat for a while, but we're making consistent progress on the backlog now that we moved to `pallets`. If you're still interested in helping out on the project, we'd certainly appreciate your help tackling some of the thornier issues like #19 and #298

snoack:
Fine with me. I didn't meant to urge you. It just wondered why a maintainer gives LGTM, without merging the change or any further comment. Anyway, thanks for moving things forward here.

jeffwidman:
https://github.com/pallets/meta/issues/14 :wink:

jeffwidman:
I'm going to go ahead and merge this.

PR #471, opened 2015-08-05
Improve with_metaclass()

snoack:
With mitsuhiko/flask#1539 the `with_metaclass` function got impoved in Flask, see the comments there. So we should use the same implementation in jinja2 as well. Note that this PR merely copied over the code from Flask.

### DIFF
--- a/jinja2/_compat.py
+++ b/jinja2/_compat.py
@@ -86,23 +86,14 @@ else:
 
 
 def with_metaclass(meta, *bases):
+    """Create a base class with a metaclass."""
     # This requires a bit of explanation: the basic idea is to make a
-    # dummy metaclass for one level of class instanciation that replaces
-    # itself with the actual metaclass.  Because of internal type checks
-    # we also need to make sure that we downgrade the custom metaclass
-    # for one level to something closer to type (that's why __call__ and
-    # __init__ comes back from type etc.).
-    #
-    # This has the advantage over six.with_metaclass in that it does not
-    # introduce dummy classes into the final MRO.
-    class metaclass(meta):
-        __call__ = type.__call__
-        __init__ = type.__init__
+    # dummy metaclass for one level of class instantiation that replaces
+    # itself with the actual metaclass.
+    class metaclass(type):
         def __new__(cls, name, this_bases, d):
-            if this_bases is None:
-                return type.__new__(cls, name, (), d)
             return meta(name, bases, d)
-    return metaclass('temporary_class', None, {})
+    return type.__new__(metaclass, 'temporary_class', (), {})
 
 
 try:


davidism:
We should really have a common library for this stuff.  I know @untitaker has been talking about that with Werkzeug in the past.

untitaker:
We do, it's called six. I don't recall talking about moving compat into an external lib.

On 5 August 2015 18:40:31 CEST, David Lord <notifications@github.com> wrote:
>We should really have a common library for this stuff.  I know
>@untitaker has been talking about that with Werkzeug in the past.
>
>---
>https://github.com/mitsuhiko/jinja2/pull/471#issuecomment-128065797

-- 
Sent from my phone. Please excuse my brevity.

snoack:
Not that I'm a particular fan of six. But out of curiosity, why isn't
Werkzeug/Jinja2/Flask using six?

untitaker:
The current compat module is reasonably small to be vendored. We'd use very little of six, and afaik it does a lot of magical stuff.

On 5 August 2015 19:13:11 CEST, Sebastian Noack <notifications@github.com> wrote:
>Not that I'm a particular fan of six. But out of curiosity, why isn't
>Werkzeug/Jinja2/Flask using six?
>
>
>---
>https://github.com/mitsuhiko/jinja2/pull/471#issuecomment-128077158

-- 
Sent from my phone. Please excuse my brevity.

davidism:
@untitaker I didn't mean move to six, I just remembered you talking about extracting common stuff from Werkzeug, not specifically compat.  My point was that it's the same code in each project.

untitaker:
Yeah, but I didn't really mean compat, but contrib.

On 5 August 2015 19:20:09 CEST, David Lord <notifications@github.com> wrote:
>@untitaker I didn't mean move to six, I just remembered you talking
>about extracting common stuff from Werkzeug, not specifically compat. 
>My point was that it's the same code in each project.
>
>---
>https://github.com/mitsuhiko/jinja2/pull/471#issuecomment-128081932

-- 
Sent from my phone. Please excuse my brevity.

snoack:
Well, its actually not the same code except for some particular
implementations like metaclasses and a few trivial declarations. Most code
in the specific _compat modules however is tooled towards the project
specific needs. Which I'd agree is a reasonable approach.

However, if there is a dependency on six anyway, I wonder why we don't use
their with_metaclass function. On the other hand I'd personally would
rather get rid of six completely.

untitaker:
There is no dependency on six.

On 5 August 2015 19:39:52 CEST, Sebastian Noack <notifications@github.com> wrote:
>Well, its actually not the same code except for some particular
>implementations like metaclasses and a few trivial declarations. Most
>code
>in the specific _compat modules however is tooled towards the project
>specific needs. Which I'd agree is a reasonable approach.
>
>However, if there is a dependency on six anyway, I wonder why we don't
>use
>their with_metaclass function. On the other hand I'd personally would
>rather get rid of six completely.
>
>
>---
>https://github.com/mitsuhiko/jinja2/pull/471#issuecomment-128086577

-- 
Sent from my phone. Please excuse my brevity.

snoack:
Sorry, misread your comment above. Deleted my comment. I agree that having vendored `_compat` modules in the respective projects is a reasonable approach, given that they are fairly trivial and tooled towards the project-specific needs. So I suppose there is no objection merging this pull request?

untitaker:
Thanks!

PR #472, opened 2015-08-06
Remove conditions segment of Django section

pydanny:
* http://jinja.pocoo.org/docs/dev/switching/#conditions has been completely inaccurate since the Django 1.4 release, which was over 2 years ago. Reference: https://docs.djangoproject.com/en/1.4/ref/templates/builtins/#if
* Modern Django (1.8) has full support for `if`, `elif`, and `else` statements: https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#if

Just noticed this discrepancy again while integrating Jinja2 into djangopackages.com. :smile:

### DIFF
--- a/docs/switching.rst
+++ b/docs/switching.rst
@@ -116,37 +116,6 @@ Python syntax. Thus this Django code::
 This allows you to pass variables to the method, which is not possible in
 Django. This syntax is also used for macros.
 
-Conditions
-~~~~~~~~~~
-
-In Django you can use the following constructs to check for equality::
-
-    {% ifequal foo "bar" %}
-        ...
-    {% else %}
-        ...
-    {% endifequal %}
-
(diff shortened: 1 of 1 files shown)


untitaker:
I'm not really familiar with Django, but it seems that `ifequal` is still there. So perhaps it makes more sense to rewrite this section to point out that Jinja doesn't have it.

Phlip:
LGTM

pydanny:
@untitaker As far as I can see, no modern Django tutorial or reference uses `ifequal` anymore. No project created in the past two years have used it, and many older ones have converted to the cleaner, `jinja2`-style dialogue. It's only kept for backwards compatibility and it should arguably be removed from the template language. In fact, you inspired me to file https://code.djangoproject.com/ticket/25236 :wink:

untitaker:
Makes sense, thanks!

PR #473, opened 2015-08-07
Got rid of redundant sorting in groupby filter

snoack:
I realized that the `groupby` filter applies sorting on the grouped result. This however is unnecessary, since we already sort the input using the same expression we group by and `itertools.groupby` doesn't change the order (which is the reason you have to manually sort the input for in the first place).

So sorting the the result again doesn't have any effect, beside hurting performance and increasing memory usage, since it will effectively sort for the expression it has been sorted by before, as it's computed value is stored in the first item of the tuples in the sequence.

Note that I also had to replace `map` with a list comprehension to make sure that we still return a list in Python 3, where `map` returns a lazy-evaluated iterable. And since `_GroupTuple` doesn't have to take a single argument then anymore (and since we don't support Python versions older than 2.6 anymore), we can simply use `collections.namedtuple()` to define `_GroupTuple` now.

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -12,8 +12,8 @@ import re
 import math
 
 from random import choice
-from operator import itemgetter
 from itertools import groupby
+from collections import namedtuple
 from jinja2.utils import Markup, escape, pformat, urlize, soft_unicode, \
      unicode_urlencode
 from jinja2.runtime import Undefined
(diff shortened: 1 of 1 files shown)


jeffwidman:
LGTM. Actually it looks fantastic, you did a great job. 

It does touch just enough internal stuff that I'm not comfortable merging until another maintainer takes a look as well, just in case I missed something.

jeffwidman:
Took another look at it, and think it's fine to merge.

PR #475, opened 2015-08-07
Added min and max filters

snoack:
4 years ago I suggested to add `min` and `max` filters, along with some other new filters as part of #66. That pull request has been closed, after some parts of it got implemented independently, even though somebody already complained there about these filters having been left out.

Also I still think adding these particular filters would be useful, as I've seen quite some projects adding custom filters for these purpose. So it seems to be a quite common use case. Also having them built into jinja2, allows us to support the optional `attribute` and `case_sensitive` options/semantics used by other built-in filters.

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -13,7 +13,7 @@ import math
 
 from random import choice
 from operator import itemgetter
-from itertools import groupby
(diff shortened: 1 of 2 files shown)


davidism:
Continued in #737

PR #476, opened 2015-08-08
Fixed optimization of filters and misoptimizations with changed eval context

snoack:
I realized that some filters aren't inlined during optimizations. It turned out that this effects all environment and eval context filters. That is because the environment or eval context isn't passed as first but second argument. Therefore the filter function raises an exception and the optimizer skips the node. This one was easy to fix, see the first commit in this PR.

However, now things get funny. With that fix, the tests for the `{% autoescape %}` tag suddenly start to fail. As it turned out the optimizer were agnostic to the eval context, and the `as_const()` method created implicitly a new eval context for every node, ignoring `{% autoescape %}` tags up the tree. So I fixed this as well, see the second commit in this PR. Therefore optimizer now keeps track of the eval context, and updates it when visiting `ScopedEvalContextModifier` nodes. To avoid future issues like that and to get rid of dead code, I also made the eval context passed to `as_const()` mandatory.

### DIFF
--- a/jinja2/nodes.py
+++ b/jinja2/nodes.py
@@ -529,7 +529,7 @@ class Filter(Expr):
     fields = ('node', 'name', 'args', 'kwargs', 'dyn_args', 'dyn_kwargs')
 
     def as_const(self, eval_ctx):
-        if eval_ctx.volatile or self.node is None:
+        if self.node is None:
             raise Impossible()
         # we have to be careful here because we call filter_ below.
         # if this variable would be called filter, 2to3 would wrap the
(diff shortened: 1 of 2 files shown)


jeffwidman:
I suspect you haven't gotten much feedback simply because there's a lot going on here. I don't have the time to review it in depth either right now. 

Since none of these change external APIs (other than fixing bugs), then if you are confident that these are still good, I'd say go ahead and merge it.

davidism:
I'm having trouble merging and evaluating this. The first commit is already applied in master, and no tests fail.  While keeping the context during optimization seems correct, I don't know the original intention for not doing this. I cannot get `Optimizer.visit_EvalContextModifier` to trigger, so I don't know if this is needed. The final commit also does not appear to be necessary.

@snoack if you can submit a new PR that merges and has tests, that would be appreciated.

PR #478, opened 2015-08-10
Prevent random filter from being inlined

snoack:
Currently, due to #476 environment filters like `random` aren't optimized/inlined. While this is a bug for any other filter, the `random` filter should never be inlined, but always be evaluated at runtime. Otherwise, once the template is compiled, the `random` filter would always give the same results. So this pull request turns `random` into a `contextfilter` to prevent it from being inlined.

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -10,8 +10,8 @@
 """
 import re
 import math
+import random
 
-from random import choice
 from operator import itemgetter
 from itertools import groupby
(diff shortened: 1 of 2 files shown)


k4nar:
I've encountered this bug with Jinja 2.9.6 and this patch fixes it.

If this isn't merged, a workaround is to simply override `env.filters['random']` with the `do_random` function from this patch.

davidism:
New test doesn't fail.

davidism:
Continued in #734.

PR #479, opened 2015-08-10
Inline template tests during optimization

snoack:
This pull requests implements the `as_const()` method for `Test` nodes. That way tests get inlined along with filters and function calls by the optimizer. To avoid/remove code duplication, I moved the common code, parsing the arguments, into a helper function.


davidism:
Continued in #740

PR #492, opened 2015-09-17
Handle python -OO

DentonGentry:
With optimization -OO docstrings are omitted and will be
None. Only append to the docstring if it exists.


davidism:
Yeah, looks like this should already be caught by `__debug__`.

PR #493, opened 2015-09-18
Fix next() method in Cycler class for Python 3

tark-hidden:
issue https://github.com/mitsuhiko/jinja2/issues/474

### DIFF
--- a/jinja2/utils.py
+++ b/jinja2/utils.py
@@ -494,7 +494,7 @@ class Cycler(object):
         if not items:
             raise RuntimeError('at least one item has to be provided')
         self.items = items
-        self.reset()        
+        self.reset()
 
     def reset(self):
         """Resets the cycle."""
--- a/jinja2/utils.py
+++ b/jinja2/utils.py
@@ -487,7 +487,6 @@ except ImportError:
     pass
 
 
-@implements_iterator
 class Cycler(object):
     """A cycle helper for templates."""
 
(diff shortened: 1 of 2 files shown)


ThiefMaster:
s/Standart/Standart/ but I think you could easily squash those three commits into a single one.

jinja2/utils.py
```diff
@@ -495,7 +494,7 @@ def __init__(self, *items):
         if not items:
             raise RuntimeError('at least one item has to be provided')
         self.items = items
-        self.reset()
+        self.reset()
```
ThiefMaster: :bomb: Trailing whitespace is a bad idea :)

tests/test_api.py
```diff
@@ -53,7 +53,7 @@ def test_cycler(self, env):
         c = Cycler(*items)
         for item in items + items:
             assert c.current == item
-            assert next(c) == item
+            assert c.next() == item
```
ThiefMaster: Both `next(c)` and `c.next()` should be tested

jinja2/utils.py
```diff
@@ -512,6 +511,8 @@ def __next__(self):
         self.pos = (self.pos + 1) % len(self.items)
         return rv
 
+    next = __next__
```
ThiefMaster: I'd do it the other way round and rename the method since a method named `next()` is part of the public API while `__next__` is just a nice extra.

tark-hidden: In that case test will be failed, there are two iteration ways... I'm sorry, right now I can use only web for pull request and I do not know how to squash commits here.

PR #494, opened 2015-09-19
Simplify and correct include validation

watsonator:
* Allow for parent directories to be included
* Disallow absolute paths as includes (these were being misinterpreted)
* Expanded unit tests to match

### DIFF
--- a/jinja2/loaders.py
+++ b/jinja2/loaders.py
@@ -10,6 +10,7 @@
 """
 import os
 import sys
+import re
 import weakref
 from types import ModuleType
 from os import path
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Is there any real gain from being able to use `foo/../bar.html` in a template path? This seems more like something that would just encourage some people to write less readable code.

jinja2/loaders.py
```diff
@@ -20,20 +21,14 @@
...
-           piece == path.pardir:
-            raise TemplateNotFound(template)
-        elif piece and piece != '.':
-            pieces.append(piece)
+    """Split a path into segments and perform a sanity check.  If it detects 
+    an absolute path it will raise a `TemplateNotFound` error."""
+    template = os.path.normpath(template)
+    altsep = os.altsep if os.altsep else ''
+    pieces = re.split(r'[%s%s]' % (os.sep, altsep), template)
+    if pieces[0] == '': raise TemplateNotFound(template)
```
ThiefMaster:
```
if not pieces[0]:
    raise ...
```

davidism:
I don't like that this changes how template paths are split, see my comments on #558.  Template paths are not filesystem paths, they just happen to look similar.

ThiefMaster:
It also seems very confusing when you have more complex loader setups where you don't simply load templates from a certain filesystem location.

Imagine you have a loader that takes care of `/foo/*` and one that loads everything else. Would `/foo/../bar.html` be `/bar.html` for the `foo` loader, an invalid path (trying to go above the root of `/foo/`) or `/bar.html` in the other loader?

PR #495, opened 2015-09-20
Added Python 3.5 to tox and Travis configuration

snoack:
Python 3.5 is out. So we should run tests there as well.

Note that Travis currently requires to specify `3.5.0` explicitly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,13 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5.0"
 
 install:
     - pip install tox
 script:
     - tox -e \
-      $(echo py$TRAVIS_PYTHON_VERSION | tr -d . | sed -e 's/pypypy/pypy/')
+      $(echo $TRAVIS_PYTHON_VERSION | sed 's/^\([0-9]\)\.\([0-9]\).*/py\1\2/')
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,12 +4,13 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5.0"
```
untitaker: You can just use "3.5" here, technically no need to fix the regex. Still like your version better though.

untitaker:
Thanks!

.travis.yml
snoack: The problem is that Travis CI currently doesn't have an alias for `3.5.0` -> `3.5`. Hence you need to request `3.5.0` explicitly, for the time being. However, in order to support the longer version number I had to improve the conversion below. Which I agree, is cleaner that way, anyway.

untitaker: I don't see how that's true. https://travis-ci.org/mitsuhiko/jinja2/builds/81271984

snoack: Alright.It seems they just fixed it, see travis-ci/travis-ci#4788.

snoack:
We forgot to update the classifiers in `setup.py` as well. I submitted another PR: #500.

PR #496, opened 2015-09-20
Use container-based travis infrastructure

ThiefMaster:
This should result in faster Travis builds and also gets rid of the warning currently shown on builds.


adrianmoisey:
\o/

PR #497, opened 2015-09-20
Added pypy to the list of Python versions tested by Travis CI

snoack:
Travis CI supports PyPy as well. So we should run the tests there as well.

Note that there is also `pypy3`. However, PyPy 3 is currently based on Python 3.2 which isn't supported.


untitaker:
Thanks!

PR #499, opened 2015-10-05
Simplified tox invocation from Travis configuration

snoack:
Apparently, we can call tox for the current Python version, in a way simpler fashion, from Travis, by simply calling `tox -e py`. This is how Werkzeug and Flask are doing it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,7 @@ python:
 install:
     - pip install tox
 script:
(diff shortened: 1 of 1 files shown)


jeffwidman:
Looks like this just got fixed in d760b78a2f74156f6db85f5f50304b906f3f5d6b. Sorry for missing this earlier PR.

PR #500, opened 2015-10-06
Added Python 3.4 and 3.5 to the classifiers in setup.py

snoack:
It seems that somebody forgot to keep the compatible Python versions listed in `setup.py` up to date. Python 3.5 is already out, and the tests are already performed and pass there. However, in `setup.py`, Python 3.3 is the latest version listed.


jeffwidman:
Thanks!

PR #501, opened 2015-10-11
Copies philosophy from README

ughitsaaron:
This copies the (more gender neutral) philosophy description from the README to setup.py.


ThiefMaster:
why not just s/him/them/?

ughitsaaron:
@ThiefMaster because this the wording already used in the readme (not to mention that it's less clumsy)

PR #502, opened 2015-10-13
runtime: avoid assumption that all objects provide __call__

berrange:
Objects which are backed by native extensions do not provide
a `__call__` attribute, but are none the less callable if the
native extension provides a `tp_call` implementation.

The `jinja2.runtime.Context.call` method unconditionally
access the `__call__` attribute causing an exception to be
raised if the object was a native extension method.

A demo of the problem can be seen using PyGObject:

```
  $ cat demo.py
  #!/usr/bin/python

  from gi.repository import Gio

  f = Gio.File.new_for_path("/some/file.txt")
  print f.get_uri()

  t = Environment().from_string("""{{f.get_uri()}}""")
  print t.render(f=f)
```

Which when run results in

```
 $ ./demo.py
 file:///some/file.txt
   File "./demo.py", line 10, in <module>
     print t.render(f=f)
   File "<template>", line 1, in top-level template code
     AttributeError: 'gi.FunctionInfo' object has no attribute '**call**'
```

After this patch it produces

```
 $ ./demo.py
 file:///some/file.txt
 file:///some/file.txt
```

Signed-off-by: Daniel P. Berrange berrange@redhat.com

### DIFF
--- a/jinja2/runtime.py
+++ b/jinja2/runtime.py
@@ -177,13 +177,14 @@ class Context(object):
             __traceback_hide__ = True  # noqa
 
         # Allow callable classes to take a context
-        fn = __obj.__call__
-        for fn_type in ('contextfunction',
-                        'evalcontextfunction',
-                        'environmentfunction'):
-            if hasattr(fn, fn_type):
-                __obj = fn
-                break
+        if hasattr(__obj, '__call__'):
+            fn = __obj.__call__
+            for fn_type in ('contextfunction',
+                            'evalcontextfunction',
+                            'environmentfunction'):
+                if hasattr(fn, fn_type):
+                    __obj = fn
+                    break
 
         if isinstance(__obj, _context_function_types):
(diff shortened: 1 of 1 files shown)


ThiefMaster:
just a small heads-up: you can use fenced code blocks (triple backticks before and after) to format something as code on github. right now your demo code in the PR's description is not very readable

berrange:
@ThiefMaster all the text in the PR was just automatically included by github from the commit message https://github.com/berrange/jinja2/commit/91255f86f6dc3b828142cc339155bcb5a7797486

davidism:
You can edit your posts.

davidism:
@mitsuhiko checking if the callable object is a `FunctionType` or `MethodType` seems unnecessary. Removing it doesn't break any tests and allows callable C objects to mark themselves as context functions too by setting a class attribute instead of a function attribute.

Is there a reason not to remove the `isinstance(obj, _context_function_types)` check?

PR #507, opened 2015-10-24
Escape target attribute in the urlize function in utils.py.

darkryder:
See PR #356.


darkryder:
Review please?

davidism:
Target doesn't really seem like the kind of thing a user would be entering for you, but ok, this seems fine.

PR #525, opened 2015-12-29
Give Cycler a next() method so that it works in python2 and 3

### LINKED ISSUE #474: Cycler object has no attribute 'next' in Python 3.4.3
Python 3.4.3, Pyramid 1.5.7, Jinja 2.8 (2.7.3 produce the same error)

I got a weird exception

```
jinja2.exceptions.UndefinedError: 'jinja2.utils.Cycler object' has no attribute 'next'
```

in string

```
<li class='struct-item{{ row.next() }}' data-id="{{ category._id }}" 
data-owner="{{ category.owner }}">
```

template:

```
<ul>
{% set row = cycler('', ' back') %}
{% for category in categories[0] recursive %}
    <li class='struct-item{{ row.next() }}' data-id="{{ category._id }}" 
    data-owner="{{ category.owner }}">
        <div class="title">{{ category.caption }}</div>
        <div class="action">
            {% if not loop.depth0 %}
            <a href='#' class="category-add">
                <img src='/static/admin/img/action-add.png'>
             </a>
             {% endif %}
            <a href='#' class="category-edit">
                <img src='/static/admin/img/action-edit.png'>
            </a>
            <a href='#' class="category-delete">
                <img src='/static/admin/img/action-delete.png'>
            </a>
        </div>
    </li>
    {% if category._id in categories %}
        <ul class="children">{{ loop(categories[category._id]) }}</ul>
    {% endif %}
{% endfor %}
</ul>
```

When I make a call print(dir(self)) in a constructor Cycler, echo was

```
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', 
 '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__le__', '__lt__', 
 '__module__', '__ne__', '__new__', '__next__', '__reduce__', '__reduce_ex__', 
 '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 
 'current', 'items', 'pos', 'reset']
```

that means Cycler object really has no `next` method. Temporary fix is `jinja2.utils.Cycler.next = jinja2.utils.Cycler.__next__` in a Pyramid`main` method, but...

### LINKED ISSUE #524: python3: 'jinja2.utils.Cycler object' has no attribute 'next'
The `jinja2.utils.Cycler` object no longer has a `next()` method in Python 3:

Running the following with Python 3:

``` python

env = Environment()
t = env.from_string('''
{% set a = cycler('odd', 'even') %}
{{ a.next() }}
{{ a.next() }}
''')
result = t.render()
print(result)
```

Results in this error:

```
  File "example.py", line 9, in <module>
    result = t.render()
  File "/private/tmp/a/py3/lib/python3.4/site-packages/jinja2/environment.py", line 989, in render
  File "/private/tmp/a/py3/lib/python3.4/site-packages/jinja2/environment.py", line 754, in handle_exception
  File "/private/tmp/a/py3/lib/python3.4/site-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
jinja2.exceptions.UndefinedError: 'jinja2.utils.Cycler object' has no attribute 'next'
```

I'm going to work around this by adding the new built-in `next()` function as an environment global.

iffy:
Fixes #493
Fixes #474
Fixes #524

### DIFF
--- a/jinja2/utils.py
+++ b/jinja2/utils.py
@@ -506,11 +506,13 @@ class Cycler(object):
         """Returns the current item."""
         return self.items[self.pos]
 
-    def __next__(self):
+    def next(self):
         """Goes one item ahead and returns it."""
         rv = self.current
         self.pos = (self.pos + 1) % len(self.items)
         return rv
+
+    __next__ = next
 
 
 class Joiner(object):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -59,6 +59,17 @@ class TestExtendedAPI():
         c.reset()
         assert c.current == 1
 
+    def test_cycler_nextmethod(self, env):
+        items = 1, 2, 3
+        c = Cycler(*items)
+        for item in items + items:
+            assert c.current == item
+            assert c.next() == item
+        c.next()
+        assert c.current == 2
+        c.reset()
+        assert c.current == 1        
+
     def test_expressions(self, env):
         expr = env.compile_expression("foo")
         assert expr() is None


iffy:
@ThiefMaster Let me know if I should change anything.

ThiefMaster:
Looks good for me

jinja2/utils.py
```diff
@@ -506,12 +506,14 @@ def current(self):
         """Returns the current item."""
         return self.items[self.pos]
 
-    def __next__(self):
+    def next(self):
```
kevinchevalier: This is done automatically with the implements_iterator decorator.

kevinchevalier: This is done automatically with the implements_iterator decorator.

iffy: In Python 3, `implements_iterator` does nothing.  But in Python 2, it renames `__next__` to `next` (and deletes `__next__`).  <https://github.com/mitsuhiko/jinja2/blob/master/jinja2/_compat.py#L70>

If only a `__next__` method is explicitly defined, the Python 3 tests fail with:

```
    def test_cycler_nextmethod(self, env):
        items = 1, 2, 3
        c = Cycler(*items)
        for item in items + items:
            assert c.current == item
>           assert c.next() == item
E           AttributeError: 'Cycler' object has no attribute 'next'
```

I think the test's assertion is correct because the [docs](http://jinja.pocoo.org/docs/dev/templates/#cycler.next) imply that a `Cycler` has a method named `next()`.

jeffwidman:
@ThiefMaster are we good to merge this as well as close the other linked PRs? 

I started to dig into this, but looking at the linked issues it seems you already understand what's going on here.

ThiefMaster:
Seems fine, but I'll have another look in the evening when I'm back from work

jeffwidman:
@ThiefMaster did you get a chance to review?

ThiefMaster:
Maybe it would be a good idea to remove `@implements_iterator` since it doesn't do anything now?

jinja2/utils.py
jeffwidman: I just merged https://github.com/pallets/jinja/pull/470 which affects some of the next behavior in `_compat.py`. I haven't looked closely enough at this PR to understand if it was affected, just noticed you mentioned `next`/`__next__` and thought I'd mention the merge.

Issue #529, opened 2016-01-07
Using Undefined in subtraction does not raise UndefinedError


nick-garcia:
```
>>> from jinja2 import Undefined
>>> foo = Undefined(name='foo')
>>> foo - 1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for -: 'Undefined' and 'int'
```

nick-garcia:
Created pull request #530

PR #530, opened 2016-01-07
Subtraction from Undefined does not raise UndefinedError

nick-garcia:
Setting **sub** equal to _fail_with_undefined_error so subtractions with undefined variables will fail properly like other arithemtic operations.


nick-garcia:
Fix for #529

chaosagent:
While you're at it you might also want to add `__rsub__` and everything else listed [https://docs.python.org/2/reference/datamodel.html#emulating-numeric-types](here)

PR #534, opened 2016-01-23
Grunt/Gulp-style loader and tests

gregorysenyshyn:
I wrote this loader to when I moved some personal projects over from Grunt. I'm not sure if you think it would be a useful addition, but it simplified switching to jinja2 for me.  It lets you specify a list of files using wildcards, like on the command line, and also to exclude files from that list. If you decide to merge it and need documentation written, just let me know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ build
 dist
 .DS_Store
 .tox
+.gitignore
+.cache/
+venv
--- a/jinja2/loaders.py
+++ b/jinja2/loaders.py
@@ -10,7 +10,9 @@
 """
 import os
 import sys
+import glob
 import weakref
+import itertools
 from types import ModuleType
 from os import path
 from hashlib import sha1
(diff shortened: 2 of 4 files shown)


jeffwidman:
I'm sorry, but I've never used Grunt/Gulp, so I'm not really sure how this helps. 

Can you explain a little more about why it's useful?

gregorysenyshyn:
I'm using Jinja to build static sites. This loader gives me more granular control when selecting files, which makes it much easier for me to have separate development and production build processes. 

This loader lets me do path expansion using ** (you could use "src/**/*.html" to select all the html files in the directories one level deep in src/, which helps to keep down the length of my build files, and also helps to keep my line lengths under 80 characters if I have deep directory structure or long directory names).  Python 3.5's glob module (or higher) is needed, though, so that part will only be useful for people who are running a fairly new Python.

The most useful thing to me is to be able to exclude files by prepending "!".  On builds I don't push to production, I usually exclude analytics code, for example, which is easy to do by just adding "!src/partials/analytics". It's nice to be easily able to exclude a handful of files when selecting a lot of directories at once using **.

davidism:
Thanks, but I'd rather keep this separate from Jinja core. If you or others would find this useful, please consider starting an extension package, such as `jinja-gulp-loader`.

Issue #63, opened 2011-10-02
Output finalize function issue


tzzt:
The finalize function seems to be executed on the whole template, not only on variables.

E.g. a function like 
def fin(s): print s; return s 
dumps the whole template file.

godsarmy:
is this fixed? i still see it in the latest version (2.8).

```
rom jinja2 import Environment, Template

context = {'k': 'v'}

template = """
abc
{{ k }}
def
"""

def fin(data):
    return 'replaced'

env = Environment(finalize=fin)

tpl = env.from_string(template, template_class=Template)

print tpl.render(**context)
```

It displayed 'replacedreplacedreplaced', we want this to be fixed as we need this feature very explicitly

PR #535, opened 2016-01-24
Finalize should only apply to variable #63

godsarmy:
I submit this PR to resolve the issue of #63 as i believe it is incorrect to apply finalize to any value in the template (constant, condition).

I think the root cause is self-explaining in the unit test "<{{ none }} >" as the original implementation tried to change none to empty string. However, none is not a variable in context but a constant, so instead, it should not be applied with finalize function.

```
tmpl = env.from_string('<{{ none }}>')
```

### DIFF
--- a/jinja2/compiler.py
+++ b/jinja2/compiler.py
@@ -1219,20 +1219,6 @@ class CodeGenerator(NodeVisitor):
         if self.has_known_extends and frame.require_output_check:
             return
 
-        allow_constant_finalize = True
-        if self.environment.finalize:
-            func = self.environment.finalize
-            if getattr(func, 'contextfunction', False) or \
-               getattr(func, 'evalcontextfunction', False):
-                allow_constant_finalize = False
-            elif getattr(func, 'environmentfunction', False):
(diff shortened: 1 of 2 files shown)


davidism:
The lower case versions `none`, `true`, and `false`, are equivalent to their uppercase Python names in Jinja.  Even if the variable was changed to be a truly undefined value, undefined values are changed to `none` by default.  Changing `none` to `''` is correct in this case.

LizardM4:
As a side remark, if `finalize` were not to be called on the template source, this would enable a much higher degree of customization -- one could autoescape the variables for languages other than HTML (see e.g. [this question on SO](http://stackoverflow.com/q/8001986/1749822), which originally brought me here).

and1er:
What the status of this PR? Is the problem with LaTeX autoescape still cannot be solved https://stackoverflow.com/q/8001986/7444782 ?

davidism:
`{{ none }}` is still an expression, just one that has been optimized to a constant. There is a distinction between static template data and constant expressions. All expressions should pass through finalize, template data should not.

Issue #516, opened 2015-12-01
gettext fails when source text has a "%" sign in it


eastside:
Take this example--

```
{{_('Get 50% off your first purchase')}}
```

Raises the following error--

```
  File "/home/vagrant/perkville/businesses/templates/business_profile.jinja2", line 194, in template
    <p>{{_('Get 50% off your first purchase')}}</p>
  File "/home/vagrant/perkville.com/local/lib/python2.7/site-packages/jinja2/ext.py", line 132, in _gettext_alias
    return __context.call(__context.resolve('gettext'), *args, **kwargs)
  File "/home/vagrant/perkville.com/local/lib/python2.7/site-packages/jinja2/ext.py", line 141, in gettext
    return rv % variables
  File "/home/vagrant/perkville.com/local/lib/python2.7/site-packages/markupsafe/__init__.py", line 101, in __mod__
    return self.__class__(text_type.__mod__(self, arg))
TypeError: %o format: a number is required, not _MarkupEscapeHelper
```

The problem seems to be here--

```
def _make_new_ngettext(func):
    @contextfunction
    def ngettext(__context, __singular, __plural, __num, **variables):
        variables.setdefault('num', __num)
        rv = __context.call(func, __singular, __plural, __num)
        if __context.eval_ctx.autoescape:
            rv = Markup(rv)
        return rv % variables
    return ngettext
```

For some reason, it always calls modulo with `variables`, even if there are no variables. This doesn't seem to be correct.

fergyfresh:
`{{_('Get 50% off your first purchase')}}`

might need raw tags, I'm not home so I can't check it. It seems like it doesn't like that you're stomping on the jinja2 template percent sign %!!!!

try 

`{{_('{{% raw %} Get 50% off your first purchase {% raw %}')}}`

It is somehow related to you not escaping that percent sign though.

ThiefMaster:
Why not use `{% trans %}Get 50% off your first purchase{% endtrans %}`?

ThiefMaster:
This has nothing to do with `{% raw %}` tags.

fergyfresh:
Sorry @ThiefMaster  I will delete my post.

ThiefMaster:
ok, deleted my comment too

eastside:
My understanding was that {% trans %} was a wrapper for gettext. And indeed, the same error results when using {% trans %} tags.

ThiefMaster:
I guess that shouldn't happen..

[...]

PR #541, opened 2016-02-13
ignore error when gettext string formatting fails with % in source text

razor-1:
If the string returned by gettext has a %, but there is no substitution to be performed (and indeed, the % may prefix an invalid character), then instead of raising an exception and failing completely, we return the original text.

### DIFF
--- a/jinja2/ext.py
+++ b/jinja2/ext.py
@@ -140,7 +140,7 @@ def _make_new_gettext(func):
             rv = Markup(rv)
         try:
             return rv % variables
-        except:
+        except TypeError:
             return rv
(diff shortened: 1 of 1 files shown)


jeffwidman:
Shouldn't you catch only the specific exception mentioned in #516 and not blindly swallow everything?

davidism:
Shouldn't the string returned use `%%` if it needs the symbol itself?

davidism:
Thanks for working on this. Based on my investigation, I believe the correct answer here is to use `%%` to escape percent signs in format strings.

PR #558, opened 2016-03-07
Do not hardcode a forward slash as path separator.

masasin:
path.sep is already being used below, but the hardcoding prevents programs from using something made with path.join on Windows.

### DIFF
--- a/jinja2/loaders.py
+++ b/jinja2/loaders.py
@@ -24,7 +24,7 @@ def split_template_path(template):
     '..' in the path it will raise a `TemplateNotFound` error.
     """
     pieces = []
-    for piece in template.split('/'):
+    for piece in template.split(path.sep):
         if path.sep in piece \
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Might want to add a testcase (that also ensures you can use forward slashes even if `os.path.sep` is a backslash, e.g. on windows). Many people do not want to use backslashes in paths in their code (and template paths are more often than not static strings, i.e. do not use `os.path.join`)

davidism:
This is intended behavior.  The only path separator is `/` because template paths are not filesystem paths, although they do look similar.  Use `'/'.join(path, values)` to build a template path.

davidism:
This change would break paths working across systems, devs on Windows wouldn't be able to find templates paths written by devs on Linux.

masasin:
Unswearoos.

PR #560, opened 2016-03-13
Move jinja.vim to syntax directory

robsonpeixoto:
Now the vim plugin can be installed using plugin manager like Plug.
```vim
Plug 'mitsuhiko/jinja2', {'rtp': 'ext/Vim'}
```


jeffwidman:
Do you know if this change affects any other vim/neovim plugin managers? I'm not sure if the `vim/syntax` directory is a common idiom in that world... or if it'd break any of the others.

robsonpeixoto:
I think that it will start to work after this change.

Other example: https://github.com/docker/docker/tree/master/contrib/syntax/vim

jeffwidman:
I just looked at the Plug docs a little more, and it doesn't require the plugin to be in `syntax` directory... see also examples: 
http://vi.stackexchange.com/questions/5393/whats-the-correct-way-to-install-youcomplete-using-vim-plug-in-vim-nvim
http://www.adamwadeharris.com/how-to-switch-from-vundle-to-vim-plug/

So I don't really see the point of this.

PR #577, opened 2016-05-01
Small documentation fixes

### LINKED ISSUE #543: There is no 'sudo' command in the documentation
From Installation part of the documentation:
(If you are installing from the Windows command line OMIT THE SUDO and make sure to run the command as user with administrator rights)

But there is no sudo command, there is only
easy_install Jinja2
pip install Jinja2

Which sudo i need to omit? Can you add it to easy_install and pip?

### LINKED ISSUE #544: Word true need to be `true`
In the line https://github.com/mitsuhiko/jinja2/blob/master/jinja2/environment.py#L182 the word true should be `True`

boidolr:
- fixes #543 (removing `sudo` from installation instructions)
- fixes #544 (using `True` instead of true)

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -36,15 +36,12 @@ You can install the most recent Jinja2 version using `easy_install`_ or `pip`_::
 This will install a Jinja2 egg in your Python installation's site-packages
 directory.
 
-(If you are installing from the Windows command line omit the `sudo` and make
-sure to run the command as user with administrator rights)
-
 From the tarball release
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


jeffwidman:
Thanks!

davidism:
rst uses double ticks for code, single ticks denote references.  I realize other parts of the documentation don't follow this, but please use the correct syntax in future changes.  Thanks.

jeffwidman:
Thanks for catching that. I wasn't aware of that til now.

boidolr:
Ok, thanks for the info. I've only noticed the other parts and adapted that part.

PR #582, opened 2016-05-04
Bug: Jinja fails with unknown tags in {% for %} which is not executed

atodorov:
Hey guys,
I have a code similar to this test, which loops over defined (Pelican) plugins and tries to use the webassets module if the plugin is enabled. If not falls back to not using it. However when rendering the template Jinja raises
```
TemplateSyntaxError: Encountered unknown tag 'assets'. Jinja was looking for the following tags: 'endfor' or 'else'. The innermost block that needs to be closed is 'for'.
```

which tells me that it tries to parse the `{% assets %}` tag before actually running the loop (or not running it more precisely). Is there any way to solve this or that's a known limitation and can't be solved ?

### DIFF
--- a/tests/res/templates/unknown_tag_in_for_loop.html
+++ b/tests/res/templates/unknown_tag_in_for_loop.html
@@ -1,0 +1,7 @@
+{% set PLUGINS = ['something-else'] %}
+{% for name in PLUGINS if name == 'assets' %}
+  {% assets %}
+  {% endassets %}
+{% else %}
+  assets is disabled
+{% endfor %}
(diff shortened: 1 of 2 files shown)


ThiefMaster:
I think the current behavior makes lots of sense. Using an undefined tag is like a syntax error and such errors should always cause an error, even if that part of the code is not executed.

atodorov:
@ThiefMaster - are there any alternatives instead of fail? Maybe an option to silently ignore tags which are undefined ?

ThiefMaster:
For your specific usecase you could probably put the code in a separate template and conditionally `{% include %}` the template only if you want to use assets.

atodorov:
Thanks, the conditional include did work.

Issue #485, opened 2015-09-01
2.8.0 performance regression compared to 2.7.3


jstasiak:
It seems like with Jinja2 2.8.0 there's extra IO being performed even with templates being cached and auto reloading disabled.

Test case (we use template includes a lot and this is a relatively good test):

```python
# main.py
from jinja2 import Environment, FileSystemLoader

if __name__ == '__main__':
    environment = Environment(
        auto_reload=False,
        loader=FileSystemLoader(['.']),
    )
    template = environment.get_template('template.txt')

    for i in xrange(1000):
        template.render()
```

```
# template.txt
{% for i in range(100) %}
    {% include 'part.txt' %}
{% endfor %}
```

```
# part.txt
this is part
```

With Jinja2 2.7.3:

```
./jinja2-27/bin/python -m cProfile -s tottime ./main.py | head -n 20
         2536172 function calls (2535412 primitive calls) in 2.276 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   101000    0.436    0.000    1.033    0.000 runtime.py:53(new_context)
   101000    0.380    0.000    0.546    0.000 runtime.py:112(__init__)
   301000    0.309    0.000    2.148    0.000 template.txt:5(root)
   100001    0.178    0.000    0.665    0.000 environment.py:770(get_template)
   100003    0.143    0.000    0.205    0.000 utils.py:380(__getitem__)
   100001    0.097    0.000    0.394    0.000 environment.py:756(_load_template)
   100003    0.084    0.000    0.290    0.000 utils.py:336(get)
   101003    0.082    0.000    0.094    0.000 nodes.py:81(__init__)
   101000    0.077    0.000    1.110    0.000 environment.py:995(new_context)
   201010    0.070    0.000    0.103    0.000 _compat.py:57(<lambda>)
1003/1002    0.065    0.000    2.213    0.002 {method 'join' of 'unicode' objects}
   102365    0.055    0.000    0.055    0.000 {isinstance}
   100017    0.042    0.000    0.042    0.000 {locals}
   100006    0.040    0.000    0.040    0.000 {method 'acquire' of 'thread.lock' objects}
   201010    0.033    0.000    0.033    0.000 {method 'iteritems' of 'dict' objects}
```

With Jinja2 2.8.0:

```
./jinja2-28/bin/python -m cProfile -s tottime ./main.py | head -n 20
[...]

PR #583, opened 2016-05-09
Change cache key definition in environment

pgjones:
In 6671b973e6de5abc46829a27fd3bbb989d68ca3a the load_template method
was altered to use a cache key other than the template name. The key
chosen was the abs path as returned from the loader get_source
method. Unless there is no path in which case the name is
used. Unfortunately this introduced a performance regression, #485, as
the get_source method (in the FileStoreLoader) loads the template
(causing IO).

The purpose of #332 was to allow the loader to change whilst ensuring
the correct template was loaded, i.e. to fix this case

    env.loader = loader1
    env.get_template('index.html') # return loader1/index.html
    env.loader = loader2
    env.get_template('index.html') # also return loader1/index.html because of cache

This commit changes the cache key to be a combination of a hash of the
loader and the template name. Therefore fixing the above case without
calling the get_source method and thereby avoiding the IO load.

This additionally fixes the above use case for loaders that do not return filenames, e.g. `DictLoader`. However it will likely introduce cache misses should properties of the loader change (and hence the hash changes).

### DIFF
--- a/jinja2/environment.py
+++ b/jinja2/environment.py
@@ -769,15 +769,7 @@ class Environment(object):
     def _load_template(self, name, globals):
         if self.loader is None:
             raise TypeError('no loader for this environment specified')
-        try:
-            # use abs path for cache key
-            cache_key = self.loader.get_source(self, name)[1]
-        except RuntimeError:
-            # if loader does not implement get_source()
-            cache_key = None
-        # if template is not file, use name for cache key
-        if cache_key is None:
-            cache_key = name
+        cache_key = (id(self.loader), name)
         if self.cache is not None:
             template = self.cache.get(cache_key)
             if template is not None and (not self.auto_reload or
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -78,19 +78,31 @@ class TestLoaders():
         assert tmpl is not env.get_template('template')
         changed = False
 
-        env = Environment(loader=TestLoader(), cache_size=0)
-        assert env.get_template('template') \
-            is not env.get_template('template')
+    def test_no_cache(self):
+        mapping = {'foo': 'one'}
+        env = Environment(loader=loaders.DictLoader(mapping), cache_size=0)
+        assert env.get_template('foo') is not env.get_template('foo')
 
-        env = Environment(loader=TestLoader(), cache_size=2)
+    def test_limited_size_cache(self):
+        mapping = {'one': 'foo', 'two': 'bar', 'three': 'baz'}
+        loader = loaders.DictLoader(mapping)
+        env = Environment(loader=loader, cache_size=2)
         t1 = env.get_template('one')
         t2 = env.get_template('two')
         assert t2 is env.get_template('two')
         assert t1 is env.get_template('one')
         t3 = env.get_template('three')
-        assert 'one' in env.cache
-        assert 'two' not in env.cache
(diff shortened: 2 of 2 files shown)


jeffwidman:
Thanks for putting this together, it's definitely something that needs to get fixed. Your suggested solution makes sense to me.

Looks like some tests are failing though?

pgjones:
Thanks, I've rebased with updates to the tests. I've both fixed the braking test and added one to ensure that the correct template is returned after changing the loader.

jinja2/environment.py
```diff
@@ -769,15 +769,7 @@ def join_path(self, template, parent):
...
-        try:
-            # use abs path for cache key
-            cache_key = self.loader.get_source(self, name)[1]
-        except RuntimeError:
-            # if loader does not implement get_source()
-            cache_key = None
-        # if template is not file, use name for cache key
-        if cache_key is None:
-            cache_key = name
+        cache_key = "%d%s" % (hash(self.loader), name)
```
iceboy-sjtu: What about `cache_key = (self.loader, name)`?

This avoids string conversion and hash collision.

pgjones: Good idea, I've updated.

pgjones:
I've just rebased following @iceboy-sjtu comment.

pgjones:
@jeffwidman Anything else I need to do?

jeffwidman:
LGTM, but I'd like another maintainer to take a look as well before merging just in case I missed any edge cases. Especially since the last time someone tried to improve this section of code it introduced this bug... Definitely would like this to land in the next release though.

cc @davidism @untitaker

untitaker:
With this we would break cases where somebody sets `env.cache` to something that stores data outside of the current Python process. I don't know if that makes sense at all (given that the external support for `bytecode_cache` already exists), and if it's something we want to support. If not, I don't understand why `env.cache` is a public attribute and not preceded by a underscore.

pgjones:
@untitaker Thanks; I think that it doesn't make sense to change the `env.cache` given the `bytecode_cache`. If it does however, a potential issue could be the change from str key to tuple, which I made following a [suggestion](https://github.com/pallets/jinja/pull/583#discussion-diff-62604115R772). 

I'd like to add that the performance regression #485 does break jinja for us, so I'd like to help reach a solution.

untitaker:
@mitsuhiko confirmed in IRC that cache should not be overridden in such a way.

However, he voiced the concern that a reference to the loader is now always kept. He suggests replacing `loader` with `id(loader)` for the cache key.

pgjones:
@untitaker Thanks, change made.

untitaker:
Thanks!

pgjones:
Thank you

iceboy-sjtu:
It seems that cache key should be some weak reference to the loader. By using `id()` you have the potential risk that a new cache having a same address gets old content from the cache.

untitaker:
@iceboy-sjtu You may be right, please open a PR.

pgjones:
@iceboy-sjtu what about going back to the `hash()`?

iceboy-sjtu:
@pgjones `weakref.ref()` seems most suitable for this case. And it's even better to remove the cache entries when a loader is destroyed.

Personally I hate using `hash()` this way, where there will be one in a million case that the hash collides and some mysteries strange behavior appears.

PR #585, opened 2016-05-21
Change environment cache key construction

pgjones:
Changing from a tuple of the loader ID and template name to a weakref
to the loader and the template name should avoid situations whereby
the loader has changed, yet the cached templates are returned. This
would occur if the id of the new loader matches the old. A weakref is
preferred over a direct reference so that the loader can be garbaged
collected.

This change was suggested in #583 by @iceboy-sjtu.

### DIFF
--- a/jinja2/environment.py
+++ b/jinja2/environment.py
@@ -10,6 +10,7 @@
 """
 import os
 import sys
+import weakref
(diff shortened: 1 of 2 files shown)


jeffwidman:
Thanks!

PR #590, opened 2016-06-08
Let the length template filter work on iterables.

alexjurkiewicz:
This will allow a construction like this:
```
{{ mylist|reject(...)|length }}
```
Instead of:
```jinja
{{ mylist|reject(...)|list|length }}
```

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -359,6 +359,18 @@ def do_last(environment, seq):
         return environment.undefined('No last item, sequence was empty.')
 
 
+def do_length(seq):
+    """Return the length of a sequence, or iterable."""
+    try:
+        return len(seq)
+    except TypeError:
+        if isinstance(seq, collections.Iterable):
+            # len(list(seq)) would read the entire seq into RAM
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Not a good idea since you consume the iterable that way, meaning you cannot use it anymore afterwards. It might not be an issue in your case but what if you pass an iterable from outside?

I'd rather add a custom filter (in the application, not jinja itself) like this:

```python
def itercount(iterable):
    return sum(1 for _ in iterable)
```

alexjurkiewicz:
Either way the iterable is consumed, right? I would just like it to be part of the core filter to simplify the above construction, which I've written a lot lately.

ThiefMaster:
Yes, but `len()` in python also fails when called on an generator, probably to make it clear that you cannot get the length of a generator without consuming it.

alexjurkiewicz:
I was hoping the explicitly different name of `length` gave some semantic wriggle room here :)

davidism:
You can always write a `listlength` filter to do this, but `length` should behave similarly to the built-in.

PR #595, opened 2016-06-18
Simple text change

bngsudheer:
look and see are redundant


ThiefMaster:
look-see is an actual word. https://en.wiktionary.org/wiki/look-see

PR #600, opened 2016-07-20
Move version headings directly beneath topic headings in documentation

zyphlar:
To someone reading the documentation, "how do I use Jinja" is their first question and "how do I accomplish a specific task" is their second question. When documented options are not available to most users yet, it's important for readability that disclaimers be treated as section headings, not footers. https://www.prismnet.com/~hcexres/textbook/headings.html

I realize that Python docs themselves put version info in the footer of each section, but I feel that is an anti-pattern unless the entire docs are versioned (I spent about 5 minutes futilely hunting for "Jinja 2.7 docs" -- the only docs I found are these "Warning: development version" docs.)

*tl;dr: version tags are top-level information, not footer-level information. Manuals are for people who are new, and may not understand your writing style yet.*

**This is a blind untested edit just to do the work I'm requesting be done; it should be double-checked and tested before merging.**

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -44,22 +44,24 @@ additional methods:
 
 .. method:: jinja2.Environment.install_gettext_translations(translations, newstyle=False)
 
+    .. versionchanged:: 2.5 newstyle gettext added
+
     Installs a translation globally for that environment.  The translations
     object provided must implement at least `ugettext` and `ungettext`.
     The `gettext.NullTranslations` and `gettext.GNUTranslations` classes
     as well as `Babel`_\s `Translations` class are supported.
 
-    .. versionchanged:: 2.5 newstyle gettext added
-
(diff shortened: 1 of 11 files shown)


davidism:
The version information is for users who are upgrading from previous versions.  The current usage is most important, followed by any notable changes.  Reading through a changelog before getting to the actual docs each time is counterproductive in the vast majority of cases.

davidism:
There should be builds for other versions of the docs, but that's not solved here, it's a matter of enabling the builds.  There is a meta issue tracking doc builds.

zyphlar:
It may be intended for users who are upgrading, but the actual use-case of a newbie trying to write a template without necessarily knowing what Jinja version they have and just wanting to write a damn `if` statement is probably more common. My point about version info being at the top of the hierarchy not a footnote stands, and moving versioning info to the top does nothing to detract from the upgrading use-case.

Just because Python does it doesn't mean it's the best way to do it.

PR #602, opened 2016-08-02
Prevent open_if_exists from raising an exception

### LINKED ISSUE #399: method open_if_exists raises exception, instead of returning None
According to description, method open_if_exists instead of raising exception should return None.

https://github.com/mitsuhiko/jinja2/blob/master/jinja2/utils.py#L146

EDIT: Forgot to specify - happens only on windows.

edouardklein:
If the directory filename supposedly lives in does not exist, open_if_exists raises a NotADirectoryError, which may, or may not, be the expected behavior.

Fixes #399, which caused https://github.com/noirbizarre/flask-restplus/issues/187 <- an exemple of client code where not raising the exception was the expected behavior.

### DIFF
--- a/jinja2/_compat.py
+++ b/jinja2/_compat.py
@@ -36,6 +36,7 @@ if not PY2:
         if value.__traceback__ is not tb:
             raise value.with_traceback(tb)
         raise value
+    NotADirectoryError = NotADirectoryError
 
     ifilter = filter
     imap = map
@@ -62,6 +63,8 @@ else:
     NativeStringIO = BytesIO
 
     exec('def reraise(tp, value, tb=None):\n raise tp, value, tb')
+    class NotADirectoryError(Exception):
+        pass
 
     from itertools import imap, izip, ifilter
     intern = intern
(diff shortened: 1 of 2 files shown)


edouardklein:
It appears that the code I proposed is not python2-compatible, because NotADirectoryError does not exist in python2...

edouardklein:
OK, so I redefined NotADirectoryError in __compat.py, the travis builds now pass even for Python2. I would be happy if you could consider the pull request.

Thanks :)

fabioramponi:
Hi guys, will this pull request be considered for future releases? Otherwise, could you suggest some alternative solution to prevent [#399](https://github.com/pallets/jinja/issues/399)?

Thanks!

davidism:
I'm not merging this as-is, catching all exceptions (the Python 2 compat) is a bad idea.

davidism:
`NotADirectoryError` is a subclass of `IOError` (which is an alias for `OSError`) in Python 3. The correct change would be to add `errno.ENOTDIR` to the list of ignored errors.

However, the error you linked to is due to you installing your package as an egg when the loader being used expects a filesystem. So the file does exist, but the path is invalid for the loader, in which case I wouldn't want to silence the error. All that would lead to would be getting `TemplateNotFound` instead, which would be even more confusing since the template definitely exists, it's just not installed in a way that can be loaded with the given configuration.

PR #613, opened 2016-10-13
Fix warnings, typos and spelling.

romeojulietthotel:
Warnings and typos and spelling, oh my.


romeojulietthotel:
I can't tell if continuous integration builds the docs. I didn't find anything like that. Did I miss some step or am I looking in the wrong place? I looked here: [travis-ci](https://travis-ci.org/pallets/jinja/builds/167228684) Or do the docs not get built automatically?

ThiefMaster:
Some of your changes are useful (e.g. the actual fixes), but others seem rather pointless. "bytecode" is actually more correct than "byte-code" according to google and "autoescaping" makes more sense than "auto escaping" since the Jinja feature this is about is called "autoescape".

ThiefMaster:
Also, your git is misconfigured. You seem to be committing as `None <none>`...

PR #616, opened 2016-10-30
Drop Python 2.6 support

di:
Yesterday marks 3 years since the last release of Python 2.6! 🎉

To celebrate, I'm attempting to drop support for it from 156 prominent Python packages (one for every week it's past end-of-life)--including this one!

I've tried my best to remove as much 2.6-specific cruft as I can, but at the very least, this PR will remove the `'Programming Language :: Python :: 2.6'` trove classifier from this projects `setup.py`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: python
 python:
-  - "2.6"
   - "2.7"
   - pypy
   - "3.3"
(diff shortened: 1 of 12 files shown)


ThiefMaster:
As much as I think Python 2.6 needs to die, there is absolutely no gain from doing this except locking out users.

PR #617, opened 2016-10-30
Don't skip an erroneous comma when parsing "from"

jackwilsdon:
Currently we skip an extra comma when "with context" is not provided to the "from" statement. This allows invalid code such as this:

``` django
{% from "functions.html" import my_function,, %}
```

The primary issue with this is that it is not consistent when providing "with context". The following code throws an error, contrary to the previous example:

``` django
{% from "functions.html" import my_function,, with context %}
```

It seems that the comma skipping was originally found in 0611e49 and was accidentally put inside the "with context" check in ea847c5. It was then updated to use "skip_if" in fdf9530.

There doesn't seem to be any reason for this check existing, as commas should never be allowed and serve no purpose in the statement.

Sadly this may be a "semi-breaking" change, as it removes some functionality that should have never existed and hopefully nobody relies on.

This PR ensures that both examples throw an error as expected.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -28,11 +28,12 @@ Version 2.10
 - Add ``min`` and ``max`` filters. (`#475`_)
 - Add tests for all comparison operators: ``eq``, ``ne``, ``lt``, ``le``,
   ``gt``, ``ge``. (`#665`_)
-- ``import`` statement cannot end with a trailing comma. (`#618`_)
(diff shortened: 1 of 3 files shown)


davidism:
After #618, the example already fails, removing this doesn't change anything. However, since it seems redundant, I'll still remove it and add a test.

PR #620, opened 2016-11-02
Add camelize filter

borismod:
camelize filter converts as following:

```
abstract_factory ➡️  AbstractFactory
abstract-factory ➡️  AbstractFactory
abstract factory  ➡️  AbstractFactory
AbstractGeneric-factory   ➡️  AbstractGenericFactory
```

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -15,14 +15,14 @@ from random import choice
 from itertools import groupby
 from collections import namedtuple
 from jinja2.utils import Markup, escape, pformat, urlize, soft_unicode, \
-     unicode_urlencode
+    unicode_urlencode
 from jinja2.runtime import Undefined
 from jinja2.exceptions import FilterArgumentError
 from jinja2._compat import imap, string_types, text_type, iteritems
 
-
 _word_re = re.compile(r'\w+(?u)')
 _word_beginning_split_re = re.compile(r'([-\s\(\{\[\<]+)(?u)')
+_word_split_snake_for_camelize = re.compile(r'(?:^|_|-|\s)([a-z])')
 
 
 def contextfilter(f):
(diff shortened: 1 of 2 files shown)


ThiefMaster:
I don't think this is that useful in the core of Jinja...

Also, general advice for PRs and feature branches in general: Rebase them to upstream instead of merging upstream. Having merge commits in a PR's branch is rather ugly.

ThiefMaster:
Also, please do not perform formatting cleanups in unrelated PRs. As nice as having no trailing whitespace etc. is, it just adds noise to a PR whose objective is to add a new feature.

ThiefMaster:
> AbstractGeneric-factory   ➡️  GenericAbstractFactory

Did you mean `AbstractGeneric-factory   ➡️  AbstractGenericFactory`?

borismod:
@ThiefMaster thanks for your feedback. What alternative do you suggest instead of adding a filter when I need to camelize a string using Jinja?

ThiefMaster:
You can add it in your application. It does not need to be part of Jinja itself.

http://jinja.pocoo.org/docs/dev/api/#writing-filters

davidism:
I agree, adding this to your own app is trivial: install inflection, add relevant function to filters.

borismod:
Dear @davidism and @ThiefMaster thank you for your comments. I understand that this filter can be added in application that consumes Jinja. In my case it's [Cookiecutter ](https://github.com/audreyr/cookiecutter)- a command line tool for creating projects based on templates.  At the end we're using Cookiecutter in [ShellFoundry ](https://github.com/QualiSystems/shellfoundry)- a command line utility for creating CloudShell shells.

Both tools are not aware of the set of filters that Jinja provides and use them as a black box. That's the way we consume dependencies. Camelize is a popular string manipulation that may be used by many users of Jinja in creating templates for their project. It converts a snake style string to Camel Case, for example code_generator.py to CodeGenerator, i.e. from file name to class name.

Can you please reconsider your decision and accept this PR ❓

ThiefMaster:
It still sounds very code generation specific so me so it's something I'd rather to inside cookiecutter's code (i.e. do a PR against cookiecutter to add this, and possibly also the reverse, i.e. snakify)

PR #624, opened 2016-12-08
Add test for greater/less than

major:
It can be helpful to have a test for values that are greater or less than
another value, especially when looping over lists and using `select()` or
`selectattr()` filters. This patch adds tests that allow a developer to
check if a value is greater or less than another value.

For example, if I have a list of vehicles and I want to find the motorcycles:

    vehicles | selectattr('wheels', 'lessthan', 3)

Or, if I want to find large trucks:

    vehicles | selectattr('wheels', 'greaterthan', 4)

Tests are included.

### DIFF
--- a/jinja2/tests.py
+++ b/jinja2/tests.py
@@ -152,6 +152,16 @@ def test_escaped(value):
     return hasattr(value, '__html__')
 
 
+def test_greaterthan(value, other):
+    """Check if value is greater than other."""
+    return value > other
+
+
+def test_lessthan(value, other):
+    """Check if value is less than other."""
+    return value < other
+
+
 TESTS = {
     'odd':              test_odd,
     'even':             test_even,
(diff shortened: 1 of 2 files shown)


jeffwidman:
I'm in favor of this.

major:
Thanks, @jeffwidman. I couldn't figure out how to make the new tests show up in the documentation, though. Please let me know if I'm missing a step somewhere.

davidism:
All that should be required is modifying [`jinja2.tests.TESTS`](https://github.com/pallets/jinja/blob/6363243f3f71f3f4cf3ab2dac51b71a3bab37b1c/jinja2/tests.py#L155), which is [used by the `jinjatests` directive](https://github.com/pallets/jinja/blob/b89d1a8fe3fcbd73a8f4cebd4358eadebc2d8a9d/docs/jinjaext.py#L134), which you did. If rebuilding the docs doesn't render the new sections, then try wiping the `docs/_build` directory to do a clean build.

major:
Got it. Thanks, @davidism.

davidism:
So the docs are building correctly for you?

major:
@davidism: I figured out the issue. The jinja2 installation in my virtualenv was interfering with the documentation build. I installed the version of jinja from this PR and the docs built fine with the new tests. :)

major:
@davidism Do I need to do anything else on this PR? :)

major:
Thanks, @davidism! :)

PR #628, opened 2016-12-16
Uses re.compile flags argument, not inline flags

jfinkels:
This commit replaces the use of terminal inline flags in a regular expression in
`re.compile`,

    re.compile(r'\w+(?u)')

with arguments to the `re.compile` function itself,

    re.compile(r'\w+', re.UNICODE)

because the former is deprecated as of Python 3.6 (see https://bugs.python.org/issue22493).


jfinkels:
FYI This seems to allow all tests to pass with Travis Python versions "3.6-dev" and "nightly" (3.7).

PR #647, opened 2017-01-08
[WIP] Fail on DeprecationWarning

ThiefMaster:
need travis to test it, gentoo has no python 3.6 yet and i'm lazy ;)


ThiefMaster:
pytest is really weird with errors during test collection time. Having a look at this again in the future when I can test with 3.6 on a decent system (ie not windows) without spending time compiling python myself

PR #661, opened 2017-01-12
fix failing tests: removed some broken code

drewja:
This is the cause of failing tests. Attempt to overload methods on namedtuple with that of builtin tuple with attribute \_\_unicode__ is not needed. tuple and namedtuple do not have a \_\_unicode__ method

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -696,8 +696,6 @@ def do_round(value, precision=0, method='common'):
 _GroupTuple = namedtuple('_GroupTuple', ['grouper', 'list'])
 _GroupTuple.__repr__ = tuple.__repr__
 _GroupTuple.__str__ = tuple.__str__
-if not PY2:
-    _GroupTuple.__unicode__ = tuple.__unicode__
 
 @environmentfilter
 def do_groupby(environment, value, attribute):


jinja2/filters.py
```diff
@@ -696,8 +696,6 @@ def do_round(value, precision=0, method='common'):
 _GroupTuple = namedtuple('_GroupTuple', ['grouper', 'list'])
 _GroupTuple.__repr__ = tuple.__repr__
 _GroupTuple.__str__ = tuple.__str__
-if not PY2:
```
davidism: I think this was supposed to be `if PY2`.

jinja2/filters.py
```diff
@@ -696,8 +696,6 @@ def do_round(value, precision=0, method='common'):
...
-    _GroupTuple.__unicode__ = tuple.__unicode__
```
davidism: And this should have a corresponding `__unicode__ = __str__` in an `else`.

drewja: I think so too but it is still not necessarily, unicode(_GroupTuple('test', ('foo', 'bar'))) is still calling the overloaded __repr__

jinja2/filters.py
drewja: I think so too but it is still not necessarily, unicode(_GroupTuple('test', ('foo', 'bar'))) is still calling the overloaded repr.

drewja: also 'tuple' has no attribute '__unicode__'  by default

untitaker:
Please merge stuff like this into 2.9-maintenance, and that branch into master

Issue #681, opened 2017-02-08
Strip trailing whitespace from the `indent` filter


ypid:
## Expected Behavior

```
first.line

....third.line
```

## Actual Behavior

```
first.line
....
....third.line
```

## Template Code/MWE

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

"""first line

third line
"""

from jinja2 import Template

t = Template("{{ multline_string | indent(4) }}")
print(t.render(multline_string=__doc__).replace(" ", "."))
```

## Your Environment
* Python version: 2.7.9-1
* Jinja version: 2.8-1~bpo8+1

davidism:
closed by #685

PR #685, opened 2017-02-28
Ignore empty lines in indent filter #681

jkfran:
Ignore empty lines in indent filter #681

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -526,32 +526,44 @@ def do_urlize(eval_ctx, value, trim_url_limit=None, nofollow=False,
     return rv
 
 
-def do_indent(s, width=4, indent_first=False, indent_blank_lines=False, indentfirst=None):
-    """Return a copy of the passed string, each line indented by
-    4 spaces. The first line is not indented. If you want to
-    change the number of spaces or indent the first line too
-    you can pass additional parameters to the filter:
+def do_indent(
+    s, width=4, first=False, blank=False, indentfirst=None
+):
+    """Return a copy of the string with each line indented by 4 spaces. The
+    first line and blank lines are not indented by default.
 
-    .. sourcecode:: jinja
+    :param width: Number of spaces to indent by.
+    :param first: Don't skip indenting the first line.
+    :param blank: Don't skip indenting empty lines.
 
-        {{ mytext|indent(2, true) }}
-            indent by two spaces and indent the first line too.
+    .. versionchanged:: 2.10
+        Blank lines are not indented by default.
+
+        Rename the ``indentfirst`` argument to ``first``.
     """
     if indentfirst is not None:
-        warnings.warn('The use of indentfirst is obsolete. '
-            'You should use indent_first instead.', DeprecationWarning)
-        indent_first = indentfirst
-    s += '\n'  # this quirk is necessary for splitlines method
+        warnings.warn(DeprecationWarning(
+            'The "indentfirst" argument is renamed to "first".'
+        ), stacklevel=2)
+        first = indentfirst
+
+    s += u'\n'  # this quirk is necessary for splitlines method
     indention = u' ' * width
-    if indent_blank_lines:
+
+    if blank:
         rv = (u'\n' + indention).join(s.splitlines())
     else:
         lines = s.splitlines()
         rv = lines.pop(0)
+
         if lines:
-            rv += u'\n' + u'\n'.join(indention + line if line else line for line in lines)
-    if indent_first:
+            rv += u'\n' + u'\n'.join(
+                indention + line if line else line for line in lines
+            )
+
+    if first:
         rv = indention + rv
+
     return rv
 
 
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -11,6 +11,7 @@
 import re
 import math
 import random
+import warnings
 
 from itertools import groupby, chain
 from collections import namedtuple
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Could you please rebase the PR? Generally it's a good idea to use separate branches based off upstream/master for PRs instead of doing the changes in your own master branch.

While this can be done when merging a PR, sometimes people merge using a merge commit instead of rebase/squash and in that case the extra noise would remain in the upstream history forever.

jinja2/filters.py
```diff
@@ -457,7 +457,9 @@ def do_indent(s, width=4, indentfirst=False):
             indent by two spaces and indent the first line too.
     """
     indention = u' ' * width
-    rv = (u'\n' + indention).join(s.splitlines())
+    lines = s.splitlines()
+    rv = lines.pop(0) + u'\n'
+    rv += u'\n'.join([indention + l if l else l for l in lines])
```
ThiefMaster: lowercase-L is not a good choice (and actually disallowed as per PEP8, just like uppercase-i)

jkfran:
Hello @ThiefMaster thank you!
I'm sorry for the merge commits. What do you think now?

jinja2/filters.py
```diff
@@ -457,7 +457,9 @@ def do_indent(s, width=4, indentfirst=False):
             indent by two spaces and indent the first line too.
     """
     indention = u' ' * width
-    rv = (u'\n' + indention).join(s.splitlines())
+    lines = s.splitlines()
+    rv = lines.pop(0) + u'\n'
+    rv += u'\n'.join([indention + line if line else line for line in lines])
```
ThiefMaster: the `[]` isn't needed since there's only one argument to `join()`

jkfran: @ThiefMaster well it's Generator Expressions vs. List Comprehension, I was reading and you are right for these case Generator Expression is better.
http://stackoverflow.com/questions/47789/generator-expressions-vs-list-comprehension

jkfran: @ThiefMaster well it's Generator Expressions vs. List Comprehension, I was reading and you are right for these case Generator Expression is better.
http://stackoverflow.com/questions/47789/generator-expressions-vs-list-comprehension

jkfran:
Hello @ThiefMaster what do you think?

jkfran:
@ThiefMaster Excuse me but I need to do something? Close de PR or something? When it will be merged?
(Newbie sorry) :)

ThiefMaster:
Could you add an entry to the changelog file? The PR looks good so far (not a fan of the all-lowercase-no-underscore naming of the argument but the existing one uses the same scheme it's consistent).

jkfran:
@ThiefMaster Thank you! Everything is alright?

jkfran:
@ThiefMaster

davidism:
I'd like to see a more thorough test suite, it seems like we could be missing a lot of corner cases with this. Also, be sure to test the different options independently, don't just add the new behavior on to an existing test.

jinja2/filters.py
```diff
@@ -457,7 +457,12 @@ def do_indent(s, width=4, indentfirst=False):
             indent by two spaces and indent the first line too.
     """
     indention = u' ' * width
-    rv = (u'\n' + indention).join(s.splitlines())
+    if indentblanklines:
+        rv = (u'\n' + indention).join(s.splitlines())
+    else:
+        lines = s.splitlines()
+        rv = lines.pop(0) + u'\n'
```
davidism: This adds a trailing newline if there is only one line, unlike `indentblanklines=True`.

jinja2/filters.py
```diff
@@ -457,7 +457,12 @@ def do_indent(s, width=4, indentfirst=False):
...
+        rv += u'\n'.join(indention + line if line else line for line in lines)
```
davidism: This removes trailing blank lines, unlike `indentblanklines=False`.

tests/test_filters.py
```diff
@@ -126,11 +126,12 @@ def test_format(self, env):
         assert out == 'a|b'
 
     def test_indent(self, env):
-        tmpl = env.from_string('{{ foo|indent(2) }}|{{ foo|indent(2, true) }}')
-        text = '\n'.join([' '.join(['foo', 'bar'] * 2)] * 2)
+        tmpl = env.from_string('{{ foo|indent(2) }}|'
+                               '{{ foo|indent(2, true, true) }}')
```
davidism: Let's split each case into separate renders rather than separating them with `|`.

tests/test_filters.py
```diff
@@ -126,11 +126,12 @@ def test_format(self, env):
...
+        text = '\n'.join(['foo bar', '', 'foo baz'] * 2)
         out = tmpl.render(foo=text)
-        assert out == ('foo bar foo bar\n  foo bar foo bar|  '
-                       'foo bar foo bar\n  foo bar foo bar')
+        assert out == ('foo bar\n\n  foo baz\n  foo bar\n\n  foo baz|  '
+                       'foo bar\n  \n  foo baz\n  foo bar\n  \n  foo baz')
```
davidism: This doesn't cover `indent(2, False, True)` or `indent(2, True, False)`.

jinja2/filters.py
```diff
@@ -445,7 +445,7 @@ def do_urlize(eval_ctx, value, trim_url_limit=None, nofollow=False,
     return rv
 
 
-def do_indent(s, width=4, indentfirst=False):
+def do_indent(s, width=4, indentfirst=False, indentblanklines=False):
```
davidism: I know @ThiefMaster said it's consistent with `indentfirst`, but I'd rather just break consistency and use the easier to read `indent_blank_lines`.

davidism: If you want to, also change the signature to `do_indent, s, width=4, indent_first=False, indent_blank_lines=False, indentfirst=None)` and raise a deprecation warning if `indentfirst` is not None. Not required though, I can follow up with this after merging.

PR #690, opened 2017-03-09
Consider xml-escaped templates

maxrosenblattl:
Add xml-entity-operators _after_ creation of reverse_operators since those
escapings are only necessary for lexing, not for reverse interpretation.
The change allows usage and editing of jinja2-Templates in applications
that use XML/HTML data, such as Inkscape for SVG.

### DIFF
--- a/jinja2/lexer.py
+++ b/jinja2/lexer.py
@@ -31,7 +31,14 @@ _lexer_cache = LRUCache(50)
 # static regular expressions
 whitespace_re = re.compile(r'\s+', re.U)
 string_re = re.compile(r"('([^'\\]*(?:\\.[^'\\]*)*)'"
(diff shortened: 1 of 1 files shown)


ThiefMaster:
This sounds like a change that would break many use cases. If I have something that looks like entities in my template I surely do not want Jinja to convert them to anything else.

maxrosenblattl:
You are right, use cases like

`{{ foo(bar="foo&quot;bar") }}`

Will break with the current state, this needs some rework.

The use case for the request is actually the other way round: it is about creation of templates in the application, not using rendered output.

ThiefMaster:
That doesn't seem like something for the Jinja core. Maybe you can use a Jinja extension to preprocess your templates and do this conversion?

PR #692, opened 2017-03-09
Dictsort reverse

dnmvisser:


ThiefMaster:
Let's add a test for this in `test_dictsort`. It's just two lines :p
Also, mention it in CHANGES.

Oh, and it'd be great if you could rebase against the latest master and squash the commits into a single one. While this can be done while merging it's easy to forget, which results in a somewhat ugly git log.

PR #693, opened 2017-03-17
add examples showing > 1 argument in filter

tedder:
This is implied but poorly documented. A few examples make it clear.

The first example, "select odd", was already duplicated- I don't know why.


davidism:
I'd like to get #665 cleaned up and merged so that we can show `<` and `==` (or `lt` and `eq`) instead of the long names. But I do like this overall, so I'll merge it now.

PR #700, opened 2017-04-05
Fixed incorrect merge in setup.py

brunodoamaral:
I just don't know if the correct version is 2.9.7.dev or 2.10.dev

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,11 +40,7 @@ from setuptools import setup
 
 setup(
     name='Jinja2',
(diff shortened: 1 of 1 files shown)


davidism:
The version should be 2.10.dev on master, 2.9.7.dev on 2.9-maintenance.

ThiefMaster:
Maybe we should extract the version from `jinja2/__init__.py` like it's done e.g. in flask...

PR #745, opened 2017-07-14
`_compat.py` fails flake8 tests on Python 3

cclauss:
`_compat.py` fails flake8 tests on Python 3 so other projects that replicate `_compat.py` also fail flake8 tests on Python 3.

`# stop the build if there are Python syntax errors or undefined names`
$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__

```
./jinja2/_compat.py:51:17: F821 undefined name 'unicode'
    text_type = unicode
                ^

./jinja2/_compat.py:52:18: F821 undefined name 'xrange'
    range_type = xrange
                 ^

./jinja2/_compat.py:53:26: F821 undefined name 'unicode'
    string_types = (str, unicode)
                         ^

./jinja2/_compat.py:54:27: F821 undefined name 'long'
    integer_types = (int, long)
                          ^

./jinja2/_compat.py:80:33: F821 undefined name 'unicode'
        if isinstance(filename, unicode):
                                ^
```

### DIFF
--- a/jinja2/_compat.py
+++ b/jinja2/_compat.py
@@ -14,19 +14,35 @@ import sys
 
 PY2 = sys.version_info[0] == 2
 PYPY = hasattr(sys, 'pypy_translation_info')
-_identity = lambda x: x
+_identity = lambda x: x  # noqa: E731
 
+
+# avoid flake8 F821 undefined name 'long'
+try:
+    integer_types = (int, long)  # Python 2
+except NameError:
+    integer_types = (int, )      # Python 3
+
+# avoid flake8 F821 undefined name 'unicode'
+try:
+    text_type = unicode  # Python 2
+    string_types = (str, unicode)
+except NameError:
+    text_type = str      # Python 3
+    string_types = (str, )
+
+# avoid flake8 F821 undefined name 'xrange'
+try:
+    range_type = xrange  # Python 2
+except NameError:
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Shouldn't this whole file be excluded from flake8 tests?

cclauss:
Flake8 can catch issues that are useful to know about when submitting pull requests.  Taking the blanket approach hides those issues from the contributor.  Taking the targeted approach (`noqa: F821`) is more verbose but it signals a conscious choice ("Kinda like: Yes, I am over 18, I know what I am doing...")

jinja2/_compat.py
```diff
@@ -76,10 +87,15 @@ def implements_to_string(cls):
...
 
-    def encode_filename(filename):
+
+def encode_filename(filename):
+    # avoid flake8 F821 undefined name 'unicode'
+    try:               # Python 2
         if isinstance(filename, unicode):
             return filename.encode('utf-8')
-        return filename
+    except NameError:  # Python 3
```
davidism: This incurs `except` overhead *every time* the function is called in Python 3.

davidism:
I don't find the changes here compelling. If other projects want a `_compat` that passes Flask8, but want to copy this one, they can do that however they desire after the copy.

PR #766, opened 2017-09-07
Let wordwrap filter respect existing newlines

### LINKED ISSUE #175: wordwrap filter ignores existing newlines
Hi there,

I've noticed that when using this filter to (for example) ensuring an email message is always wrapped to 72 characters (where possible) to ensure a nice display of it as often as possible, we noticed that wordwrap would not use existing newlines in the message as clue but would instead insert newlines strictly every 72 characters.

This is our workaround

``` python
## Workaround bug in do_wordwrap where it disregards existing linebreaks 
## when wrapping text

from jinja2.filters import environmentfilter
import re
@environmentfilter
def do_wordwrap(environment, s, width=79, break_long_words=True):
    """
    Return a copy of the string passed to the filter wrapped after
    ``79`` characters.  You can override this default using the first
    parameter.  If you set the second parameter to `false` Jinja will not
    split words apart if they are longer than `width`.
    """
    import textwrap
    accumulator = []
    # Workaround: pre-split the string
    for component in re.split(r"\r?\n", s):
        # textwrap will eat empty strings for breakfirst. Therefore we route them around it.
        if len(component) is 0:
            accumulator.append(component)
            continue
        accumulator.extend(
            textwrap.wrap(component, width=width, expand_tabs=False,
                replace_whitespace=False,
                break_long_words=break_long_words)
        )
    return environment.newline_sequence.join(accumulator)
```

Not nice and complete but works for us.

I'd like this (or something similar) to be included in jinja2 proper, as it makes the wordrwap filter much more usefull.

What do you think?

fheinle:
`textwrap.wrap()` from stdlib has unexpected behaviour where when
wrapping multiple paragraphs it will not consider existing newlines.

I.e. when your first paragraph ends on col 75, the next paragraph
will be wrapped on col 5 already.

This patch is wrapping each line individually and combining it back
together.

Fixes #175.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -62,6 +62,9 @@ Unreleased
     the correct queue. :issue:`843`
 -   Compiling templates always writes UTF-8 instead of defaulting to the
     system encoding. :issue:`889`
+-   ``|wordwrap`` filter treats existing newlines as separate paragraphs
+    to be wrapped individually, rather than creating short intermediate
+    lines. :issue:`175`
 
 
 Version 2.10.3
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -700,9 +700,23 @@ def do_wordwrap(environment, s, width=79, break_long_words=True,
     if not wrapstring:
         wrapstring = environment.newline_sequence
     import textwrap
-    return wrapstring.join(textwrap.wrap(s, width=width, expand_tabs=False,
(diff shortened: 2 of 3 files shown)


pief:
Any chance we can see this merged?

dwt:
I like it. @maintainers - is there any reason not to include this change?

PR #785, opened 2017-11-17
Loader for Python modules

TomekTrzeciak:
Load Python modules as Jinja2 templates.
```
        >>> loader = PyModuleLoader()
```
This loader returns an empty template exporting Python module's namespace. It piggybacks on the Jinja2 import mechanism to provide a convienient way to access functionality from any Python module in ``sys.path``.

### DIFF
--- a/jinja2/loaders.py
+++ b/jinja2/loaders.py
@@ -479,3 +479,48 @@ class ModuleLoader(BaseLoader):
 
         return environment.template_class.from_module_dict(
             environment, mod.__dict__, globals)
+
+
+class PyModuleLoader(BaseLoader):
+    """Loads Python modules as Jinja2 templates.
+
+        >>> loader = PyModuleLoader()
+
+    This loader returns an empty template exporting Python module's namespace.
+    It piggybacks on the Jinja2 import mechanism to provide a convienient way
+    to access functionality from any Python module in ``sys.path``.
+    """
+
+    # no source access for this loader
+    has_source_access = False
+
+    def __init__(self, prefix='__python__'):
+        self._templates = {}
+        # prefix that can be used to avoid name collisions with template files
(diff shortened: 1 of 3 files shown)


ThiefMaster:
What's the benefit of it? 

Without an example (your unit test doesn't qualify IMO) it reminds me of this:

useless machine

ThiefMaster:
I really don't think that exposing the whole python library via a template loader is even close to a good idea, even less something that should be in the Jinja core

TomekTrzeciak:
The main motivation is to be able to easily import additional functionality from Python modules using the same interface as for importing variables/macros from other templates, e.g.:
```
{% from "itertools" import izip %}
```
My particular use case is templating configuration files of [Cylc workflow engine](https://github.com/cylc/cylc). To use Cylc users are generally assumed to already have full shell access, so exposing whole Python library is really of no concern. Cylc already exposes adding user defined Jinja2 filters, but it's tedious to write filter wrappers for every little scrap of functionality.

From your comment it's quite clear that you're not very keen on the idea, so arguing my point is most likely moot. But let me just point out that Jinja is used also in systems, where tight control over exposed functionality is not required, so I think this PR could be of broader appeal than my particular use case. Anyway, I rest my case ;-) Feel free to reject this PR if you think it's a bad idea.

ThiefMaster:
I think it makes more sense to add this loader in your application then.

FYI, if you want to expose modules, you can pass e.g. `itertools=itertools` as a kwarg when rendering the template.

davidism:
This is clever, but it seems way too easy for people to shoot themselves in the foot with it. I'm sure people would take advantage of a `jinja-import-loader` if you released this separately. It doesn't look like there's anything here that would benefit from being in core.

kblomqvist:
I would take advantage of a jinja-import-loader in my project (yasha).

PR #796, opened 2018-01-10
Update line statement lexer regexp to stop at the first newline

captainmalloc:
When using line statement prefix in Jinja2 templates it is impossible to add a blank line after the Jinja2 tag because successive newlines are recognized as TOKEN_LINESTATEMENT_END. 

The following example:
```
# for i in range(2)
#  for j in range(2)
{{i}}_{{j}}
#  endfor

# endfor
```
Will output:
```
0_0
0_1
1_0
1_1
```

A workaround is to add an explicit newline:
```
# for i in range(2)
#  for j in range(2)
{{i}}_{{j}}
#  endfor
{{'\n'}}
# endfor
```

With this simple fix, the lexer will recognize the end of the line statement at the first newline and allow the user to put empty lines as if (s)he is using blocks.

The given example will respect user wish to insert an empty line after each inner loop by generating the following output:
```
0_0
0_1

1_0
1_1

```

### DIFF
--- a/jinja2/lexer.py
+++ b/jinja2/lexer.py
@@ -536,7 +536,7 @@ class Lexer(object):
             ],
             # line statements
             TOKEN_LINESTATEMENT_BEGIN: [
-                (c(r'\s*(\n|$)'), TOKEN_LINESTATEMENT_END, '#pop')
+                (c(r'[ \t\r\f\v]*(\n|$)'), TOKEN_LINESTATEMENT_END, '#pop')
             ] + tag_rules,
             # line comments
             TOKEN_LINECOMMENT_BEGIN: [
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -161,7 +161,7 @@ class TestBug(object):
 #   print 'D'
 # endif
     ''')
-        assert t.render(a=0, b=False, c=42, d=42.0) == '1111C'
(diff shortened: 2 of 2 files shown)


davidism:
Did you mean to include your other PR in this PR? I don't want to deal with merge conflicts.

captainmalloc:
I didn't mean to add those commits. I am sorry about that, I'll clean it on a new commit.

davidism:
You can `git rebase -i master` then `git push -f` to clean it up.

captainmalloc:
Done. Sorry for the inconvenience.

davidism:
The fact that line statements behave the same as `-%}`  seems intentional. If you need to only strip one line, use `trim_blocks=True` with regular (not line) statements.

captainmalloc:
IMHO line statement should be considered as "line" statement. There is not real reason to make them consume multiple lines when parsed.

This is why I am not  satisfied with your explanation and your decision to reject the this request.

samuelcolvin:
I know this is an old PR, but I'll like to add my voice to support this PR.

Line statements looked like a create solution for my application of templated markdown files use by non technical people. 

But this problem means that they're basically useless for this application since text inside an `if` statement can never be its own paragraph.

samuelcolvin:
surely stripping newlines after line statements should follow `trim_blocks` `True/False`

PR #797, opened 2018-01-11
Pass python_requires argument to setuptools

jdufresne:
Helps pip decide what version of the library to install.

https://packaging.python.org/tutorials/distributing-packages/#python-requires

> If your project only runs on certain Python versions, setting the
> python_requires argument to the appropriate PEP 440 version specifier
> string will prevent pip from installing the project on other Python
> versions.

https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords

> python_requires
>
> A string corresponding to a version specifier (as defined in PEP 440)
> for the Python version, used to specify the Requires-Python defined in
> PEP 345.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,7 @@ setup(
         'Topic :: Text Processing :: Markup :: HTML'
     ],
     packages=['jinja2'],
+    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     install_requires=['MarkupSafe>=0.23'],
     extras_require={'i18n': ['Babel>=0.8']},
     include_package_data=True,


davidism:
I don't think any version of pip that supports Python requires supports 2.6, so you can remove that. Not sure if it supports 3.3, but might as well drop that too since we're going to do that anyway. Can always use `--ignore-requires-python` if there's an issue.

jdufresne:
In that case, shall I remove Python 2.6 & 3.3 from the trove classifiers and the testing matrix? That would signal to users that 2.6 & 3.3 are not intentionally or directly supported by the Jinja2 project. What do you think?

Here are the pip downloads from Jinja2 over the last 30 days:

```
$ pypinfo --percent --pip jinja2 pyversion
| python_version | percent | download_count |
| -------------- | ------- | -------------- |
| 2.7            |   58.7% |      2,109,719 |
| 3.6            |   18.9% |        679,141 |
| 3.5            |   14.0% |        504,670 |
| 3.4            |    7.5% |        270,369 |
| 2.6            |    0.6% |         20,313 |
| 3.3            |    0.2% |          6,765 |
| 3.7            |    0.1% |          5,322 |
| 3.2            |    0.0% |            289 |
| None           |    0.0% |             26 |
```

jdufresne:
I've made the requested change.

jdufresne:
Removing 2.6 & 3.3 caused the tests to fail. Both in tox and Travis. It failed to install the Jinja package with 

```
Jinja2 requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*' but the running Python is 2.6.9
```

Until Jinja officially drops support for 2.6 & 3.3. This line should allow them. I've updated the PR to include them again.

jdufresne:
I have rebased this on the latest master. Now that Python 2.6 and 3.3 have been dropped, I think this should be reconsidered.

- Dropped 2.6: 5425921899f7c84b3a08e2dd06610beacb2fc20e
- Dropped 3.3: d1fb22fd45ad6d0caa314fe373a377fbbf82ba08

Issue #174, opened 2013-01-17
Implement builtin template helper to dump available template variables


techtonik:
Smarty template engine has a very convenient [{debug}](http://www.smarty.net/docs/en/language.function.debug.tpl) built-in that gives a quick overview of variables available for designer.

![Smarty_debug_console](https://f.cloud.github.com/assets/515889/74069/73886ee8-6077-11e2-96e3-41808672d2fe.jpg)

It would be nice to have the same stuff in Jinja2 [by default](http://stackoverflow.com/questions/3398850/how-to-get-a-list-of-current-variables-from-jinja-2-template)

hvisage:
+1 :)

Moini:
+1 from me, too (would like to have it available in pelican)

westurner:
I'd have to check, but I think this should work:

```
<h2>locals()</h2>
<pre>{{ locals() }}</pre>
<h2>globals()</h2>
```

... https://stackoverflow.com/questions/3398850/how-to-get-a-list-of-current-variables-from-jinja-2-template

@mitsuhiko from https://stackoverflow.com/a/13757358/188833 : 

    import jinja2
    
    @jinja2.contextfunction
    def get_context(c):
	return c
    
    app.jinja_env.globals['context'] = get_context
    app.jinja_env.globals['callable'] = callable

for debugging, e.g. pyramid_debugtoolbar can show the template context if the view function -- appropriately abstracted IMHO -- just returns a context object (e.g. dict, OrderedDict)

* https://github.com/Pylons/pyramid_debugtoolbar
* https://westurner.org/wiki/awesome-python-testing#pyramid

## use case
My use case here is passing a few args from a CLI utility through to a jinja2 template (where I'd like to define defaults within the template (e.g. for Dockerfiles (without ``autoescape``, of all things))

What I'd like to do is

```jinja
{% set ENVVAR=context.get('ENVVAR', 'DEFAULT') %}
```

but, alas, the actual ``context`` object is [...]

techtonik:
@westurner, so does it work? I am unable to check it right now, but it looks to me that `<pre>{{ locals() }}</pre>` should fail if it contains non-convertible bytes.

westurner:
* ``json.dumps(locals())`` won't work without a custom JSONEncoder (with eg ``repr()`` for [code objects,])
* maybe ``pprint.pformat(locals())``? (this would be a convenient builtin)

westurner:
[...]

PR #798, opened 2018-01-14
Add support for a {% debug %} extension tag.

ShaheedHaque:
This dumps the available variables, filters and tests. See Issue #174.
This is roughly equivalent to the Django Template Language {% debug %} tag, and typical usage like this:

`
        {% debug %}
`

produces output like this:

```
        {'context': {'_': <function _gettext_alias at 0x7f9ceabde488>,
                 'csrf_token': <SimpleLazyObject: 'lfPE7al...q3bykS4txKfb3'>,
                 'cycler': <class 'jinja2.utils.Cycler'>,
                 ...
                 'view': <polls.views_auth.Login object at 0x7f9cea2cbe48>},
        'filters': ['abs', 'add', 'addslashes', 'attr', 'batch', 'bootstrap',
                 'bootstrap_classes', 'bootstrap_horizontal',
                 'bootstrap_inline', ... 'yesno'],
        'tests': ['callable', 'checkbox_field', 'defined', 'divisibleby',
               'escaped', 'even', 'iterable', 'lower', 'mapping',
               'multiple_checkbox_field', ... 'string', 'undefined', 'upper']}
```

The output under Python2 and Python3 below 3.4 is slightly less compact, as pprint.format() did not introduce the 'compact' argument before 3.4.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,16 @@ unreleased
 .. _#765: https://github.com/pallets/jinja/issues/765
 
 
+Version 2.10.1
+--------------
+
+Released 2019-04-06
+
+-   ``SandboxedEnvironment`` securely handles ``str.format_map`` in
+    order to prevent code execution through untrusted format strings.
+    The sandbox already handled ``str.format``.
+
+
 Version 2.10
 ------------
 
--- a/jinja2/sandbox.py
+++ b/jinja2/sandbox.py
@@ -135,7 +135,7 @@ class _MagicFormatMapping(abc.Mapping):
 def inspect_format_method(callable):
     if not isinstance(callable, (types.MethodType,
                                  types.BuiltinMethodType)) or \
-       callable.__name__ != 'format':
+       callable.__name__ not in ('format', 'format_map'):
         return None
     obj = callable.__self__
     if isinstance(obj, string_types):
@@ -400,7 +400,7 @@ class SandboxedEnvironment(Environment):
             obj.__class__.__name__
         ), name=attribute, obj=obj, exc=SecurityError)
 
-    def format_string(self, s, args, kwargs):
+    def format_string(self, s, args, kwargs, format_func=None):
         """If a format call is detected, then this is routed through this
         method so that our safety sandbox can be used for it.
         """
@@ -408,6 +408,17 @@ class SandboxedEnvironment(Environment):
             formatter = SandboxedEscapeFormatter(self, s.escape)
         else:
             formatter = SandboxedFormatter(self)
+
+        if format_func is not None and format_func.__name__ == 'format_map':
+            if len(args) != 1 or kwargs:
+                raise TypeError(
+                    'format_map() takes exactly one argument %d given'
+                    % (len(args) + (kwargs is not None))
+                )
+
+            kwargs = args[0]
+            args = None
+
         kwargs = _MagicFormatMapping(args, kwargs)
         rv = formatter.vformat(s, args, kwargs)
         return type(s)(rv)
@@ -416,7 +427,7 @@ class SandboxedEnvironment(Environment):
         """Call an object from sandboxed code."""
         fmt = inspect_format_method(__obj)
         if fmt is not None:
-            return __self.format_string(fmt, args, kwargs)
+            return __self.format_string(fmt, args, kwargs, __obj)
 
         # the double prefixes are to avoid double keyword argument
         # errors when proxying the call.
--- a/tests/test_security.py
+++ b/tests/test_security.py
@@ -187,3 +187,22 @@ class TestStringFormat(object):
         env = SandboxedEnvironment()
         t = env.from_string('{{ ("a{0.foo}b{1}"|safe).format({"foo": 42}, "<foo>") }}')
         assert t.render() == 'a42b&lt;foo&gt;'
+
+
+@pytest.mark.sandbox
+@pytest.mark.skipif(not hasattr(str, 'format_map'), reason='requires str.format_map method')
+class TestStringFormatMap(object):
+    def test_basic_format_safety(self):
+        env = SandboxedEnvironment()
+        t = env.from_string('{{ "a{x.__class__}b".format_map({"x":42}) }}')
+        assert t.render() == 'ab'
+
+    def test_basic_format_all_okay(self):
+        env = SandboxedEnvironment()
+        t = env.from_string('{{ "a{x.foo}b".format_map({"x":{"foo": 42}}) }}')
+        assert t.render() == 'a42b'
+
+    def test_safe_format_all_okay(self):
+        env = SandboxedEnvironment()
+        t = env.from_string('{{ ("a{x.foo}b{y}"|safe).format_map({"x":{"foo": 42}, "y":"<foo>"}) }}')
+        assert t.render() == 'a42b&lt;foo&gt;'


davidism:
It would be interesting to support syntax like `{% debug html-table %}` to produce a table when you know you're in an HTML template.

ShaheedHaque:
To be honest, that seems like overkill given that this is a debug facility intended to efficiently serve the all-too-common scenario of "Hmmm, I'm not quite sure what all the context variables are" using built-in Python string formatting. If table support was needed, perhaps it would be better to introduce some kind of a magic {{ debug }} which returned the equivalent dict, and allowed the user to select what was needed and how it should be styled?

davidism:
That's the other thing I was thinking. Using a tag means you can't access the data directly, unless you were to turn it into a block tag like `call`.

ShaheedHaque:
True, but TBH, that sounds like a whole different thing: its not a tag for the reason you suggest, nor is the idea of a "magic" context variable that contains all other context variables/filters/tests a "nice" thing. That all but calls for new syntax {{ %reserved-thing% }}, for example. I believe the requirement being addressed here is much simpler than full scale access to that data.

ShaheedHaque:
If you'd prefer not to accept this PR, please do let me know (the downstream project I'd originally submitted this to might be an alternative home).

davidism:
I'd like to hear from @mitsuhiko or @ThiefMaster before merging.

In order to use it, you need to know about and enable the extension first, add a tag, and you only get a squashed, non-interactive representation that isn't JSON compatible.

There are already things like [Flask-DebugToolbar](https://github.com/mgood/flask-debugtoolbar). The user can also set a breakpoint in a debugger and look at the context directly:

pycharm-jinja-debug

jinja2/ext.py
```diff
@@ -434,6 +446,57 @@ class AutoEscapeExtension(Extension):
     pass
 
 
+class DebugExtension(Extension):
+    """
+    A {% debug %} tag that dumps the available variables, filters and tests.
```
davidism: Use inline code markup

    ``{% debug %}``

jinja2/ext.py
```diff
@@ -434,6 +446,57 @@ class AutoEscapeExtension(Extension):
...
+    This is roughly equivalent to the Django Template Language {% debug %} tag,
```
davidism: No need to mention Django here, this won't mean anything if you haven't used the Django debug tag, and it's not really equivalent anyway.

jinja2/ext.py
```diff
@@ -434,6 +446,57 @@ class AutoEscapeExtension(Extension):
...
+    and typical usage like this:
+
+        <pre>{% debug %}</pre>
```
davidism: Code blocks must be preceded with two colons: `::` or (preferably with non-Python) `.. codeblock:: html+jinja`.

jinja2/ext.py
```diff
@@ -434,6 +446,57 @@ class AutoEscapeExtension(Extension):
...
+        result = {
+            'filters': sorted(self.environment.filters.keys()),
+            'tests': sorted(self.environment.tests.keys()),
+            'context': context.get_all()
+        }
+        #
+        # We set the depth since the intent is basically to show the top few
+        # names. TODO: provide user control over this?
+        #
+        if (version_info[0], version_info[1]) >= (3,4):
```
davidism: `version_info[:2]`

tests/test_ext.py
```diff
@@ -240,6 +240,22 @@ class T2(Extension):
         assert ext[0].__class__ is T1
         assert ext[1].__class__ is T2
 
+    def test_debug(self):
+        """Test for {% debug %}"""
+        env = Environment(extensions=['jinja2.ext.debug'])
+        tmpl = env.from_string('''Hello{% debug %}Bye''')
+        out = tmpl.render()
+        out = out.replace('&#x27;', "'").replace('&lt;', '<').replace('&gt;', '>')
```
davidism: Why replace these?

ShaheedHaque:
Of course, it is completely reasonable that you and the community must feel comfortable with what is proposed. If its not the right thing for Jinja, so be it. In the meantime, thanks for the excellent feedback, I will address the points noted.

On the general comments...

- about this being an extension. Yes, I packaged it that way because I thought that seemed like a plausible approach. I'm fine if you prefer to make it part of the default tag set.
- about the output not being JSON, I should just point out that the data exposed is commonly not vanilla data, at least in the Django context. There *are* ellipses present, but that's only because I specifically limited the output depth. If you prefer, I could remove that limit. Barring that, it looks like JSON to me...though I've not reviewed how pprint works..is there some case you have in mind?
- about the output not being interactive. I'm not sure quite what you had in mind, but interaction was not something I think belongs in this context. 
- about alternate debug facilities. I use PyCharm too, but it turns out that at least for Django-users, breakpoints are not supported with Jinja...only DTL is [supported](https://www.jetbrains.com/help/pycharm/debugging-django-templates.html). There are also plenty of scenarios where I think a simple dump is useful.

jinja2/ext.py
ShaheedHaque: Ack.

jinja2/ext.py
ShaheedHaque: Ack.

jinja2/ext.py
ShaheedHaque: I had not appreciated the code was being processed for markup. I will look into this.

jinja2/ext.py
ShaheedHaque: Duh.

tests/test_ext.py
ShaheedHaque: If an assert fails, the debug output is much easier to read, and so the tests are easier to maintain.

jinja2/ext.py
```diff
@@ -4,13 +4,24 @@
...
+        - a {% debug %} tag
 
     :copyright: (c) 2017 by the Jinja Team.
     :license: BSD.
 """
+try:
+    # Python3
+    from html import escape
+except ImportError:
+    from cgi import escape
```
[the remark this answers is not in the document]
ShaheedHaque: Thanks for the tip. I'll look into it.

davidism:
JSON uses double quotes around strings. Values that aren't JSON types should be converted to string (this is not trivial).

Interactive as in I can access it as a varaible and render it how I want.

PyCharm says it supports debugging Jinja templates, so if that isn't the case it needs to be reported to PyCharm. Are you sure you selected "Jinja2" for the template language in the project? Are you sure that the file is detected as a Jinja template (it can be changed in the file association settings)?

ShaheedHaque:
I think I've addressed the comments, but it is not clear to me how to verify the markup-related suggestions (I had to work around an unrelated doc build failure locally, but the Travis build fails in some way that seems likely related to my change...I'll have to experiment to fix it).

- On the JSON format. I'm happy to look at that if that becomes the blocking issue.
- On the interactive thing. I have no idea how to implement that and it is not of interest to me.
- On the PyCharm support. The link I provided is to the official PyCharm docs: the combination of Django and Jinja is not expected to work.

jinja2/ext.py
davidism: Pygments recognizes `html+jinja`, use that over `html`.

jinja2/ext.py
```diff
@@ -434,6 +442,59 @@ class AutoEscapeExtension(Extension):
...
+    """
+    A ``{% debug %}`` tag that dumps the available variables, filters and tests.
+    Typical usage like this:
+
+    .. codeblock:: html
+        <pre>{% debug %}</pre>
+
+    produces output like this:
+
+    ``
```
davidism: Use `::` for Python code blocks. Double backticks are only for inline code.

ShaheedHaque: OK. Can you educate me on where to look to see what is produced? I've tried running sphinx by hand, and searched in vain for where this stuff ends up under the docs/_build tree. (I'm a sphinx newb).

ShaheedHaque:
I've no idea why the doc build is failing under Travis:

> Warning, treated as error:
/home/travis/build/pallets/jinja/docs/templates.rst:1314:Could not lex literal_block as "jinja". Highlighting skipped.
ERROR: InvocationError: '/home/travis/build/pallets/jinja/.tox/docs-html/bin/sphinx-build -W -b html -d /home/travis/build/pallets/jinja/.tox/docs-html/tmp/doctrees docs docs/_build/html'

I *do* get a different error locally, but if I make this change on about line 307 of docs/extensions.rst:

> \-   \:members\: push, look, eos, skip, \_\_next\_\_, next_if, skip_if, expect
> \+   \:members\: push, look, eos, skip, next_if, skip_if, expect

then I get a clean local build. I don't see how this can be related to my changes, so I fear I'm a bit stuck.

ShaheedHaque:
I note that PR #802 has the same doc build failure.

dwt:
@maintainers: any progress on this? I've just been searching for this again too.

davidism:
Continued and merged in #983.

PR #799, opened 2018-01-17
Allow to pass a list to native_concat

mkrizek:
This fixes the issue mentioned in https://github.com/ansible/ansible/pull/32738#issuecomment-358010879:
```
from jinja2 import Environment, BaseLoader
from jinja2.nativetypes import NativeEnvironment

data = "[{{ 'all' }}]"
loader = BaseLoader()
t = Environment(loader=loader).from_string(data)
t_native = NativeEnvironment(loader=loader).from_string(data)
print t.render()
print t_native.render()
```
Before:
```
[all]
[all[all]
```
After:
```
[all]
[all]
```

### DIFF
--- a/jinja2/nativetypes.py
+++ b/jinja2/nativetypes.py
@@ -21,12 +21,11 @@ def native_concat(nodes):
     if not head:
         return None
 
-    if isinstance(nodes, types.GeneratorType):
-        nodes = chain(head, nodes)
-
     if len(head) == 1:
         out = head[0]
     else:
+        if isinstance(nodes, types.GeneratorType):
+            nodes = chain(head, nodes)
         out = u''.join([text_type(v) for v in nodes])
 
     try:
--- a/jinja2/nativetypes.py
+++ b/jinja2/nativetypes.py
@@ -1,4 +1,5 @@
 import sys
+import types
 from ast import literal_eval
 from itertools import islice, chain
 from jinja2 import nodes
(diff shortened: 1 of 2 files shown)


jinja2/nativetypes.py
```diff
@@ -20,10 +21,13 @@ def native_concat(nodes):
     if not head:
         return None
 
+    if isinstance(nodes, types.GeneratorType):
```
davidism: `collections.Generator` is probably more correct here.

davidism: Let's keep this in the `else` block, no need to run the check if `len(head) == 1`.

mkrizek: Looks like `collections.abc.Generator` [1] is new in Python 3.5 though. We'd need to use `types.GeneratorType` for previous versions?

[1] https://docs.python.org/3/library/collections.abc.html#collections.abc.Generator

mkrizek: :+1:

mkrizek:
@davidism anything else you want me to change here?

davidism:
Thanks! I'm not sure exactly when I can make a new release, but let me know if Ansible's timeline is affected by this.

PikachuEXE:
@davidism 
Hi, I raised an issue https://github.com/ansible/ansible/issues/49012
And it seems that I am supposed to use `master` of the repo
Will there be a release soon?

PR #802, opened 2018-01-29
More closely match python splatting for calls

abadger:
In Python, it's legal to use splats in function calls like this:

    call_function('123', *b, c=5, **d)

Prior to this change, jinja2 did not allow that. It mandated that all
splatting had to take place at the end of the call:

    call_function('123', c=5, *b, **d)

This commit allows both orders.

Additionally, this commit adds tests for more things that are illegal:
namely giving a function a positional argument after using a splat to
unpack positional arguments or giving a function a named argument after
using a double splat to unpack keyword arguments.

### DIFF
--- a/jinja2/parser.py
+++ b/jinja2/parser.py
@@ -777,16 +777,18 @@ class Parser(object):
                 next(self.stream)
                 dyn_kwargs = self.parse_expression()
             else:
-                ensure(dyn_args is None and dyn_kwargs is None)
                 if self.stream.current.type == 'name' and \
                    self.stream.look().type == 'assign':
+                    # Parsing a kwarg
+                    ensure(dyn_kwargs is None)
                     key = self.stream.current.value
                     self.stream.skip(2)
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Only recent Python 3 versions allow this natively. Jinja templates should work regardless of the Python version they are used in, so I don't think such a change makes any sense before Python 2 is EOL and support for it is dropped in Jinja - emulating it would just be lots of work for little gain.

abadger:
@ThiefMaster That's not true.  I checked this behaviour on Python-2.6 and Python-2.7 before submitting this PR:

```
$ cat test.py
def testing(a, b, c=None, d=None):
    print(a)
    print(b)
    print(c)
    print(d)

b = [4]
d = {'d': 6}
testing('123', *b, c=5, **d)

import sys
print(sys.version_info)

$ python2 test.py
123
4
5
6
(2, 6, 6, 'final', 0)
```

PR #807, opened 2018-02-07
Rewrite default filter docs

chamini2:
I think this way is a little bit clearer.


chamini2:
@davidism, what do you think?

davidism:
Thanks, but this doesn't seem to improve what's being said, it just moves it around.

PR #826, opened 2018-03-19
Fix escaped characters in indent filter

### LINKED ISSUE #823: Incorrect behavior of indent filter
## Expected Behavior
I expect that my code generates following text:
```
  "A"
  "B"

  "A"
  "B"
  
  "A"
  "B"

  "A"
  "B"

```

## Actual Behavior
Code generates following
```
  "A"
  "B"

  "A"
  "B"
  
  "A"
  &#34;B&#34;

  "A"
  "B"
  
```

## Template Code
```python
import jinja2

env = jinja2.Environment()

text = '"A"\n"B"\n'

print env.from_string('{{ t | indent(2, True) }}').render(t=text)
print env.from_string('{{ t | indent(2, True, True) }}').render(t=text)

print env.from_string('{{ t | indent(2, True) }}').render(t=jinja2.Markup(text))
print env.from_string('{{ t | indent(2, True, True) }}').render(t=jinja2.Markup(text))
```

## Your Environment
* Python version: 2.7.13
* Jinja version: 2.10

terminalmage:
This ~~unescapes after concatenating~~ ensures that we only concatenate `Markup` types to other `Markup` types, to fix a bug introduced in Jinja 2.10 where concatenated lines contain escaped versions of quotes and other characters.

Resolves #823.

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -567,6 +567,8 @@ def do_indent(
             rv += u'\n' + u'\n'.join(
                 indention + line if line else line for line in lines
             )
+            # Unescape any strings that were escaped due to concatenation
+            rv = Markup(rv.unescape())
 
     if first:
         rv = indention + rv
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -557,8 +557,19 @@ def do_indent(
     s += u'\n'  # this quirk is necessary for splitlines method
     indention = u' ' * width
 
+    def _unescape(val):
+        '''
+        Unescape any strings that were escaped due to concatenation. If the
+        value passed is not a Markup instance, return the original value.
+        '''
+        try:
+            return Markup(val.unescape())
+        except AttributeError:
+            # s was not a Markup instance
+            return val
+
     if blank:
-        rv = (u'\n' + indention).join(s.splitlines())
+        rv = _unescape((u'\n' + indention).join(s.splitlines()))
     else:
         lines = s.splitlines()
         rv = lines.pop(0)
@@ -567,12 +578,7 @@ def do_indent(
             rv += u'\n' + u'\n'.join(
                 indention + line if line else line for line in lines
             )
-            try:
-                # Unescape any strings that were escaped due to concatenation
-                rv = Markup(rv.unescape())
-            except AttributeError:
-                # s was not a Markup instance
-                pass
+            rv = _unescape(rv)
 
     if first:
         rv = indention + rv
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -554,31 +554,25 @@ def do_indent(
         ), stacklevel=2)
         first = indentfirst
 
-    s += u'\n'  # this quirk is necessary for splitlines method
     indention = u' ' * width
+    newline = u'\n'
 
-    def _unescape(val):
-        '''
-        Unescape any strings that were escaped due to concatenation. If the
-        value passed is not a Markup instance, return the original value.
-        '''
-        try:
-            return Markup(val.unescape())
-        except AttributeError:
-            # s was not a Markup instance
-            return val
+    if isinstance(s, Markup):
+        indention = Markup(indention)
+        newline = Markup(newline)
+
+    s += newline  # this quirk is necessary for splitlines method
 
     if blank:
-        rv = _unescape((u'\n' + indention).join(s.splitlines()))
+        rv = (newline + indention).join(s.splitlines())
     else:
         lines = s.splitlines()
         rv = lines.pop(0)
 
         if lines:
-            rv += u'\n' + u'\n'.join(
+            rv += newline + newline.join(
                 indention + line if line else line for line in lines
             )
-            rv = _unescape(rv)
 
     if first:
         rv = indention + rv
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -567,8 +567,12 @@ def do_indent(
             rv += u'\n' + u'\n'.join(
                 indention + line if line else line for line in lines
             )
-            # Unescape any strings that were escaped due to concatenation
-            rv = Markup(rv.unescape())
+            try:
+                # Unescape any strings that were escaped due to concatenation
+                rv = Markup(rv.unescape())
+            except AttributeError:
+                # s was not a Markup instance
+                pass
 
     if first:
         rv = indention + rv
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ venv-*/
 .coverage
 .coverage.*
 htmlcov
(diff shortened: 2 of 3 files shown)


terminalmage:
I'm trying to add a unit test for this case and found a failure, I'm working on it now.

davidism:
Thanks for looking into this!

jinja2/filters.py
```diff
@@ -567,6 +567,8 @@ def do_indent(
             rv += u'\n' + u'\n'.join(
                 indention + line if line else line for line in lines
             )
+            # Unescape any strings that were escaped due to concatenation
+            rv = Markup(rv.unescape())
```
davidism: Is this the correct place to unescape? Also, calling `Markup` right after `unescape` seems like a no-op, and might fail if `rv` isn't already markup. `s` should probably be unescaped as soon as possible and only if it's `Markup`, then escaped again at the end if it was previously escaped or if autoescape is enabled in the current context.

terminalmage: @davidism Yeah, I'm fixing now. I wasn't aware that there were cases where a non-Markup type would be passed to the filter, as this is my first time diving into jinja2 code.

davidism: I'm not certain on this, I'd have to play around with it, but it needs to be checked.

ThiefMaster:
Sounds like something that should be covered with unit tests that run in different autoescape modes, and with both normal and Markup strings.

jinja2/filters.py
terminalmage: The idea to call `Markup()` again after unescaping is because it will give us a `Markup` instance with the escaped chars unescaped. My fix will contain this solution instead:

```python
if lines:
    rv += u'\n' + u'\n'.join(
        indention + line if line else line for line in lines
    )
    try:
        # Unescape any strings that were escaped due to concatenation
        rv = Markup(rv.unescape())
    except AttributeError:
        # s was not a Markup instance
        pass
```

This way, if `s` is not a `Markup` instance, then the `unescape()` triggers an `AttributeError`, and `rv` will be left unmodified.

terminalmage: Here's another example of what I mean by calling `Markup()`. This is because `.unescape()` returns a unicode type.

```python
>>> from markupsafe import Markup
>>> x = Markup('foo ') + "'bar baz'"
>>> x
Markup(u'foo &#39;bar baz&#39;')
>>> Markup(x.unescape())
Markup(u"foo 'bar baz'")
>>>
```

terminalmage:
@ThiefMaster I'm adding additional test cases for input that is a `Markup` instance.

terminalmage:
PR updated with a fix and additional test coverage. Travis failure is in the docs build and looks to be unrelated.

terminalmage:
@davidism would you mind taking another look?

jinja2/filters.py
```diff
@@ -567,8 +567,12 @@ def do_indent(
             rv += u'\n' + u'\n'.join(
                 indention + line if line else line for line in lines
             )
-            # Unescape any strings that were escaped due to concatenation
-            rv = Markup(rv.unescape())
+            try:
+                # Unescape any strings that were escaped due to concatenation
+                rv = Markup(rv.unescape())
```
davidism: Why does this need to be done here but not in the `if blank:` case above?

davidism: Why doesn't this need to be done to the `if blank:` case above?

davidism: This is incorrect if autoescape is currently disabled.

jinja2/filters.py
```diff
@@ -567,6 +567,12 @@ def do_indent(
             rv += u'\n' + u'\n'.join(
                 indention + line if line else line for line in lines
             )
+            try:
+                # Unescape any strings that were escaped due to concatenation
+                rv = Markup(rv.unescape())
```
davidism: Why doesn't this need to be done to the `if blank:` case above?

davidism: This is incorrect if autoescape is currently disabled.

jinja2/filters.py
terminalmage: You're right. This is probably just a case of me coding my fix to the use case I saw.

davidism:
Wait, what comment did I delete? GitHub's being weird, my review was at the same time as you pushed another commit.

terminalmage:
I pushed a commit to fix a spelling error.

jinja2/filters.py
terminalmage: Can you explain? I don't see autoescape factoring into `do_indent()` anywhere, and like I said, this is my first crack at working with the jinja2 source code. The idea of this try/escape is that we attempt to unescape, and if it's a string type instead of a `Markup` then `rv` remains unmodified. EAFP.

terminalmage: You're right, this was just a case of me coding my fix to the use case with which I was familiar.

davidism: OK, I think my original statement may be somewhat inaccurate, but there's an even bigger issue here. Now you're unescaping whatever came in and wrapping in in `Markup`, indicating to Jinja's autoescape mechanism that the contents are safe. So for example if I did `{{ user_input|indent }}`, and the user passed in `<big xss injection>`, your code undoes any escaping that was already done to that and tells Jinja not to escape it, so the injection gets through.

terminalmage:
@davidism I've updated the PR to apply the unescape fix in the top half of the if/else block as well. I had one question about your comment [here](https://github.com/pallets/jinja/pull/826#discussion_r175580948) though.

davidism:
OK, for some reason GitHub keeps hiding my reviews, so I'm putting this in a comment instead.

You've introduced a security issue. Your code unescapes the input then marks it safe, so Jinja will not escape it later. So if I do `{{ user_input|indent }}` and the user passes `<script injection>`, now `indent` undoes any existing escaping and prevents further escaping, thus passing through the injection.

terminalmage:
When the code that you've commented on changes, GitHub hides it, you can unhide it by clicking "Show outdated".

So, how would you suggest fixing this?

terminalmage:
I think I have an idea. If we just make sure the values being concatenated are `Markup` instances (when `s` is one), then we avoid having the string literal converted to a `Markup()` type, and we don't need the hacky "unescape" solution.

terminalmage:
Yep, that works, and tests pass with the unescape hack removed.

terminalmage:
@davidism Any other concerns?

unix196:
in which version Jinja2 it was fixed? Now in pip last version 2.10 and bug is still exist.

terminalmage:
This bug was introduced in 2.10, and there has been no release since then. You may want to try version 2.9.

PR #822, opened 2018-03-14
ignore errno.ENOTDIR in open_if_exists

praiskup:
Fixes #821

PR #839, opened 2018-04-10
Continue on search path if ENOTDIR returned when looking for templates

ishakir:
This became a problem in my build environment where we're running from python packaged into `.jar` files. The code looks for a path of the form `run.jar/templates/my_template.html` which returns `errno.ENOTDIR` breaking out of the whole search loop. This fixes that.

### DIFF
--- a/jinja2/utils.py
+++ b/jinja2/utils.py
@@ -153,7 +153,7 @@ def open_if_exists(filename, mode='rb'):
     try:
         return open(filename, mode)
     except IOError as e:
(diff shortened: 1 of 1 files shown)


ishakir:
I believe the build failure to be related to existing issues

ishakir:
This is a duplicate of the following:

https://github.com/pallets/jinja/pull/822

I have, however, gone through to check that there are no tests for `open_if_exists`

PR #846, opened 2018-04-25
Introduced the AbsolutePathFileSystemLoader

mpreisler:
It's a simplified FileSystemLoader that insists on absolute paths,
doesn't provide iteration over available templates and has no search
path functionality.


davidism:
Thanks for suggesting this! You're welcome to use this in your own project, or provide it as a separate package, but I don't see a strong use for this directly in Jinja.

Issue #584, opened 2016-05-09
Documentation: Please add an index to list of built-in filters and tests


aptituz:
It would be really helpful, if there were an index for the built-in filters and tests, so that one does not have to work through the whole list in order to find a certain (not yet known) filter or test.

jeffwidman:
PR welcome...

PR #853, opened 2018-05-14
Quick lookup table for builtin filters and tests-- references #584

doobeh:
Might need a little padding tweak to the CSS for spacing under the table.  References issue #584 

image

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -1319,6 +1319,23 @@ evaluates into an undefined object::
 List of Builtin Filters
 -----------------------
 
+===================  ===================  ==================  ======================  ====================
+Built-in Filters Index
+==========================================================================================================
+:func:`abs`          :func:`attr`         :func:`batch`       :func:`capitalize`      :func:`center`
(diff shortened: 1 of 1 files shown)


ThiefMaster:
While it looks pretty, it's going to be an absolute pain in the ass to add a new test/filter to that table.

(also, from a semantic point of view, this isn't data that belongs in a table since the rows/cols are meaningless)

doobeh:
I'll look into constructing something more automated-- similar to how the table directive for `jinja_filters` is built.  Can you think of a more semantic layout (py3 docs use the table layout for similar overviews).  Columns of lists I guess would be best and degrade well.

davidism:
If the new pallets-sphinx-themes doesn't handle table spacing correctly, it should be easy to fix there. I do like how the layout is used in the Python docs, I agree that it should be automated though.

davidism:
I'm going to merge this, but it's likely to get removed eventually. I'm working on improving the directives in [pallets-sphinx-themes](https://github.com/pallets/pallets-sphinx-themes).

Issue #748, opened 2017-07-15
Cannot use the plus sign(+) for whitespace control: tag name expected


nalzok:
## Expected Behavior

Assuming `trim_blocks` is always `False`.

When `lstrip_blocks` is set to `True`, jinja should output
```html
<!-- 1 -->
    <!-- 2 -->
foo<!-- 3 -->
<!-- 4 -->
<!-- 5 -->
```

And when it's `False`, jinja should output
```html
    <!-- 1 -->
    <!-- 2 -->
    foo<!-- 3 -->
    <!-- 4 -->
    <!-- 5 -->
```

## Actual Behavior

In either case,

> jinja2.exceptions.TemplateSyntaxError: tag name expected

## Template Code
```jinja
    <!-- 1 -->
    {%+ if True %}<!-- 2 -->
    foo<!-- 3 -->
    {% endif %}<!-- 4 -->
    <!-- 5 -->
```

The following template also has a similar issue:

```jinja
    <!-- 1 -->
    {% if True %}<!-- 2 -->
    foo<!-- 3 -->
    {%+ endif %}<!-- 4 -->
    <!-- 5 -->
```

## Full Traceback
```pytb
Traceback (most recent call last):
  File "/home/sunqingyao/Envs/flask-mega/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/sunqingyao/Envs/flask-mega/lib/python3.6/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/home/sunqingyao/Envs/flask-mega/lib/python3.6/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sunqingyao/Envs/flask-mega/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/sunqingyao/Envs/flask-mega/lib/python3.6/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sunqingyao/Envs/flask-mega/lib/python3.6/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sunqingyao/Envs/flask-mega/lib/python3.6/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
[...]

PR #858, opened 2018-05-27
Fix lexer being extremely slow on large amounts of whitespace. Fixes #857

### LINKED ISSUE #857: Template parsing (lexing) can have quadratic time complexity
## Expected Behavior
Parsing the following template should have linear complexity and be done in reasonable time.
```
Environment().from_string(' ' * 20000)
```

## Actual Behavior
It has quadratic complexity and takes a few seconds at 20000, a minute at 100k.

**In general, it appears any large segments of whitespace (" \t\n") without jinja tags that consume whitespace will degrade template parsing performance significantly.**

```
# This parses quickly due to "-"
Environment().from_string('{{ 0 -}}' + ' ' * 20000)
Environment().from_string(' ' * 20000 + '{{- 0 }}')
# This is very slow
Environment().from_string('{{ 0 }}' + ' ' * 20000)
Environment().from_string(' ' * 20000 + '{{ 0 }}')
```

## Cause
I investigated the cause of the slow rendering and it lies in two pathological regular expressions used in Jinja lexer, "root" and TOKEN_RAW_BEGIN (not illustrated in examples but suffers from the same issue). In the **extremely simplified** version of the root lexer regex shown below, notice whitespace may be consumed either by `(.*?)` or by `\s*`, with the latter having priority due to being greedy, but only in case the block of whitespace is followed by variable (or block/comment) opening tag with the "-" modifier (that forces left-stripping whitespace). This is a pathological case for the built-in Python `re` module and will have quadratic complexity due to backtracking done after consuming every single whitespace character (don't quote me on that though).
```
r'(.*?)(?:(?P<variable_begin>\s*{{-|{{))'
```
Jinja lexer does that to implement the whitespace stripping feature on the left side of tags, but this approach is extremely inefficient with built in Python `re` module (note the `regex` library doesn't suffer from this, but it is slower in general). 

## Proposed fix
I created a fix for this issue by re-implementing the way striping whitespace on the left side of tags is done (including the lstrip_blocks environment flag) by doing this outside of regular expressions, which in my opinion simply are not the right tool for this task. I didn't change the implementation of striping whitespace to the right of tags as the "-" character is matched **before** any whitespace, changing the state of the regex automaton before any backtracking needs to be done.

https://github.com/pallets/jinja/pull/858

## Implications
The most significant implication is Jinja being extremely slow when parsing templates with segments of whitespace in tens of thousands. Such templates should be rare, but may occur. The application where I discovered this issue has almost 100k Jinja templates created by semi-technical users and a few of them contained such large segments of whitespace, most likely added by accident.

The more frequent implication is Jinja template parsing being sub optimal, as most Jinja templates will contains blocks of whitespace in tens or hundreds of characters. The performance impact of quadratic complexity on such small scale won't be that dramatic, but I measured over 2x average improvement in template parsing speeds after applying the fix on the above mentioned almost 100k different templates. I will post a histogram in the PR description.

## Environment
* Python version: tested with 2.7, 3.6, apparently any Python version
* Jinja version: 2.8.1, current master, apparently any Jinja version in last few years

petee-d:
Fix for #857, where parsing whitespace in templates usually has quadratic time complexity.

It addresses two pathological regular expressions used in Jinja lexer, "root" and TOKEN_RAW_BEGIN (not illustrated in examples but suffers from the same issue). In the **extremely simplified** version of the root lexer regex shown below, notice whitespace may be consumed either by `(.*?)` or by `\s*`, with the latter having priority due to being greedy, but only in case the block of whitespace is followed by variable (or block/comment) opening tag with the "-" modifier (that forces left-stripping whitespace). This is a pathological case for the built-in Python `re` module and will have quadratic complexity due to backtracking done after consuming every single whitespace character (don't quote me on that though).
```
r'(.*?)(?:(?P<variable_begin>\s*{{-|{{))'
```
Jinja lexer does that to implement the whitespace stripping feature on the left side of tags, but this approach is extremely inefficient with built in Python `re` module (note the `regex` library doesn't suffer from this, but it is slower in general).

My approach uses `.rstrip` or `.rfind` + another regular expression to remove the whitespace inside the tokenizer.

## Potential improvements
* I found the code to be quite hacky, especially the inspection of type of `tokens` in method `tokeniter` (shouldn't it be `tokenizer` btw?), when I needed to add a flag for a rule to attempt l-stripping, I saw no better way to continue in that pattern if I wanted to avoid larger refactoring. Any ideas?
* Is this project still sticking to 80 character maximum per line, as lexer.py apparently does? It makes the code very ugly and I see some modules no longer adhere to that limit. Is 120 character line maximum acceptable?
* It feels like the code somewhat avoids private helper methods. I especially feel like my new code block for l-stripping should be in such a separate method and I will be happy to move it to one unless you are actually trying to avoid the overhead of calling methods (which in my opinion should be negligible here).

## Testing and performance measurement
I tested the fix on almost 100k of different Jinja2 templates, created by users with different technical skills and used for various purposes (from templates without any jinja tags to huge personalized email templates). Parsing all templates produced the exact same parsing tree both before and after my fix. The Jinja2 test suite of course passes as well.

The table below shows the parsing times of the templates I tested the fix with, each parsing attempt before or after was added to a bucket based on rounded log2 of the parsing time. You can see the 4 extremely slow cases being fixed and a general improvement of parsing time.
```
Seconds | Before | After
0.00006 |    873 |  1775
0.00012 |  19690 | 21348
0.00024 |   8805 |  9151
0.00049 |   5294 |  5331
0.00098 |   3603 |  2146
0.00195 |   2002 |  4824
0.00391 |   3109 |  8404
0.00781 |   7778 |  7759
0.01563 |   8455 |  3419
0.03125 |   4889 |  2234
0.06250 |   1711 |    63
0.12500 |    177 |     3
0.25000 |     67 |     0
0.50000 |      0 |     0
1.00000 |      0 |     0
2.00000 |      0 |     0
4.00000 |      2 |     0
8.00000 |      2 |     0
```

I didn't add any test case. I could add a test that parses something like (`' ' * 40000`) and asserts the time it took didn't cross a reasonably large boundary, but I generally dislike such tests as they may be fragile to CI environment issues. Let me know what you think.


petee-d:
Note the build failed on something unrelated to my changes (Python 3.3.6 environment creation and it passes locally for me), could someone with repo access try to restart it or fix the underlying issue?

davidism:
Sorry for letting this sit so long! I've marked this for 2.11. Unfortunately, there have been some other changes to the lexer and `lstrip_blocks`, and this doesn't merge cleanly. If you see this and can address this, let me know, as you understand the changes better than me.

davidism:
I rebased, and it seems like your fix for this also addressed #748, or at least it passes the tests added by #1014. I'll check a little more.

petee-d:
Oh, hey there. I'm afraid I only barely remember the changes I have done almost year and a half ago, but can help if needed. I understand you already rebased this (without pushing I guess), so I won't dive into it unless you say you'd like me to.

BTW, I would be quite happy if this got merged (or otherwise fixed), our project was using a fork with my fix ever since I made it and being able to update Jinja would be nice.

davidism:
No problem, I think I have a handle on it, I'll try to push and merge later today.

davidism:
For some reason I can't force push to your branch, even though Kevin was able to earlier this year. Would you check that the option to allow maintainer pushes to this PR is enabled?

davidism:
For some reason I couldn't push my rebase to the source repo for this PR, but GitHub happily let me merge and commit to it using the online editor. :man_shrugging:

jinja2/lexer.py
```diff
@@ -423,6 +421,17 @@ def get_lexer(environment):
     return lexer
 
 
+class OptionalLStrip(tuple):
+    """A special tuple for marking a point in the state that can have
+    lstrip applied.
+    """
+
```
jab: Worth adding `__slots__ = ()` so we don’t give up slots optimization from tuple?

davidism: Yeah, I missed that

davidism: 6b9eb6df5a7804ec4210bf449296aae71eb5cd3e

Issue #444, opened 2015-04-28
urlencode filter does not escape slashes


catharsis:
The driving function behind the urlencode filter (from jinja2/_compat.py)
````python
from urllib import quote as url_quote
````
doesn't escape slashes by default. This isn't documented (as far as I've been able to find) for the urlencode filter.

Is this intentional? If it's not, how about supplying a way to overriding the default behaviour or even make slash-escaping the default behaviour (although I expect that might be trickier, since it might break running code)? If it is intentional, documenting that fact would be helpful.

catharsis:
Added a PR with suggestion on doc-change, if that's the way to go.

untitaker:
The PR was closed through https://github.com/mitsuhiko/jinja2/commit/46acbf02ed9ab58c7a92553c95790e75bdc3c930, closing this too.

PR #864, opened 2018-06-06
Implement do_urlencodecomponent

agowa338:
To have a filter that acts like do_urlencode, but also encodes slashes.
Some api's require passing slashes in there encoded format as get parameters. For example gitlab does so for querying the projects api.

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -112,6 +112,26 @@ def do_urlencode(value):
                      for k, v in itemiter)
 
 
+def do_urlencodecomponent(value):
+    """Escape strings for use in URLs (uses UTF-8 encoding).
+    It accepts regular strings, other types are handled by urlencode().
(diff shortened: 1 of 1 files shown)


agowa338:
This fixes
#515 
#444

agowa338:
@davidism Sorry for mentioning, but may I ask if something is missing from this pr to get it merged? This pr is now open for over a year and that makes me think that I missed something.

davidism:
Thanks for working on this. However, based on the existing feedback we've decided not to include this in Jinja core. See @mitsuhiko's comment here for the latest statment: https://github.com/pallets/jinja/issues/515#issuecomment-548960613.

When building URLs, you should be working with more precise functions than manually building them with partial filters like this one. Write a filter appropriate for users of your project, such as `url_unsplit`, or Flask's `url_for`.

PR #867, opened 2018-06-27
Import abstract base classes from collections.abc

The-Compiler:
In Python 3.7, importing ABCs directly from the `collections` module shows a
warning (and in Python 3.8 it will stop working) - see
https://github.com/python/cpython/commit/c66f9f8d3909f588c251957d499599a1680e2320

This fixes various DeprecationWarnings such as those:

```
.../jinja2/utils.py:485: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import MutableMapping

.../jinja2/runtime.py:318: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Mapping
```

### DIFF
--- a/jinja2/runtime.py
+++ b/jinja2/runtime.py
@@ -20,7 +20,7 @@ from jinja2.exceptions import UndefinedError, TemplateRuntimeError, \
      TemplateNotFound
 from jinja2._compat import imap, text_type, iteritems, \
      implements_iterator, implements_to_string, string_types, PY2, \
-     with_metaclass
+     with_metaclass, abc
 
 
 # these variables are exported to the template runtime
@@ -313,12 +313,7 @@ class Context(with_metaclass(ContextMeta)):
         )
 
 
-# register the context as mapping if possible
-try:
-    from collections import Mapping
-    Mapping.register(Context)
-except ImportError:
-    pass
+abc.Mapping.register(Context)
 
 
 class BlockReference(object):
--- a/jinja2/utils.py
+++ b/jinja2/utils.py
@@ -14,7 +14,7 @@ import errno
 from collections import deque
 from threading import Lock
 from jinja2._compat import text_type, string_types, implements_iterator, \
-     url_quote
+     url_quote, abc
 
 
 _word_split_re = re.compile(r'(\s+)')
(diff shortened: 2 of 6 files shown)


jinja2/runtime.py
```diff
@@ -313,12 +313,7 @@ def __repr__(self):
         )
 
 
-# register the context as mapping if possible
-try:
-    from collections import Mapping
-    Mapping.register(Context)
-except ImportError:
-    pass
```
The-Compiler: I'm not sure why this `ImportError` handling was here, seeing that Python 2.6 always has `collections.Mapping` available. It was added a long time ago (>10 years) in 9bb7e4779182490abc6e1784b0ee63d22b91b11e, so I suppose that was needed for an older Python version but isn't anymore.

jinja2/utils.py
```diff
@@ -480,12 +480,7 @@ def __reversed__(self):
     __copy__ = copy
 
 
-# register the LRU cache as mutable mapping if possible
-try:
-    from collections import MutableMapping
-    MutableMapping.register(LRUCache)
-except ImportError:
-    pass
+abc.MutableMapping.register(LRUCache)
```
The-Compiler: I'm not sure why this `ImportError` handling was here, seeing that Python 2.6 always has `collections.MutableMapping` available. It was added a long time ago (>10 years) in 9bb7e4779182490abc6e1784b0ee63d22b91b11e, so I suppose that was needed for an older Python version but isn't anymore.

The-Compiler:
The Travis failure is unrelated, it happens because Python 3.3 got dropped in `wheel` (and probably elsewhere too).

reinout:
Looking at the changelog: this fix doesn't seem to be released yet, right?
If so, will there be a release fairly soon? My project's pytest output is showing quite a number of these deprecation warnings :-)

devdave:
Just checking in on the release schedule for when this is going out to pypi.

cepbrian:
This fix has STILL not been released as of 2.10. 

I'm using pip, python3.7, and jinja2 version 2.10. I still get this warning.

jbarreneche:
Can this be backported to the 2.10.x branch? There was a release 2 month ago but it doesn't include this fix

rodneywells01:
Would love for this to be merged in. Can confirm it does not appear fixed.

tobiasfeil:
Still getting the warning with pytest 5.0.0, Jinja2 2.10.1 and pyparsing 2.4.0 (everything most current). Am I importing something wrongly? Or is this still an issue?

PR #868, opened 2018-06-27
Drop Python 2.6 and 3.3 support

The-Compiler:
It's currently broken on Travis (as it was dropped in 'wheel'), so continuing to
support it is probably not worth the pain.


davidism:
Drop 2.6 as well.

The-Compiler:
Done!

PR #872, opened 2018-06-30
Cast searchpath to string in case it is a pathutils Path object

breisfeld:


jinja2/loaders.py
```diff
@@ -160,7 +160,7 @@ class FileSystemLoader(BaseLoader):
     def __init__(self, searchpath, encoding='utf-8', followlinks=False):
         if isinstance(searchpath, string_types):
             searchpath = [searchpath]
-        self.searchpath = list(searchpath)
+        self.searchpath = [str(_) for _ in list(searchpath)]
```
davidism: Don't use `_` as a variable name if you're not ignoring the variable.

breisfeld: Thank you for the tip.

davidism:
Moved to #873

PR #873, opened 2018-06-30
Cast searchpaths to strings in case they are pathutils Path objects

breisfeld:


jinja2/loaders.py
```diff
@@ -160,7 +160,7 @@ class FileSystemLoader(BaseLoader):
     def __init__(self, searchpath, encoding='utf-8', followlinks=False):
         if isinstance(searchpath, string_types):
             searchpath = [searchpath]
-        self.searchpath = list(searchpath)
+        self.searchpath = [str(x) for x in list(searchpath)]
```
davidism: Seems like this will still fail if searchpath is a single value. It's not `string_types`, so instead you'll try to iterate over a path.

jinja2/loaders.py
```diff
@@ -160,7 +160,7 @@ class FileSystemLoader(BaseLoader):
     def __init__(self, searchpath, encoding='utf-8', followlinks=False):
         if isinstance(searchpath, string_types):
             searchpath = [searchpath]
-        self.searchpath = list(searchpath)
```
davidism: Don't need the inner `list` call.

breisfeld:
Will submit a new pull request after testing

ThiefMaster:
Just reopen this one.

breisfeld:
I am really trying to make a positive contribution to jinja. 
I am a competent programmer in python, but the various git/github peculiarities have me stymied. 
My latest pull request is here: https://github.com/pallets/jinja/pull/876

PR #874, opened 2018-06-30
Cast searchpath to string in case it is a pathlib Path object

breisfeld:


tests/test_loader.py
```diff
@@ -42,6 +48,11 @@ def test_filesystem_loader(self, filesystem_loader):
         assert tmpl.render().strip() == 'BAR'
         tmpl = env.get_template('foo/test.html')
         assert tmpl.render().strip() == 'FOO'
+        if has_pathlib:
```
davidism: Write a separate test rather than adding to an existing test. Use `pytest.mark.skipif` to skip the test if pathlib isn't available.

tests/test_loader.py
```diff
@@ -15,6 +15,12 @@
 import pytest
 import weakref
 
+try:
+    import pathlib
+    has_pathlib = True
+except ImportError:
+    has_pathlib = False
```
davidism: `pathlib = None` rather than a separate flag.

breisfeld:
Revised logic and tests per review comments

PR #876, opened 2018-07-04
Cast searchpath to string in case it is a pathlib Path object

breisfeld:


ThiefMaster:
Just thinking.. would it make more sense to check for:

- is a string type (like before)
- has a [PEP519-style `__fspath__` method](https://www.python.org/dev/peps/pep-0519/)
- otherwise assume iterable, and use fspath if available or the plain value

Like this it would not only support pathlib paths / objects whose `__str__` is a path, but any object that represents a filesystem path.

Since pathlib support is a new feature in general I don't think the fact that `__fspath__` is Python3.6-only (probably one could use a pathlib backport for any older Python version) is a problem.

breisfeld:
I was not aware of the __fspath__ method. Because PEP 519 was accepted, your idea makes a lot of sense and would be more forward-thinking than my implementation.

PR #878, opened 2018-07-16
enhancement: add support of glob include to jinja2.FileSystemLoader

ssato:
Extend FileSystemLoader to allow include multiple template files at once
with 'include' directive contains shell glob patterns, such like
'{% include "/path/to/templates/*.j2" %}' for example.

Also, add a simple test cases and template files and make template files
ignored by coverage in .coveragerc.


ssato:
why?

PR #892, opened 2018-08-30
Add trove classifier for PyPy support

jdufresne:
Tested during CI since 761da6ff612966214fe9e34ac6bc86499e34da26.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,8 @@ setup(
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: Implementation :: CPython',
+        'Programming Language :: Python :: Implementation :: PyPy',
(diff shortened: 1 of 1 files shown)


davidism:
Is there a reason beyone "users could technically filter by them in PyPI" to use these? I don't use them in any other projects, in Pallets or otherwise.

jdufresne:
I find them useful because they allow potential users to know -- at a quick glance -- what Pythons are supported by a library. Sometimes when integrating a library with my own projects, I'll use this information to help decide if it is good for integration.

As the trove classifiers have a consistent form, they are also used by sites and offline scripts to analyze the changes in the Python community. Some examples:

- https://python3wos.appspot.com/
- https://hugovk.github.io/drop-python/

If you don't support PyPy as a platform, then I don't think you should add it. But PyPy is included in CI, so it appears that compatibility is desired. I think adding CPython is only useful in contrast to PyPy so as not to give the impression that the library is PyPy only.

Up to you.

PR #894, opened 2018-09-06
This fixes #727

### LINKED ISSUE #727: Jinja2 get_source() function crashes when template parameter arrives with full path
Let me try to explain the problem because is very specific.

I have a Pyramid+Jinaj2 applications that is extendable using plugins. The main applications loads external python programs from eggs that can extend the templates of the parent. See [here](https://github.com/qlands/pyramid-pca-jinja2)

The main application creates two jinja2 extensions "extend_this" and "snippet" :
```python
def render_snippet(template_name, **kw):
    ''' This function will render the snippet.

    This code is based on CKAN 
    :Copyright (C) 2007 Open Knowledge Foundation
    :license: AGPL V3, see LICENSE for more details.

     '''
    request = get_current_request()

    template = jinjaEnv.get_template(template_name)
    output = template.render(kw, renderer='snippet',_=request.translate)
    output = '\n<!-- Snippet %s start -->\n%s\n<!-- Snippet %s end -->\n' % (template_name, output, template_name)
    return literal(output)


class extendThis(ext.Extension):
    ''' Custom {% extend_this %} tag that allows templates
    to inherit from the host template futher down the template search path.

    This code is based on CKAN 
    :Copyright (C) 2007 Open Knowledge Foundation
    :license: AGPL V3, see LICENSE for more details.

    '''

    tags = ['extend_this']

    def __init__(self, environment):
        ext.Extension.__init__(self, environment)
        try:
            self.searchpath = jinjaEnv.loader.searchpath[:]
        except AttributeError:
            # this isn't available on message extraction
            pass

    def parse(self, parser):
        lineno = next(parser.stream).lineno
        node = nodes.Extends(lineno)
        template_file = parser.filename
        template_path = parser.filename

        # We need to have a list of template paths to look for
        if not hasattr(self, 'searchpath'):
            return node

        # First we remove the templates path from the file
        # so to have the just the template file or a template file in a subdirectory of templates
        for searchpath in self.searchpath:
            template_file = template_file.replace(searchpath, '')
        # Here we get the template path of the file
        template_path = template_path.replace(template_file, '')

        # Find the position of the template's path in the list of paths
        index = -1
        try:
            index = self.searchpath.index(template_path)
        except:
            pass
        if index == -1:
            return node

        # index is the position of the this template's path
        # so we search down stream for the template in other paths
        file_to_extend = ''
        for pos in range(index + 1, len(self.searchpath)):
            if os.path.exists(self.searchpath[pos] + template_file):
                file_to_extend = self.searchpath[pos] + template_file
                break

        # If the file to extend from exits then set it as a template
        if file_to_extend == '':
            return node
        else:
            node.template = nodes.Const(file_to_extend)

        return node


class BaseExtension(ext.Extension):
    ''' Base class for creating custom jinja2 tags.
    parse expects a tag of the format
    {% tag_name args, kw %}
    after parsing it will call _call(args, kw) which must be defined.

    This code is based on CKAN 
    :Copyright (C) 2007 Open Knowledge Foundation
    :license: AGPL V3, see LICENSE for more details.

    '''

    def parse(self, parser):
        stream = parser.stream
        tag = stream.next()
        # get arguments
        args = []
        kwargs = []
        while not stream.current.test_any('block_end'):
            if args or kwargs:
                stream.expect('comma')
            if stream.current.test('name') and stream.look().test('assign'):
                key = nodes.Const(stream.next().value)
                stream.skip()
                value = parser.parse_expression()
                kwargs.append(nodes.Pair(key, value, lineno=key.lineno))
            else:
                args.append(parser.parse_expression())

        def make_call_node(*kw):
            return self.call_method('_call', args=[
                nodes.List(args),
                nodes.Dict(kwargs),
            ], kwargs=kw)

        return nodes.Output([make_call_node()]).set_lineno(tag.lineno)

class SnippetExtension(BaseExtension):
    ''' 

    This tags inject small portions of reusable code i.e. snippets
    into a jinja2 template

    {% snippet <template_name> [, <keyword>=<value>].. %}

    This code is based on CKAN 
    :Copyright (C) 2007 Open Knowledge Foundation
    :license: AGPL V3, see LICENSE for more details.


    '''

    tags = ['snippet']

    @classmethod
    def _call(cls, args, kwargs):
        assert len(args) == 1
        return render_snippet(args[0], **kwargs)

```
The "extend_this" basically look for the template file name down the jinja2 search paths and then sets the node.template to a parent template. So a plugin can extend a template of the main application. Is like a version of {% extend 'template' %} but allowing a path relative to the main application and the plugin.

The "snippet" just injects code in a template by calling get_template and render.

The problem happens ONLY when the snippet renders a template that has {% extend_this %}. Meaning {% snippet %} calls render then render finds {% extend_this %}

For some reason ONLY in this particular case the function get_source(self, environment, template) in line 167 in loaders.py of Jinja2 source arrives with a full path (not relative like in the rest of the cases). This function assumes relative path thus when join happens in line 170 no template file is found.

I patched the Jinja2 code to overcome the problem:
```python
def get_source(self, environment, template):
        joinPath = True
        for searchpath in self.searchpath:
          # If the template file contains any of the search paths
	  if template.find(searchpath) >= 0:
            joinPath = False

        pieces = split_template_path(template)

        for searchpath in self.searchpath:
	    if joinPath:
	      filename = path.join(searchpath, *pieces)
            else:
	      filename = template

            f = open_if_exists(filename)
            if f is None:
                continue
            try:
                contents = f.read().decode(self.encoding)
            finally:
                f.close()

            mtime = path.getmtime(filename)

            def uptodate():
                try:
                    return path.getmtime(filename) == mtime
                except OSError:
                    return False
            return contents, filename, uptodate

```

---

## Expected Behavior
The template parameter should arrive with relative path as other cases

## Actual Behavior
The template parameter arrives with full path

## Template Code calling snippet
```jinja
{% snippet 'snippets/question/question-item.jinja2', request=request, UserQuestion =UserQuestion, QuestionsOptions =QuestionsOptions, User = "", showtools = False %}
```

## The template question-item.jinja2 
```jinja
{% extend_this %}

{% block qstitemtoolextra %}
    <input name="CHK{{ UserQuestion.question_id }}" type="checkbox">
{% endblock qstitemtoolextra %}
```

## Your Environment
* Python version: 2.7.13
* Jinja version: 2.9.6

qlands:
In some cases when the template inheritance gets too complex the template files arrive with absolute path. If the template file has an absolute path then just use the template file and not join the parts.


qlands:
This is not necessary anymore. I realized (late) that jinja2 has include thus my extension called snippet is not necessary. The include clause does not generate the said behavior.

PR #906, opened 2018-10-04
Add a flip filter

mattiasb:
The flip filter applies a filter with arguments flipped in reverse order. This
becomes useful with filters like map where you can't easily re-order the
arguments yourself.

See these Ansible issues and PR's for context:
[Ansible Issue #46215](https://github.com/ansible/ansible/issues/46215)
[Ansible PR #46255](https://github.com/ansible/ansible/pull/46255)
[Ansible PR #46340](https://github.com/ansible/ansible/pull/46340)

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -941,6 +941,29 @@ def do_attr(environment, obj, name):
 
 
 @contextfilter
+def do_flip(context, value, name, *args, **kwargs):
+    """Applies a filter with arguments flipped in reverse order.
+    This becomes useful with filters like map where you can't easily re-order
+    the arguments yourself.
+
+    Say you want to find all numbers in a list that divides 10 you'd do it like
+    this_
+
+    .. sourcecode:: jinja
+
+        {{ [1,2,3,4,5,6,7,8,9] | select('flip', 'divisibleby', 10) | list }}
+
+
+    This would give you back `[1, 2, 4, 5]`.
+
+    .. versionadded:: 2.11
+    """
+    args = args[::-1] + (value,)
+    return context.environment.call_filter(
+        name, args[0], args[1:], kwargs=kwargs, context=context)
+
+
+@contextfilter
 def do_map(*args, **kwargs):
     """Applies a filter on a sequence of objects or looks up an attribute.
     This is useful when dealing with lists of objects but you are really
@@ -1153,6 +1176,7 @@ FILTERS = {
     'escape':               escape,
     'filesizeformat':       do_filesizeformat,
     'first':                do_first,
+    'flip':                 do_flip,
     'float':                do_float,
     'forceescape':          do_forceescape,
     'format':               do_format,
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -946,17 +946,51 @@ def do_flip(context, value, name, *args, **kwargs):
     This becomes useful with filters like map where you can't easily re-order
     the arguments yourself.
 
-    Say you want to find all numbers in a list that divides 10 you'd do it like
-    this_
+    Example 1:
+
+    Say you have a list of names and you want to prefix them all with "Name: "
+    you could do it like this:
 
     .. sourcecode:: jinja
 
-        {{ [1,2,3,4,5,6,7,8,9] | select('flip', 'divisibleby', 10) | list }}
+        {{ ["Carl", "Jane"] | map("flip", "format", "Name: %s") | join(",") }}
 
+    This would give you back:
 
-    This would give you back `[1, 2, 4, 5]`.
+        Name: Carl, Name: Jane
+
+    Example 2:
+
+    Say you have a User object like below:
+
+    .. sourcecode:: python
+
+        class User(object):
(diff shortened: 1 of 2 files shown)


jinja2/filters.py
```diff
@@ -940,6 +940,29 @@ def do_attr(environment, obj, name):
...
+    .. sourcecode:: jinja
+
+        {{ [1,2,3,4,5,6,7,8,9] | select('flip', 'divisibleby', 10) | list }}
+
+
+    This would give you back `[1, 2, 4, 5]`.
+
+    .. versionadded:: 2.11
+    """
+    args = list(args)[::-1] + [value]
```
ThiefMaster: `args = args[::-1] + (value,)`?

mattiasb: I don't understand. :)

ThiefMaster: avoids converting to a list - you don't need the mutability so no reason to not stay with tuples

mattiasb: Ah! I didn't know `*args` was a tuple.

mattiasb: Fixed!

davidism:
The semantics of this are really hard for me to keep straight. What the filter actually produces depends on the filter it's flipping. Would you add a few more examples to demonstrate different uses of this?

mattiasb:
@davidism I think the best example is in the original [Ansible PR](https://github.com/ansible/ansible/pull/46340).

But in short, this is what mad *me* want this:
In Ansible we have a custom-filter [combine](https://docs.ansible.com/ansible/2.5/user_guide/playbooks_filters.html#combining-hashes-dictionaries) that you can use to merge two objects. It has override-semantics by default so for example:
```
{{ {'a':'upstream', 'b':'upstream'} | combine({'b':'overridden'}) == {'a':'user-set', 'b':'overridden'} }}
```
So if you want to apply default values from the `defaults` object and override from the `overrides` object on an unknown single object `upstream` you can do it like this:
```
{{ defaults | combine(upstream) | combine(overrides) }}
```
This all works good until you want to apply defaults and overrides to all elements in *a list* of objects with the `map` filter. You can apply overrides easily like this:
```
{{ upstream_objs | map('combine', overrides) }}
```
If you want to apply defaults you'll have to resort to the statement-language like so (in ansible):
```
objects_with_defaults_applied: >-
  {%- result=[] -%}
  {%- for obj in upstream_objs -%}
  {%-     do result.append(defaults | combine(obj)) -%}
  {%- endfor -%}
  {{ result | to_json | from_json }}
```
My immediate thought was to implement a version of combine that took arguments in reverse order. See [Ansible #46255](https://github.com/ansible/ansible/pull/46255), but that PR got rejected (and rightly so) since it was too specific. This issue apparently comes up in more places.

So that led me to think about the problem at hand. The Jinja expression language really is a small functional programming language. So my thought was "How does the FP community handle this?" and the answer usually is "with lambdas or the flip-function". See [here](https://github.com/ansible/ansible/pull/46255#issuecomment-425490148) for context.

So I implemented this for Ansible here: [Ansible #46340](https://github.com/ansible/ansible/pull/46340). But I thought that it might be better to get this upstream and @bcoca seemed to agree. So that's where I am now. :)

One issue I see with this approach is that it won't work with `select` since there's a distinction between filters and tests and `select` specifically takes `tests` as input. But other than that it seems to work fine for me and my little use case. I'm hoping that the idea has more general value though.

davidism:
Thanks, that's helpful, but I meant add some more short examples to the docs like the one that's there already. I should have been more specific.

mattiasb:
Ah of course! I need to remove the one that's there actually, since it won't work with `select`.

mattiasb:
Added a fixup commit (that I will squash later if it looks good). It's a bit hard to produce good examples with the few filters that take more than 1 parameter in just bare Jinja but I tried as good as I could. 

There's a very convoluted test that I didn't expose in the docs too for an example of how to go over the top with this.

acozine:
+1 The name/surname example demonstrates the usefulness of this filter and is easy to understand. People who want this filter for more complicated use cases should be able to extrapolate from the simple use case.

davidism:
I know this has been sitting for a long time, but now that I've been focusing on Jinja I've been thinking about it more. While I appreciate the work that went into this, I'm leaning towards not including it in Jinja.

These types of "functional" filters tend to come up when you don't have control of the data being passed to the template (or haven't formatted it appropriately at an earlier time). Contrast with the existing filters, which assume you have the right data and need to perform simple operations to get the right values to output. These requests tend to come from Ansible users a lot because they are trying to do everything in one expression rather than writing custom filters, which is the recommended approach I've heard from multiple Ansible devs.

There's definitely interest in such functional filters, but the problem is that there are a lot of them. Take the [Ramda](https://ramdajs.com/docs/) JavaScript library for example. This is just way too much to open the doors for maintaining in the Jinja code and docs, when it's fairly tangential to the functionality of Jinja itself. I'd encourage someone interested to start a "Jinja-Functional-Filters" extension package to collect these.

PR #909, opened 2018-10-10
Enable a buffer of size 1 in a template stream

tonyfast:
Currently, a `Template.stream` permits a buffer size of greater than one.  In an applications I need to stream one buffer at a time.  This pull request changes the existing logic in `enable_buffering` to stream from the buffer one at a time.

Additionally, should there be a check added for non-integer values in `enable_buffering`?

### DIFF
--- a/jinja2/environment.py
+++ b/jinja2/environment.py
@@ -1260,7 +1260,7 @@ class TemplateStream(object):
 
     def enable_buffering(self, size=5):
         """Enable buffering.  Buffer `size` items before yielding them."""
-        if size <= 1:
+        if size < 1:
             raise ValueError('buffer size too small')
 
         self.buffered = True


davidism:
I'm not clear why you need this. Don't enable buffering, that's equivalent to buffering one item, except without the overhead of the buffering method.

tonyfast:
The use case I have is a for loop that returns large slices of pandas.DataFrame as html tables during each iteration.  Sometimes, the DataFrame nodes in the loop will perform some statistical aggregations at each iteration.  With the current settings I have to receive 2 or more tables at a time or all at once.  

Perhaps I am using the wrong approach.  If so I am sorry for the noise.

davidism:
Iterating over `TemplateStream` yields one item at a time by default, buffering is for when you want to yield more than one item. It's not clear why you call `enable_buffering` if you don't want to use buffering.

tonyfast:
@davidism Thank you for taking the time to help me here; I was clearly using the wrong approach.

Issue #910, opened 2018-10-15
Any appetite to be able to pass a macro for the map() filter?


seaders:
I've basically done this in my code,

template:

    {% macro event_detail(event) -%}
        {{ event.name }}<br />{{ event.startdt }}
    {%- endmacro %}

    {{ events|custom_map(macro=event_detail)|join(', ') }}

code:

    @app.template_filter()
    def extra_map(l, macro):
        return map(macro, l)

It's a pretty simple premise, but works really nice, and makes map much more powerful and flexible.  If there's any interest, I'll do it up proper in `prepare_map` in `filters.py` and PR it.

flexferrum:
Had the same request in C++ implementation: https://github.com/flexferrum/Jinja2Cpp/issues/50

Implemented it as a dedicated `applymacro(value, name, *args, **kwargs)` filter which can be called from `map` like this: `{{ events|map('applymacro' name='event_detail')|join(', ') }}`

davidism:
@seaders seems reasonable.

As with a lot of things in Jinja, I don't know if there's a design reason the `map` filter doesn't already handle callables. Maybe @mitsuhiko can comment.

Quick look at `jinja2.filters.do_map()`, uses `context.environment.call_filter()`. Might be able to detect if the first argument is a callable instead of a string, and use `context.call()` to invoke it.

@flexferrum is there a particular reason you went with a new filter instead of extending the map filter?

flexferrum:
@davidism, I see at least two such reason:

1. You can apply macro as a filter for the single arbitrary value (not only for list/tuple)
2. You can keep the current sementic of the `map` filter which can apply only filters, but not any callable.

davidism:
See https://github.com/pallets/jinja/pull/911#issuecomment-549436594 for why I've decided to close this.

PR #911, opened 2018-10-16
Add applymacro filter.

seaders:
Based on https://github.com/pallets/jinja/issues/910

Any feedback for the documentation, or further tests absolutely welcome.

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -949,6 +949,50 @@ def do_attr(environment, obj, name):
 
 
 @contextfilter
+def do_applymacro(*args, **kwargs):
+    """Applies a macro by reference instead of directly.
+
+    Calling a macro directly like,
+
+    .. sourcecode:: html+jinja
+
+        {{ event_detail(event) }}
+
+    Is the equivalent with applymacro,
+
+    .. sourcecode:: html+jinja
+
+        {{ event|applymacro(macro='event_detail') }}
+
+    or
+
+    .. sourcecode:: html+jinja
+
+        {{ event|applymacro(macro=event_detail) }}
+
+    Which then allows you to use the macro in other things like map,
+
+    .. sourcecode:: html+jinja
+
+        {{ events|map('applymacro', macro=event_detail)|join(', ') }}
+
+    """
+    context = args[0]
+    args = args[1:]
+
+    macro = kwargs.pop('macro')
+    if isinstance(macro, str):
+        if '.' in macro:
+            templatemod, _, macroname = macro.partition('.')
+            macro = getattr(context.resolve(templatemod), macroname)
+    
+        else:
+            macro = context.resolve(macro)
+
+    return macro(*args, **kwargs)
+
+
+@contextfilter
 def do_map(*args, **kwargs):
     """Applies a filter on a sequence of objects or looks up an attribute.
     This is useful when dealing with lists of objects but you are really
@@ -1149,6 +1193,7 @@ def select_or_reject(args, kwargs, modfunc, lookup_attr):
 
 FILTERS = {
     'abs':                  abs,
+    'applymacro':           do_applymacro,
     'attr':                 do_attr,
     'batch':                do_batch,
     'capitalize':           do_capitalize,
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -1326,20 +1326,22 @@ evaluates into an undefined object:
 List of Builtin Filters
 -----------------------
 
-===================  ===================  ==================  ======================  ====================
+===================  ===================  ===================  ===================  ======================
 Built-in Filters Index
 ==========================================================================================================
-:func:`abs`          :func:`attr`         :func:`batch`       :func:`capitalize`      :func:`center`
-:func:`default`      :func:`dictsort`     :func:`escape`      :func:`filesizeformat`  :func:`first`
-:func:`float`        :func:`forceescape`  :func:`format`      :func:`groupby`         :func:`indent`
-:func:`int`          :func:`join`         :func:`last`        :func:`length`          :func:`list`
-:func:`lower`        :func:`map`          :func:`max`         :func:`min`             :func:`pprint`
-:func:`random`       :func:`reject`       :func:`rejectattr`  :func:`replace`         :func:`reverse`
-:func:`round`        :func:`safe`         :func:`select`      :func:`selectattr`      :func:`slice`
-:func:`sort`         :func:`string`       :func:`striptags`   :func:`sum`             :func:`title`
-:func:`tojson`       :func:`trim`         :func:`truncate`    :func:`unique`          :func:`upper`
-:func:`urlencode`    :func:`urlize`       :func:`wordcount`   :func:`wordwrap`        :func:`xmlattr`
-===================  ===================  ==================  ======================  ====================
+:func:`abs`          :func:`applymacro`   :func:`attr`         :func:`batch`        :func:`capitalize`
+:func:`center`       :func:`default`      :func:`dictsort`     :func:`escape`       :func:`filesizeformat`
+:func:`first`        :func:`float`        :func:`forceescape`  :func:`format`       :func:`groupby`
+:func:`indent`       :func:`int`          :func:`join`         :func:`last`         :func:`length`
+:func:`list`         :func:`lower`        :func:`map`          :func:`max`          :func:`min`
+:func:`pprint`       :func:`random`       :func:`reject`       :func:`rejectattr`   :func:`replace`
+:func:`reverse`      :func:`round`        :func:`safe`         :func:`select`       :func:`selectattr`
+:func:`slice`        :func:`sort`         :func:`string`       :func:`striptags`    :func:`sum`
+:func:`title`        :func:`tojson`       :func:`trim`         :func:`truncate`     :func:`unique`
+:func:`upper`        :func:`urlencode`    :func:`urlize`       :func:`wordcount`    :func:`wordwrap`
+:func:`xmlattr`
+
+===================  ===================  ===================  ===================  ======================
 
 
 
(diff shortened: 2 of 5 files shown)


jinja2/filters.py
```diff
@@ -940,6 +940,37 @@ def do_attr(environment, obj, name):
     return environment.undefined(obj=obj, name=name)
 
 
+@contextfilter
+def do_applymacro(*args, **kwargs):
+    """Applies a macro by reference instead of directly.
```
davidism: Use ReStructuredText formatting for docstrings. Render the docs with `pip install pallets-sphinx-themes` then `make html` from the docs folder. You can view the build by opening `_build/html/index.html` in a browser.

jinja2/filters.py
```diff
@@ -940,6 +940,37 @@ def do_attr(environment, obj, name):
...
+
+    Which then allows you to use the macro in other things like map,
+
+    {{ events|map('applymacro', macro=event_detail)|join(', ') }}
+
+    """
+    context = args[0]
+    args = args[1:]
+
+    macro = kwargs.pop('macro', None)
```
davidism: Let's have one explicit arg and switch on it's type automatically.

jinja2/filters.py
```diff
@@ -940,6 +940,37 @@ def do_attr(environment, obj, name):
...
+    if macro is None:
+        macro = context.vars[kwargs.pop('macroname')]
+
+    return macro(*args, **kwargs)
```
davidism: Is this the correct way to call macros? This probably requires `context.call()` or an equivalent for macros.

jinja2/filters.py
```diff
@@ -1165,6 +1196,7 @@ def select_or_reject(args, kwargs, modfunc, lookup_attr):
     'list':                 do_list,
     'lower':                do_lower,
     'map':                  do_map,
+    'applymacro':           do_applymacro,
```
davidism: Keep list sorted.

jinja2/filters.py
ThiefMaster: Do we even need both options? What's the benefit of a macro name instead of passing a macro? If you want to pick the macro dynamically in the code you can always use `{% set macro = foo if x else bar %}` with foo/bar being macros.

flexferrum: Can the actual macro name come from the outside (as the template param)? I suppose, yes.

seaders: Do you mean one explicit arg, as in `macro` as opposed to a change to the signature of the function, `def do_map(macro, *args, **kwargs)`?

> switch on it's type automatically

I.E. check if it's a `str` and if so look for it in `context.vars`?

jinja2/filters.py
seaders: I'm not sure.  It *works*...  I'd need a pointer to the correct way, if there is one.

jinja2/filters.py
seaders: If we allow strings, we need to do,

    macro = kwargs.pop('macro')
    if isinstance(macro, str):
        if '.' in macro:
            templatemod, _, mname = macro.partition('.')
            macro = getattr(context.vars[templatemod], mname)

        else:
            macro = context.vars[macro]

So let strings be passed, or just allow the macro to be passed in directly, and not mess around with that?

jinja2/filters.py
```diff
@@ -940,6 +940,37 @@ def do_attr(environment, obj, name):
...
+
+    {{ events|map('applymacro', macro=event_detail)|join(', ') }}
+
+    """
+    context = args[0]
+    args = args[1:]
+
+    macro = kwargs.pop('macro', None)
+    if macro is None:
+        macro = context.vars[kwargs.pop('macroname')]
```
davidism: Use `context.resolve(name)` to look up a name.

davidism:
Since this deals with callable objects, you also need to update [`asyncfilters`](https://github.com/pallets/jinja/blob/master/jinja2/asyncfilters.py) with an async variant.

jinja2/filters.py
davidism: It looks like other filters don't use use it even if they take context, so we can skip that for now.

seaders:
@davidism I've no experience with asyncfilters, can you give some pointers as to what I'd need to provide for the async version?

jinja2/filters.py
seaders: Without feedback, I'm marking this as resolved.  If you guys want this removed just say, and I'll pull it, but if you're okay with it, we'll leave it as is.

seaders:
Hey @davidism I've added a best effort async version - https://github.com/pallets/jinja/pull/911/files#diff-c8e259013541ba9ee7ef783bed930fc3R115

But as I said, I've no experience with the async stuff, and from the other async filters,


    @asyncfiltervariant(filters.do_groupby)
    async def do_groupby(environment, value, attribute):
        expr = filters.make_attrgetter(environment, attribute)
        return [filters._GroupTuple(key, await auto_to_seq(values))
                for key, values in filters.groupby(sorted(
                    await auto_to_seq(value), key=expr), expr)]
    
    
    @asyncfiltervariant(filters.do_join)
    async def do_join(eval_ctx, value, d=u'', attribute=None):
        return filters.do_join(eval_ctx, await auto_to_seq(value), d, attribute)
    
    
    @asyncfiltervariant(filters.do_list)
    async def do_list(value):
        return await auto_to_seq(value)
    
    
    @asyncfiltervariant(filters.do_reject)
    async def do_reject(*args, **kwargs):
        return async_select_or_reject(args, kwargs, lambda x: not x, False)

I get that they're await-iterating lists, but I'm not sure exactly what I should be doing for applymacro.

I've added 2 async tests, 1, which is passing, which is just like that regular filter, but just using the async env, and a 2nd, which is failing, which is the map + applymacro test, https://travis-ci.org/pallets/jinja/jobs/442615314#L507

I've tried debugging through it, but I'm just not 100% understanding what's going on, nor what's trying to be achieved.

It's definitely about the warning(s),

locally I get,

    \jinja\jinja2\filters.py:405: RuntimeWarning: coroutine 'wrap_macro_invoke.<locals>.async_invoke' was never awaited
       return text_type(d).join(imap(text_type, value))

and on travis,

    /jinja/jinja2/asyncfilters.py:79: RuntimeWarning: coroutine 'do_applymacro' was never awaited
        return filters.do_join(eval_ctx, await auto_to_seq(value), d, attribute)

I don't know why they're different, from the same code & tests, and I don't know how to resolve them without a bit of help.

seaders:
Looks like it's a bug with asyncfilter's do_map already present. 

In Environment call_filter,
https://github.com/pallets/jinja/blob/master/jinja2/environment.py#L440

    Note that on Python 3 this might return a coroutine in case the
        filter is running from an environment in async mode and the filter
        supports async execution.  It's your responsibility to await this
        if needed.

Both `do_map`s use the "regular" prepare_map, https://github.com/pallets/jinja/blob/master/jinja2/asyncfilters.py#L108 which doesn't respect ^^^ from call_filter, and simply returns it (now a coroutine) in a lambda.

You can test this by adding the below test in test_asyncfilters,

    def test_map_join(env_async):
        tmpl = env_async.from_string('{{ [[1,2], [3], [4,5,6]]|map("sum")|list }}')
        assert tmpl.render() == '[3, 3, 15]'


    ================================== FAILURES ===================================
    ________________________________ test_map_join ________________________________
    
    env_async = <jinja2.environment.Environment object at 0x0000021D8D21A470>
    
        def test_map_join(env_async):
            tmpl = env_async.from_string('{{ [[1,2], [3], [4,5,6]]|map("sum")|list }}')
    >       assert tmpl.render() == '[3, 3, 15]'
    E       AssertionError: assert '[<coroutine ...21D8D284048>]' == '[3, 3, 15]'
    E         - [<coroutine object do_sum at 0x0000021D8D0E9EB8>, <coroutine object do_sum at 0x0000021D8D0E9FC0>, <coroutine object do_sum at 0x0000021D8D284048>]
    E         + [3, 3, 15]
    
    test_asyncfilters.py:180: AssertionError
    ============================== warnings summary ===============================
    <template>:13: RuntimeWarning: coroutine 'do_sum' was never awaited


From looking at the history, it was done at https://github.com/pallets/jinja/commit/f7bb60af00e7c76243c7b4409c3acf1c19c510a7, the async do_map was added, but doesn't check whether needs to check whether what's returned is a coroutine, but doesn't.   I'll open another issue for this.

davidism:
Thanks for working on this!

> I've no experience with asyncfilters

You now have more experience than I do. :wink:

seaders:
> Thanks for working on this!

Appreciated, but no thanks needed, I'll be dead proud to get some of my crappy code into the jinja repo!

> You now have more experience than I do. 😉

Haha!  'Tis so often the way in open-source country!

seaders:
So @davidism (and/or @mitsuhiko ) now that https://github.com/pallets/jinja/pull/911 has gone through, this has been unblocked, and with the changes already approved, can this get in too?

seaders:
@davidism / @mitsuhiko

?

davidism:
Thanks for working on this! I've been giving this some more thought, after recent discussions about other potential filters such as `|flip` (#906 ) and `|urlencodecomponent` (#864). Based on that I'm going to close this for now. It was definitely still helpful to have the discussion as well as discover and fix the async map issue.

It's not exactly the same use, but the decision in #906 is that trying to do too much stuff in filters might be ok, but should be considered on a per-project basis rather than in Jinja core. This filter ends up being a slightly shorter application of a for loop, and if you're in a position to write a macro you're presumably in a position to write a for loop.

It comes at the expense of mixing the context namespace with the filter namespace, which seem to have been designed to be distinct. Despite using "macro", this could really apply any callable in the context. I'm worried this would introduce too many corner cases I can't forsee, like interaction with async, or context/eval/env decorated functions.

PR #918, opened 2018-10-29
parser: Fix typo `self.lineno` -> `token.lineno`

TGogniashvili:


davidism:
Please add a test that would fail with the previous version to show how this fixes the issue.

TGogniashvili:
It was meant to be part of https://github.com/pallets/jinja/commit/ee2d3c43 I guess. There is no trivial case where this code will fail, I was adding custom parser features and this issue came up.

davidism:
You'll need to add a test before I can merge this.

davidism:
You're right, the rest of the code refers to `token.lineno`. I thought this was a bigger change since it affected the parser.

PR #919, opened 2018-10-31
Add support for auto-indented blocks

### LINKED ISSUE #178: Preserving whitespace prefix in multiline strings
In the StringTemplate engine - which I've used in some projects to emit C code - whitespace prefixes are automatically added in the output lines:

```
PrintCFunction(linesGlobal, linesLocal) ::= <<
void foo() {
    if (someRuntimeFlag) {
        <linesGlobal>
        if (anotherRuntimeFlag) {
            <linesLocal>
        }
    }
}
>>
```

When this template is rendered in StringTemplate, the whitespace prefixing the multilined linesGlobal and linesLocal strings, is copied for all the lines emitted. The generated C code is e.g.:

```
void foo() {
    if (someRuntimeFlag) {
        int i;
        i=1;   // <=== whitespace prefix copied in 2nd
        i++;   // <=== and 3rd line
        if (anotherRuntimeFlag) {
            int j=i;
            j++; //  <=== ditto
        }
    }
}
```

I am new to Jinja2 - and tried to replicate this:

```
#!/usr/bin/env python
from jinja2 import Template

linesGlobal='\n'.join(['int i;', 'i=1;'])
linesLocal='\n'.join(['int j=i;', 'j++;'])

tmpl = Template(u'''\
void foo() {
    if (someRuntimeFlag) {
        {{linesGlobal}}
        if (anotherRuntimeFlag) {
            {{linesLocal}}
        }
    }
}
''')

print tmpl.render(
    linesGlobal=linesGlobal,
    linesLocal=linesLocal)
```

...but saw it produce this:

```
void foo() {
    if (someRuntimeFlag) {
        int i;
i=1;
        if (anotherRuntimeFlag) {
            int j=i;
j++;
        }
    }
}
```

...which is not what I want. I managed to make the output emit proper whitespace prefixes with this:

```
...
if (someRuntimeFlag) {
    {{linesGlobal|indent(8)}}
    if (anotherRuntimeFlag) {
        {{linesLocal|indent(12)}}
    }
}
```

...but this is arguably bad, since I need to manually count whitespace for every string I emit...

Is there a better way that I am missing?

tomas-mazak:
Fixes #178

Blocks now support a new syntax `{%* ... %}` that aligns the indentation of
multiline string with the block statement itself. This is especially
useful when templating YAML or other languages where indentation matters. Example:

labels.j2:
```
tla: webtool
env: {{ env }}
```

deployment.yaml.j2:
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    {% include 'labels.j2' %}
  name: webtool
spec:
  selector:
    matchLabels:
      {% include 'labels.j2' %}
  strategy:
    type: Recreate
```
...renders to broken YAML:
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    tla: webtool
env: qa
  name: webtool
spec:
  selector:
    matchLabels:
      tla: webtool
env: qa
  strategy:
    type: Recreate
```

deployment_new_syntax.yaml.j2:
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    {%* include 'labels.j2' %}
  name: webtool
spec:
  selector:
    matchLabels:
      {%* include 'labels.j2' %}
  strategy:
    type: Recreate
```
...renders correctly:
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    tla: webtool
    env: qa
  name: webtool
spec:
  selector:
    matchLabels:
      tla: webtool
      env: qa
  strategy:
    type: Recreate
```

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -630,6 +630,26 @@ def do_indent(
     return rv
 
 
+def do_lineprefix(s, prefix):
+    """Return a copy of the string with each line prepended by the given string.
+
+    :param prefix: String to prepend each line with
+
+    .. versionadded:: 2.11
+        Add block auto-indent feature
+    """
+    newline = u'\n'
+
+    if isinstance(s, Markup):
+        prefix = Markup(prefix)
+        newline = Markup(newline)
+
+    lines = s.splitlines()
+    rv = newline.join(prefix + line if line else line for line in lines)
+
+    return rv
+
+
 @environmentfilter
 def do_truncate(env, s, length=255, killwords=False, end='...', leeway=None):
     """Return a truncated copy of the string. The length is specified
@@ -1208,6 +1228,7 @@ FILTERS = {
     'format':               do_format,
     'groupby':              do_groupby,
     'indent':               do_indent,
+    'lineprefix':           do_lineprefix,
     'int':                  do_int,
     'join':                 do_join,
     'last':                 do_last,
--- a/jinja2/lexer.py
+++ b/jinja2/lexer.py
@@ -487,13 +487,14 @@ class Lexer(object):
             'root': [
                 # directives
                 (c('(.*?)(?:%s)' % '|'.join(
-                    [r'(?P<raw_begin>(?:\s*%s\-|%s)\s*raw\s*(?:\-%s\s*|%s))' % (
+                    [r'(?P<raw_begin>(?:\s*%s\-|[ \t]*%s\*|%s)\s*raw\s*(?:\-%s\s*|%s))' % (
+                        e(environment.block_start_string),
                         e(environment.block_start_string),
                         block_prefix_re,
                         e(environment.block_end_string),
                         e(environment.block_end_string)
                     )] + [
-                        r'(?P<%s_begin>\s*%s\-|%s)' % (n, r, prefix_re.get(n,r))
+                        r'(?P<%s_begin>\s*%s\-|[ \t]*%s\*|%s)' % (n, r, r, prefix_re.get(n,r))
                         for n, r in root_tag_rules
                     ])), (TOKEN_DATA, '#bygroup'), '#bygroup'),
                 # data
--- a/jinja2/parser.py
+++ b/jinja2/parser.py
@@ -118,7 +118,7 @@ class Parser(object):
         nodes.Node.__init__(rv, 'fi%d' % self._last_identifier, lineno=lineno)
         return rv
 
-    def parse_statement(self):
+    def parse_statement(self, line_prefix=None):
         """Parse a single statement."""
         token = self.stream.current
         if token.type != 'name':
@@ -126,15 +126,25 @@ class Parser(object):
         self._tag_stack.append(token.value)
         pop_tag = True
         try:
+            sub = None
             if token.value in _statement_keywords:
-                return getattr(self, 'parse_' + self.stream.current.value)()
+                sub = getattr(self, 'parse_' + self.stream.current.value)()
             if token.value == 'call':
-                return self.parse_call_block()
+                sub = self.parse_call_block()
             if token.value == 'filter':
-                return self.parse_filter_block()
+                sub = self.parse_filter_block()
             ext = self.extensions.get(token.value)
             if ext is not None:
-                return ext(self)
+                sub = ext(self)
+
+            # If auto-indent feature is on '{%* ... %}', wrap statement in a filter
+            if line_prefix:
+                node = nodes.FilterBlock(lineno=token.lineno)
+                node.filter = nodes.Filter(None, 'lineprefix', [nodes.Const(line_prefix)], [], None, None, lineno=token.lineno)
+                node.body = [sub]
+                return node
+            elif sub:
+                return sub
 
             # did not work out, remove the token we pushed by accident
             # from the stack so that the unknown tag fail function can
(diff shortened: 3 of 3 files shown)


tomas-mazak:
This actually only supports blocks now, but if the idea is generally accepted, I am happy to extend support to variables as well

tomas-mazak:
I've updated the PR with variable support, so now `{{* ... }}` syntax is supported as well

Ashish-Bansal:
@davidism Any update on this ? I would love to have this feature merged.

davidism:
There's now no way to trim leading or trailing newlines though, unless I'm missing that you can combine `*` and `-`.

tomas-mazak:
@davidism yeah, I only implemented the feature I immediately needed for now... I am happy to look into extensions if required to get this merged

tomas-mazak:
@davidism Please let me know if you require any amendments to this PR, I am happy to do some extra work on it, but would really like to see this merged.

stan1y:
Does it support `extends` syntax as well?

tomas-mazak:
@stan1y I haven't tested it with extend, but it should work with any kind of block (as it wraps the whole block in a indent filter under the hood)

mildmelon:
I would love to have this feature as well, looks great!

tomas-mazak:
@davidism bump :)

stan1y:
I can confirm that it worked with `extends` as expected. Very useful.

thirtytwobits:
I pulled this into a fork I'm using to generate source code. It's an essential feature in some regards unless someone else knows a better way to do this?

153957:
👍 I like this new functionality!
However, this PR is missing new tests..

tomas-mazak:
@153957 Care to provide more details? I've closed/reopened the PR and the tests are failing now with Python nightly but that doesn't seem to be related to this PR's code ...

153957:
This PR adds new functionality but there are no changes to any files in `/tests/`, should new functionality not be tested?

tomas-mazak:
@153957 ah, thanks for heads up Arne, I will look into that

KlavsKlavsen:
This would be a really really nice feature to have.. We are currently using jinja to template yaml files.. and have hit this exact issue with no solution.. (it works in helm - which have a jinja-like templating format)..

davidism:
Due to #858, which fixed a bug that caused catastrophic backtracking while lexing whitespace, this no longer merges, and it's not trivial for me to adapt this to the new code.

Comments:

* Implementing this by automatically producing a filter call during parsing seems incorrect. Why can't the whitespace be preserved during lexing, as other whitespace handling is done?

* Needs tests. It's possible to apply this to a lot of tokens (`{% raw %}`, `{%`, `{{`, line statements) but it's not clear if all the results will make sense. #858 had to modify the `TOKEN_RAW_BEGIN` rule as well, is that applicable here?

* It would be nice to make this work with whitespace control.

* Maybe it would make more sense to implement this as another environment option similar to `trim_blocks` and `lstrip_blocks`. Maybe `block_indent`? That way you're opting in to the indent behavior everywhere, and you don't need to add new syntax.

davidism:
Specifically regarding the YAML use case, the answer is to use a YAML library, not to write Jinja templates. If you're using Ansible, you should be using their YAML functions, not making complex templates. If you're writing your own tool, use PyYAML directly.

OJFord:
> Specifically regarding the YAML use case, the answer is to use a YAML library, not to write Jinja templates. If you're using Ansible, you should be using their YAML functions, not making complex templates. If you're writing your own tool, use PyYAML directly.

So, if for example a tool like [`helm`](/helm/helm) had decided to use Jinja instead of Go's built-in templating (which of course was never going to happen because it's written in Go, but for sake of example) the advice is 'no don't do that'?

This seems like an unfortunate decision that has implications much wider than YAML - this is desirable not just for 'prettifying' but any language or markup where whitespace is significant, and where they may not be an extant 'library' (e.g. because it's new or niche or specific to the one case), or that may not make sense (as in the `helm` case where the templates are user-supplied).

A while ago I used Python 'f-strings' to template and format generated Python code, which quickly got messy and I wished I'd had the foresight to use Jinja instead, but I couldn't have done, or at least not easily or less messily without this.

davidism:
That's correct, the default advice without knowing more about your specific circumstances and requirements would be "use the right library for the job", and the right library for the job of producing a well-formatted YAML document is a YAML library.

This advice isn't just from me, multiple Ansible maintainers have said that users should be using their YAML functions, not writing templates that produce whole YAML documents themselves.

However, this topic is tangential to this issue, let's not continue it here.

davidism:
Another thing to consider here is that we also need to consider dedent as well as indent. It's common to indent both within the markup you're generating as well as the Jinja blocks, resulting in extra long indentation once the Jinja blocks are gone.

```
def validate(data):
    {% for name in required %}
        if "{{ name }}" not in data:
            raise ValidationError("{{ name }} is required")
    {% endfor %}

    return True
```

This is a contrived example, but the use of indentation to visually distinguish blocks becomes more useful as the template becomes more complex. So what you really want is to ensure that the first non-whitespace character in the contents starts at the same position as the start tag, and then apply that to the rest of the contents.

davidism:
Thanks for working on this! I'm closing it now based on my comments above. Need to find a better solution for this.

PR #920, opened 2018-11-02
Remove reference to deprecated easy_install; use pip instead

jdufresne:
easy_install is deprecated and its use is discouraged by PyPA:

https://setuptools.readthedocs.io/en/latest/easy_install.html

> Warning: Easy Install is deprecated. Do not use it. Instead use pip.

Follow upstream advice and only recommended supported tools. Replace
with pip.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -23,31 +23,11 @@ need `setuptools`_ or `distribute`_ installed at runtime.
 Installation
 ------------
 
-You have multiple ways to install Jinja2.  If you are unsure what to do, go
-with the Python egg or tarball.
+You can install the most recent Jinja2 version using `pip`_::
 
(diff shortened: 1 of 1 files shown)


docs/intro.rst
```diff
@@ -24,21 +24,20 @@ Installation
 ------------
 
 You have multiple ways to install Jinja2.  If you are unsure what to do, go
-with the Python egg or tarball.
+with ``pip`` or  the tarball.
```
ThiefMaster: There's a double space after "or".
Does it even make sense to mention the tarball at all?

jdufresne:
- Updated to remove the tarball section
- Also removed mentioning `easy_install` from the "MarkupSafe Dependency" section

PR #921, opened 2018-11-07
Update utils.py

tsopokis:
open_if_exists should not raise an exception when not a directory error is encountered.


davidism:
Going to close for the same reasons stated in #602.

Your comment in #399 indicated you were bundling something a pex file, which sounds exactly like bundling it in an egg file. If you want a Flask app to run from a single archive, you need to configure its Jinja loader appropriately.

PR #923, opened 2018-11-13
Feature/long integer with underscores

CleoQc:
It is now possible to use the '_' character to make long integers and floats more easily understandable by a user.
100_000_000_000 for example, or 100_000_000_000.000_000_001

I don't think this closes any issue but since I had my hands in regex, it was a good timing,

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -1177,13 +1177,16 @@ for Python objects such as strings and numbers.  The following literals exist:
     arguments to function calls and filters, or just to extend or include a
     template).
 
-42:
-    Integers are whole numbers without a decimal part.
+42 / 123_456:
+    Integers are whole numbers without a decimal part. The '_' character
+    can be used to separate groups for legibility.
 
-42.23 / 42.1e2:
+42.23 / 42.1e2 / 123_456.789:
     Floating point numbers can be written using a '.' as a decimal mark.
     They can also be written in scientific notation with an upper or
-    lower case 'e' to indicate the exponent part.
+    lower case 'e' to indicate the exponent part. The '_' character can
+    be used to separate groups for legibility, but cannot be used in the
+    exponent part.
 
 ['list', 'of', 'objects']:
     Everything between two brackets is a list.  Lists are useful for storing
--- a/jinja2/lexer.py
+++ b/jinja2/lexer.py
@@ -31,9 +31,23 @@ _lexer_cache = LRUCache(50)
 
 # static regular expressions
 whitespace_re = re.compile(r'\s+', re.U)
+newline_re = re.compile(r'(\r\n|\r|\n)')
 string_re = re.compile(r"('([^'\\]*(?:\\.[^'\\]*)*)'"
                        r'|"([^"\\]*(?:\\.[^"\\]*)*)")', re.S)
-integer_re = re.compile(r'(\d+_)*(\d+)')
+integer_re = re.compile(r'(\d+_)*\d+')
+float_re = re.compile(
+    r"""
+    (?<!\.)  # doesn't start with a .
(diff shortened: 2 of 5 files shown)


CleoQc:
@davidism  : how would you prefer to handle this? This feature is limited to Python 3..6 and up. Currently it doesn't degrade nicely but I could fix that. Or ignore it. 
Do you still want to support versions older than 3.6? I'm guessing yes. 

I could make the regex dependent on the python version.  I'm open to suggestions but I won't put any more work into this one until I get your feedback.

davidism:
We could support the underscore syntax for any version by removing them before calling literal_eval. Seems reasonable enough to do that.

CleoQc:
Also of note, this PR contains the changes to https://github.com/pallets/jinja/pull/922 as they are literally on the same lines.

CleoQc:
Finally passes all tests.  Let me know if it's acceptable, please.

CleoQc:
I've changed test_float and test_int but I'm thinking we should probably have a test_template that actually tests the regex.

tests/test_filters.py
```diff
@@ -10,7 +10,7 @@
 """
 import random
 import pytest
-from jinja2 import Markup, Environment
+from jinja2 import Markup, Environment, Template
```
PinkFloyded: Any reason to use `Template(str)` instead of `env.from_string(str)`?

docs/templates.rst
```diff
@@ -1140,11 +1140,23 @@ for Python objects such as strings and numbers.  The following literals exist:
...
-42 / 42.23:
-    Integers and floating point numbers are created by just writing the
-    number down.  If a dot is present, the number is a float, otherwise an
-    integer.  Keep in mind that, in Python, ``42`` and ``42.0``
+42 / 123_456_789:
+    Integer numbers are created by just writing the number down. 
+    If a dot is present, the number will be considered a float.  
+    Keep in mind that, in Python, ``42`` and ``42.0``
     are different (``int`` and ``float``, respectively).
+    When dealing with long integers, it is possible to use the '_' character
```
PinkFloyded: \`\`_\`\`?

docs/templates.rst
```diff
@@ -1140,11 +1140,23 @@ for Python objects such as strings and numbers.  The following literals exist:
...
+    to group the digits for legibility, eg: 100_000_000_000.
```
PinkFloyded: \`\`100_000_000_000\`\`?

CleoQc: Yes, those are now valid in Python. 
https://www.python.org/dev/peps/pep-0515/

tests/test_filters.py
CleoQc: env.from_string(str) didn't check the regex. So if the user were to use an older version of Python, `env.from_string()` was failing.

davidism:
I'll get to this eventually! I'm just focused on Werkzeug right now. :-)

jinja2/lexer.py
```diff
@@ -52,7 +54,7 @@
     del jinja2._identifier
     del _identifier
 
-float_re = re.compile(r'(?<!\.)\d+\.\d+')
+float_re = re.compile(r'(?<!\.)((((\d+_)*)\d+)(((\.\d+)?e-?\d+)|(\.((\d+_)*\d+))+))', re.IGNORECASE)
```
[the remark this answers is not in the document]
CleoQc: Good idea. I'll see if I have time to revisit this one. Not this weekend though.

davidism:
Rebased onto current master + #922 changes. Explained the regex with verbose mode. Allow '_' in any part, which matches Python's grammar. Moved the tests out of filters and into lexer, since that's what this was about, but cleaned up both sets of tests with `pytest.mark.parametrize`.

PR #983, opened 2019-05-06
Add support for a {% debug %} extension tag.

kevin-brown:
_Cloned from #798 and rebased with latest `master`._

----

This dumps the available variables, filters and tests. See Issue #174.
This is roughly equivalent to the Django Template Language {% debug %} tag, and typical usage like this:

```
{% debug %}
````

produces output like this:

```python
{'context': {'_': <function _gettext_alias at 0x7f9ceabde488>,
            'csrf_token': <SimpleLazyObject: 'lfPE7al...q3bykS4txKfb3'>,
            'cycler': <class 'jinja2.utils.Cycler'>,
            ...
            'view': <polls.views_auth.Login object at 0x7f9cea2cbe48>},
'filters': ['abs', 'add', 'addslashes', 'attr', 'batch', 'bootstrap',
            'bootstrap_classes', 'bootstrap_horizontal',
            'bootstrap_inline', ... 'yesno'],
'tests': ['callable', 'checkbox_field', 'defined', 'divisibleby',
        'escaped', 'even', 'iterable', 'lower', 'mapping',
        'multiple_checkbox_field', ... 'string', 'undefined', 'upper']}
```

The output under Python2 and Python3 below 3.4 is slightly less compact, as pprint.format() did not introduce the 'compact' argument before 3.4. 

---

/cc @ShaheedHaque

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,9 @@ Unreleased
 -   Fix a bug causing deadlocks in ``LRUCache.setdefault``. :pr:`1000`
 -   The ``trim`` filter takes an optional string of characters to trim.
     :pr:`828`
+-   A new ``jinja2.ext.debug`` extension adds a ``{% debug %}`` tag to
+    quickly dump the current context and available filters and tests.
+    :issue:`174`, :pr:`798, 983`
 
 
 Version 2.10.3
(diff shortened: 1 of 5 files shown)


davidism:
Changed to rely on `autoescape` rather than calling `escape` unconditionally. This way it's usable outside HTML.

Rebased and added docs and changelog.

PR #984, opened 2019-05-06
fix #884, set commentstring / comments for vim

### LINKED ISSUE #884: vim syntax file: set comments and commentstring
This is not a bug in jinja2 per se, but a minor usability issue with the included vim syntax file.

While the syntax file highlights comments correctly, it does not set the variables for `commentstring` and `comments`.  These help with folding and editing in general. A very simple fix is just to include these 2 lines in the vim syntax file:

```
setlocal commentstring={#%s#}
setlocal comments=s:{#,e:#}
```

tbnorth:
This is the correct thing to do, vim-wise, although whether
it has any impact depends on how the user has folding set up
in vim.  I confirmed that with the change, a folding setup that
uses these variables works, whereas without the change, the same
setup doesn't work.

Change only effects a file that has to be manually copied to the user's
~/.vim/syntax folder, so it won't impact jinja execution in any way.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -153,7 +153,7 @@ Why is there no Python 2.3/2.4/2.5/3.1/3.2 support?
 Python 2.3 is missing a lot of features that are used heavily in Jinja2.  This
 decision was made as with the upcoming Python 2.6 and 3.0 versions it becomes
 harder to maintain the code for older Python versions.  If you really need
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Please use separate branches for unrelated changes.

tbnorth:
Closing because in wrong branch, will move to fresh branch.

davidism:
continued in #990

Issue #557, opened 2016-03-01
Add default to map filter if attribute is not defined


Yannik:
It would be great if the map filter would return an optional default value if the attribute is not defined.

Example:

```
list_of_dicts|map(attribute='version', default='5.6')
```

pief:
+1

esaezgil:
Can anyone review my PR please? :)

DanyC97:
any reason why this issue and PR hasn't been progressed ?

Yannik:
Yeah @davidism, please finally merge this.

esaezgil:
Hi @DanyC97 and @Yannik thanks for raising this! :) it's been so long that the MR has conflicts. I will try to solve them asap

davidism:
The following is not a statement about the issue or PR itself, and I do not want to give the impression that your comments had a positive effect on our future actions with it.

Please recognize that Jinja is one of multiple packages maintained by a few people in their spare time. We are under no timeline nor obligation to merge pull requests. While this may not have been your intention, posting comments that contain no content relevant to the issue but just prod us to take action are not welcome. Please do not do that here or elsewhere in the open source community.

davidism:
@Yannik @DanyC97 The following is not a statement about the issue or PR itself, and I do not want to give the impression that your comments had a positive effect on our future actions with it.

Please recognize that Jinja is one of multiple packages maintained by a few people in their spare time. We are under no obligation to close issues on any specific timeline, although we do try to address things as we can. While this may not have been your intention, posting comments that contain no content relevant to the issue but just prod us to take action are not welcome. Please do not do that here or elsewhere in the open source community.

hsychla:
Is somebody still loooking into this? I reaaaally could have used this right now...

wedi:
Are there any news regarding this feature? Why was the pulled request closed? Should someone else take over or is this just undesired?

esaezgil:
I closed the pull request myself as I do not have any further interest pursuing it. Please feel free to contribute with another PR.

akamac:
I would love to see this implemented

matthagenbuch:
+1

shashanksingh28:
Working on this in Pycon 2019 Sprints

kbiggers:
#1018

flexferrum:
Is it a good idea to add 'default' param to the 'attr' filter as well?

PR #985, opened 2019-05-06
Adding a default parameter to builtin map filter

bkiu:
Working on issue #557 to add a `default` parameter for the `map` filter. 

This required adding a parameter to `make_attrgetter`, which is also used by many other functions. A  `default` parameter could also be used in other filters as well, like `join` or `sum`.

Tests also written and passed.

Written at sprints at Pycon2019


davidism:
Hey, really sorry about this, but I merged #1018 before I realized this PR was earlier. There's a few things in this one that I think make the other a bit cleaner, so I'm going to rebase and merge this one too.

Issue #794, opened 2018-01-08
Loop properties revindex and revindex0 still wrong in Jinja2 2.10


unusual-thoughts:
## Expected Behavior
`loop.revindex` should be 1 when using a reversed iterator of length 1
```python
jinja2.Template("{% for i in lst|reverse %}{{ loop.revindex }}:{{ i }}, {% endfor %}")
      .render(lst=[10])
```
should print out: `'1:10, '`

## Actual Behavior
The index of the only iteration is set to 2
```python
jinja2.Template("{% for i in lst|reverse %}{{ loop.revindex }}:{{ i }}, {% endfor %}")
      .render(lst=[10])
```
prints out: `'2:10, '`

**However**:
```python
jinja2.Template("{% for i in lst|reverse %}{{ loop.revindex }}:{{ i }}, {% endfor %}")
      .render(lst=[10, 11])
```
correctly prints out: `'2:11, 1:10, '`

## Template Code
```jinja
{% for i in lst|reverse %}{{ loop.revindex }}:{{ i }}, {% endfor %}
```

## Your Environment
* Python version: 3.6.4
* Jinja version: 2.10

unusual-thoughts:
Related to https://github.com/pallets/jinja/issues/244#issuecomment-356029542

FreddieChopin:
Also related to #751

asenov:
Fix is so easy

davidism:
@asenov if this is easy for you, it would be helpful if you submitted a patch. We welcome constructive contributions.

asenov:
And my question here is what about the following code and output: 
``
jinja2.Template("{% for i in lst|reverse %}{{ loop.index }}:{{ i }}, {% endfor %}")
      .render(lst=[10, 11])
``
As following: 
``
jinja2.Template("{% for i in lst|reverse %}{{ loop.revindex }}:{{ i }}, {% endfor %}")
      .render(lst=[10, 11])
``
?

wuwentao2000:
got the same length 2 error.
Any updates or workaround about this issue?

aayushuppal:
working on this as part of pycon pallets sprint.

markroth8:
Also seems related to #459

aayushuppal:
Should be fixed in https://github.com/pallets/jinja/pull/993

PR #991, opened 2019-05-06
Fix for Loop properties revindex and revindex0 still wrong in Jinja2 2.10 #794

aayushuppal:
- The approach to set length param is costly and should be avoided. only if someone expects the iterator length should the cost be borne.
- Added test cases to validate rendering and LoopContext behavior


aayushuppal:
withdrawing, will open the PR against branch 2.10.x

PR #993, opened 2019-05-06
Fix for Loop properties revindex and revindex0 still wrong in Jinja2 2.10 #794

### LINKED ISSUE #459: Loop property length does not work in 2.7.3 and 2.8dev0 version with if filter
Loop property length does not work in 2.7.3 and 2.8dev0 version with if filter
In 2.7.3 does not work when the results of if filter > 1 elements
In 2.8dev0 does not work when the results of if filter == 1 element

test template:

```
{% set numberList = [0,10,20,30,40,50,60,70,80,90] %}

{% for number in numberList %}
{% if loop.first %}
length {{ loop.length }} revindex {{ loop.revindex }}
{% endif %}
{{ loop.index }} => {{ number }}
{% endfor %}

{% for number in numberList if number < 50 %}
{% if loop.first %}
length {{ loop.length }} revindex {{ loop.revindex }}
{% endif %}
{{ loop.index }} => {{ number }}
{% endfor %}

{% for number in numberList if number < 10 %}
{% if loop.first %}
length {{ loop.length }} revindex {{ loop.revindex }}
{% endif %}
{{ loop.index }} => {{ number }}
{% endfor %}
```

results 2.7.3

```
length 10 revindex 10
1 => 0
2 => 10
3 => 20
4 => 30
5 => 40
6 => 50
7 => 60
8 => 70
9 => 80
10 => 90

length 4 revindex 4
1 => 0
2 => 10
3 => 20
4 => 30
5 => 40

length 1 revindex 1
1 => 0
```

result 2.8dev0

```
length 10 revindex 10
1 => 0
2 => 10
3 => 20
4 => 30
5 => 40
6 => 50
7 => 60
8 => 70
9 => 80
10 => 90

length 5 revindex 5
1 => 0
2 => 10
3 => 20
4 => 30
5 => 40

length 2 revindex 2
1 => 0
```

### LINKED ISSUE #751: Loop variable behaviour off if supplied with a generator
If jinja2 is iterating over a generator rather than a simple list, weird things happen. #555 is an example of this, but a simpler one I just ran in to is using the loop.revindex[0] special variables for length-1 lists. For example, in Python 3, if a list is reversed (using a filter in the template or a function in the script) then the loop indices are off-by-1.

Any other function/filter that yields a generator appears to be similarly affected, which is practically everything in Python 3 (e.g. map, zip).

My suggestion is for jinja to evaluate loop arguments once to form a list, iterate over that and hang the memory usage (which is default Python 2 behaviour), but that's not ideal..

---
```
import jinja2, sys

t = jinja2.Template(
"""
{%- for _ in l -%}
{{ loop.revindex }}, {{ loop.revindex0 }}, {{ loop.index }}, {{ loop.index0 }}
{%- endfor -%}

""")

print(t.render({'l' : [0]}))
print(t.render({'l' : reversed([0])}))
print(t.render({'l' : list(reversed([0]))}))

print(jinja2.__version__)
print(sys.version)
```
Python 3:
```
1, 0, 1, 0
2, 1, 1, 0
1, 0, 1, 0
2.9.6
3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609]
```
Python 2:
```
1, 0, 1, 0
1, 0, 1, 0
1, 0, 1, 0
2.9.6
2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609]
```

### LINKED ISSUE #794: Loop properties revindex and revindex0 still wrong in Jinja2 2.10

aayushuppal:
- The approach to set length param is costly and should be avoided. only if someone expects the iterator length should the cost be borne.
- Added test cases to validate rendering and LoopContext behavior

fixes #459
fixes #751
fixes #794

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,9 @@ Unreleased
     environment's ``undefined`` class. Omitting the ``else`` clause is a
     valid shortcut and should not raise an error when using
     :class:`StrictUndefined`). :issue:`710`, :pr:`1079`
+-   Fix behavior of ``loop`` control variables such as ``length`` and
+    ``revindex0`` when looping over a generator. :issue:`459, 751, 794`,
+    :pr:`993`
 
 
 Version 2.10.3
--- a/jinja2/runtime.py
+++ b/jinja2/runtime.py
@@ -418,19 +418,17 @@ class LoopContext(LoopContextBase):
     def __init__(self, iterable, undefined, recurse=None, depth0=0):
         LoopContextBase.__init__(self, undefined, recurse, depth0)
         self._iterator = iter(iterable)
-
-        # try to get the length of the iterable early.  This must be done
-        # here because there are some broken iterators around where there
-        # __len__ is the number of iterations left (i'm looking at your
-        # listreverseiterator!).
-        try:
-            self._length = len(iterable)
-        except (TypeError, AttributeError):
-            self._length = None
+        self._iterations_done_count = 0
+        self._length = None
         self._after = self._safe_next()
 
     @property
     def length(self):
+        """
+        Getting length of an iterator is a costly operation which requires extra memory
+        and traversing in linear time. So make it an on demand param that iterates from
+        the point onwards of the iterator and accounts for iterated elements.
+        """
         if self._length is None:
             # if was not possible to get the length of the iterator when
             # the loop context was created (ie: iterating over a generator)
@@ -438,8 +436,7 @@ class LoopContext(LoopContextBase):
             # length of that + the number of iterations so far.
             iterable = tuple(self._iterator)
             self._iterator = iter(iterable)
-            iterations_done = self.index0 + 2
-            self._length = len(iterable) + iterations_done
+            self._length = len(iterable) + self._iterations_done_count
         return self._length
 
     def __iter__(self):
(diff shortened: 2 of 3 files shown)


aayushuppal:
Issue & Fix summary
- The `LoopContext` `length` property implementation was **incorrect** & **suboptimal** which in turn messed up the values of `revindex` `revindex0`
- Fixed `LoopContext` `length` property implementation
- Added testcases

PR #994, opened 2019-05-07
exclude globals from find_undeclared_variables

### LINKED ISSUE #931: meta.find_undeclared_variables returns "range" as undeclared variable
Given the following text snippet:
```SQL
SELECT date, sku, {% for lag in range(1, 5) %} a {% endfor %} FROM `{{project}}.{{dataset}}.targets`
```
I'd expect `meta.find_undeclared_variables` to return only `project` and `dataset`.
In fact, `range` is also returned as undeclared variable.

* Python version: 3.6
* Jinja version: 2.10

bkiu:
Added a check for global variables when running `find_undeclared_variables`. Tests updated to reflect the change as well.

Written at PyCon 2019 Sprints.

fixes #931


yzfieldsky:
When does 2.11.0 milestone come out?

PR #996, opened 2019-05-07
Always include context in macros

tdhooper:
Addresses #995, when using this instead of import `with context` everywhere, I see a 50% speed improvement.

This change is a proof of concept, if there are better approaches we should be using, I'd like to know. If this might break something, I'd really like to know!


davidism:
Thanks for working on this. I don't think this is the correct change, as context is not supposed to be available everywhere. I'd rather identify the root issue with the speed of including context and address it directly.

PR #997, opened 2019-05-07
Add ChainableUndefined allowing getattr & getitem

### LINKED ISSUE #977: [Feature Request] default filter on dictionnary and undefined parent key
## Expected Behavior

If the parent key `non_existent_key` does not exist, the print statement should print the default value `string`.

For details (and some workarounds found by the community), see:
- https://stackoverflow.com/questions/19614027/jinja2-template-variable-if-none-object-set-a-default-value
- https://github.com/ansible/ansible/issues/30944

## Actual Behavior

If the parent key does not exist, it throws an error.

## Template Code
```jinja
{{ mydict.non_existent_key.var | default("string") }}
```

EtiennePelletier:
Allows using default values with chains of items or attributes that may
contain undefined values without raising a jinja2.exceptions.UndefinedError.

```python
>>> import jinja2
>>> env = jinja2.Environment(undefined=jinja2.ChainableUndefined)
>>> env.from_string("{{ foo.bar['baz'] | default('val') }}").render()
'val'
```

Fixes #977

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ dist/
 .tox/
 .cache/
 .idea/
+env/
 venv/
 venv-*/
 .coverage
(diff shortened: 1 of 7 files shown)


webknjaz:
@webknjaz

PR #1000, opened 2019-05-09
Fix deadlock in LRUCache and improve test coverage

EtiennePelletier:
I was adding tests for Undefined and other utils by running coverage and finding which lines did not have test coverage, then I found there is a deadlock bug in the LRUCache's setdefault method.

As documented in the commit: setdefault was acquiring write_lock, then calling getitem and also
potentially setitem, which both try to acquire the write lock.
Minimal snippet to reproduce the bug with Python 3.7.3 and Jinja 2.10.1:
```python
from jinja2.utils import LRUCache
x = LRUCache(1)
x.setdefault(1)  # <-- deadlock :(
```

It's fixed now and we have better test coverage :smile:

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -8,15 +8,18 @@
     :copyright: (c) 2017 by the Jinja Team.
     :license: BSD, see LICENSE for more details.
 """
+
+from collections import deque
 import gc
+import pickle
+import random
 
 import pytest
 
-import pickle
-
+from jinja2._compat import string_types, range_type
 from jinja2.utils import LRUCache, escape, object_type_repr, urlize, \
-     select_autoescape
-
+     select_autoescape, generate_lorem_ipsum, missing, consume
+from markupsafe import Markup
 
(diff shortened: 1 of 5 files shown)


tests/test_utils.py
```diff
@@ -44,6 +47,50 @@ def test_pickleable(self):
...
+
+    def test_items(self):
+        """Test various items, keys, values and iterators of LRUCache."""
+        d = LRUCache(3)
+        d["a"] = 1
+        d["b"] = 2
+        d["c"] = 3
+        assert d.items() == list(d.iteritems()) == [('c', 3), ('b', 2), ('a', 1)]
+        assert d.keys() == list(d.iterkeys()) == ['c', 'b', 'a']
+        assert d.values() == list(d.itervalue()) == [3, 2, 1]
```
kevin-brown: This was renamed in a PR that was more recent than this one, so this should now be `itervalues()` (to align with the stdlib).

Issue #999, opened 2019-05-09
LRUCache's itervalue method should be renamed itervalues


EtiennePelletier:
The LRUCache class in jinja2.utils has a method called `itervalue`. To be consistent with its functionality and the language as a whole, it should be updated to `itervalues`.

I don't believe it's ever used in the package and is currently untested (I'm adding a test for it live). To avoid breaking functionality, I believe we should for now create an `itervalues` method that does the same thing, and add a deprecation warning to `itervalue`.

## Snippet Code
```python
class LRUCache(object):
    """A simple LRU Cache implementation."""

    def items(self):
        """Return a list of items."""
        result = [(key, self._mapping[key]) for key in list(self._queue)]
        result.reverse()
        return result

    def iteritems(self):
        """Iterate over all items."""
        return iter(self.items())

    def values(self):
        """Return a list of all values."""
        return [x[1] for x in self.items()]

    def itervalue(self):
        """Iterate over all values."""
        return iter(self.values())

    def keys(self):
        """Return a list of all keys ordered by most recent usage."""
        return list(self)

    def iterkeys(self):
        """Iterate over all keys in the cache dict, ordered by
        the most recent usage.
        """
        return reversed(tuple(self._queue))
```

## Your Environment
* Python version: 3.7.3
* Jinja version: 2.10.1

jab:
Agreed, seems like this was just a typo.

Here is a link to the offending line directly in the (latest version of the) source:
https://github.com/pallets/jinja/blob/30685be0896119b3d6cd0f3d5baa8a616cfb8e14/jinja2/utils.py#L458

After renaming, should the fix for this also add an `itervalue` method that logs a deprecation warning before calling `itervalues`, so as to avoid making a breaking change, but also allow removing `itervalue` in a future major release?

davidism:
Be sure to issue a deprecation warning for the old name.

gokcegrbl:
I will work on this issue today.

PR #1008, opened 2019-05-31
Added new method itervalues() to LRUCache and deprecated itervalue()

gokcegrbl:
Added new method itervalues() to LRUCache and deprecated itervalue()

(Fixes pallets/jinja#999)

### DIFF
--- a/jinja2/utils.py
+++ b/jinja2/utils.py
@@ -11,6 +11,7 @@
 import re
 import json
 import errno
+import warnings
 from collections import deque
 from threading import Lock
 from jinja2._compat import text_type, string_types, implements_iterator, \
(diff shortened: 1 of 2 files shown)


gokcegrbl:
The Travis build for the python-nightly is failing with "ERROR: InvocationError for command /home/travis/build/pallets/jinja/.tox/py/bin/coverage run -p -m pytest (exited with code -11)". Looks like an unrelated issue.

lordmauve:
Signal 11 is segfault so this looks like a random CPython nightly issue.

jinja2/utils.py
```diff
@@ -456,6 +457,13 @@ def values(self):
         return [x[1] for x in self.items()]
 
     def itervalue(self):
+        """Iterate over all values."""
+        warnings.warn(DeprecationWarning(
+            '"itervalue()" is deprecated, use "itervalues()" instead.'
```
davidism: This should mention that it will be removed in version 2.12. Check out Werkzeug for an example of how I word deprecation warnings, this is pretty close already.

Issue #273, opened 2013-10-30
Documentation: Make it clear string methods are available


pas256:
Based on:

https://github.com/mitsuhiko/jinja2/pull/247

Plain string methods are available on string variables. Please make this abundantly clear in the documentation:

http://jinja.pocoo.org/docs/templates/

jeffwidman:
@pas256 Mind submitting a PR with your suggested improvements?

kojiromike:
It's hard to say what would be a good writeup of the issue. The documentation says things like

> …don’t use builtin Python constructs such as str.format or the string modulo operator (%).

That clearly imply it's _possible_ to use builtin Python constructs (such as the ever popular `str.split`), but afaict it's never explicitly said anywhere that such Python constructs are allowed.

edgan:
I was already using python methods in my jinja from having run across examples, but I just realized that it wasn't just the few I was already using. This is huge, and opens doors I didn't know existed. There should be a dedicated section with examples.

PR #1009, opened 2019-05-31
adds documentation on use of python methods in expressions

shamsimam:
Fixes https://github.com/pallets/jinja/issues/273

Here is a screenshot of how the page looks:

jinja-python-methods-doc

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -1309,11 +1309,33 @@ The general syntax is ``<do something> if <something is true> else <do
 something else>``.
 
 The `else` part is optional.  If not provided, the else block implicitly
-evaluates into an undefined object::
+evaluates into an undefined object:
 
 .. sourcecode:: jinja
 
     {{ '[%s]' % page.title if page.title }}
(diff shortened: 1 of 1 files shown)


docs/templates.rst
```diff
@@ -1309,13 +1309,35 @@ The general syntax is ``<do something> if <something is true> else <do
 something else>``.
 
 The `else` part is optional.  If not provided, the else block implicitly
-evaluates into an undefined object::
+evaluates into an undefined object:
```
jab: (nit: the double-colon here can stay, see e.g. https://quick-sphinx-tutorial.readthedocs.io/en/latest/rst.html ("Code blocks are introduced by a double-colon and are indented"))

shamsimam: The current syntax does not compile correctly, it can be seen in the production page at http://jinja.pocoo.org/docs/2.10/templates/#if-expression 

Screen Shot 2019-06-01 at 10 12 42 AM

Issue #416, opened 2015-03-05
Python 3 compatibility of examples


untitaker:
The documentation examples should work under both versions if feasible (and equally performant).

techtonik:
http://jinja.pocoo.org/docs/dev/templates/#for

kooky:
https://github.com/kooky/jinja2/commit/8033d4f59e1bae614556b52a63f6ee9ca8529fc8  is one fix.

Cabalist:
I think this should be reevaluated as Python3 has a much higher acceptance rate these days.  Especially for beginners.  The pull request makes sense.

davidism:
I would happily accept PRs that make examples work in 2 and 3. It doesn't have to be everything in one go.

techtonik:
One of the ways to fix this is to make examples run as a part of test suite.

hugovk:
https://github.com/pallets/jinja/pull/1010 ("Python3-ize the examples") has been merged. Can this be closed?

jab:
Looks like it, thanks for checking, @hugovk!

PR #1010, opened 2019-05-31
docs: Python3-ize the examples

dvamin:
This commit updates the examples to conform to Python 3 instead
of Python 2, as Python 3 is more acceptable these days.


dvamin:
Ref: https://github.com/pallets/jinja/issues/416

PR #1011, opened 2019-05-31
Documentation improvement, write a section about nested template extends.

### LINKED ISSUE #928: Cannot reach multiple levels up in template hierarchy with super
I asked about this on `#pocoo` and got no answer.  I'm fairly sure it's not possible, which means this is probably a feature request?

There does not not seem to be any way to get multiple-levels up in a ``super()`` call in a template.  That is, if I have template C that extends B and B extends A, I want a super from C that skips the block in question from B and goes straight to the super from A.

(I know one answer is "don't do the super in B", but in this case it's impossible because A and B are from an external package)

### a.templ

```jinja
...
{% block body %}
  {% for complex in complicated %}
    <span>
    ... something complex using {{ complex }} that I don't want to have to reproduce...
    <span>
  {% endfor %}
{% endblock %}
...
```

### b.templ

```jinja
{% extends "a.templ" %}
{% block body %}
  <div class="something_I_dont_need">
    {{ super() }}
  </div>
{% endblock %}
```

### c.templ

```jinja
{% extends "a.templ" %}
{% block body %}
  <div class="something_I_actually_DO_need">
    {{ super(2) }}  # Or maybe {{ super("a.templ") }} ?
  </div>
{% endblock %}
```

## Actual Behavior

Impossible to do, because `c.templ` has no access to the "body" block from `a.templ` without going through `b.templ` (which has parts I don't want)

scrosby:
Fixes #928

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -430,6 +430,36 @@ This gives back the results of the parent block::
     {% endblock %}
 
 
+Nesting template extends.
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the case of multiple levels of %extends, the childmost {% block %} substitutes into the parent template. For instance,
+
+::
+
+    #a1.tmpl
+    From A1: {% block body %} Want-this-to-be-included. {% endblock %}
+
+    #a2.tmpl
+    {% extends "a1.tmpl" %}
+    {% block body %} Want-this-to-be-skipped. {{super()}} {% endblock %}
+
+    #a3.tmpl
+    {% extends "a2.tmpl" %}
+    {% block body %} Some text {% endblock %}
+
+    #a4.tmpl
+    {% extends "a2.tmpl" %}
+    {% block body %} From A4: {{super.super()}} {% endblock %}
+
+
+Rendering `a2.tmpl` will output `From A1: Want-this-to-be-skipped`. The text `Want-this-to-be-included.` is rendered because the `super()` call render as the text from the parent template into the block, which will then be substitute into topmost template.
+
+Rendering `a3.tmpl` will output `From A1: Some Text`. We substitute the block text into the parent template.
+
+Rendering `a4.tmpl` will output `From A1: From A4: Want-this-to-be-included.` Here, we are accessing the `super` property from the `super` object to get access to the text in grandparent's block. Had we just done `super()`, this would have returned `From A1: From A4: Want-this-to-be-skipped. Want-this-to-be-included.`.
+
+
 Named Block End-Tags
 ~~~~~~~~~~~~~~~~~~~~
 


docs/templates.rst
```diff
@@ -430,6 +430,36 @@ This gives back the results of the parent block::
     {% endblock %}
 
 
+Nesting template extends.
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the case of multiple levels of %extends, the childmost {% block %} substitutes into the parent template. For instance,
```
jab:
```suggestion
In the case of multiple levels of ``{% extends %}``,
``super`` references may be chained (as in ``super.super``)
to walk up the inheritance tree.

For example::
```

jab: > the childmost {% block %} substitutes into the parent template

- missing double-backticks around `{% blocks %}`
- I find the reference to "parent template" potentially unclear when there is a 3-level tree. Whose parent? If the childmost template's, s/the/its would clarify.
- Isn't the chainability of `super` references the motivation behind this PR / shouldn't it be mentioned here?

docs/templates.rst
```diff
@@ -430,6 +430,36 @@ This gives back the results of the parent block::
...
+    {% block body %} Some text {% endblock %}
+
+    #a4.tmpl
+    {% extends "a2.tmpl" %}
+    {% block body %} From A4: {{super.super()}} {% endblock %}
+
+
+Rendering `a2.tmpl` will output `From A1: Want-this-to-be-skipped`. The text `Want-this-to-be-included.` is rendered because the `super()` call render as the text from the parent template into the block, which will then be substitute into topmost template.
+
+Rendering `a3.tmpl` will output `From A1: Some Text`. We substitute the block text into the parent template.
```
jab: The single backticks have no effect, rst uses double-backticks.

docs/templates.rst
```diff
@@ -430,6 +430,36 @@ This gives back the results of the parent block::
...
+    #a3.tmpl
+    {% extends "a2.tmpl" %}
+    {% block body %} Some text {% endblock %}
+
+    #a4.tmpl
+    {% extends "a2.tmpl" %}
+    {% block body %} From A4: {{super.super()}} {% endblock %}
+
+
+Rendering `a2.tmpl` will output `From A1: Want-this-to-be-skipped`. The text `Want-this-to-be-included.` is rendered because the `super()` call render as the text from the parent template into the block, which will then be substitute into topmost template.
```
jab: All the other lines in these docs wrap at 78 characters.

docs/templates.rst
```diff
@@ -430,6 +430,36 @@ This gives back the results of the parent block::
...
+    {% extends "a1.tmpl" %}
+    {% block body %} Want-this-to-be-skipped. {{super()}} {% endblock %}
+
+    #a3.tmpl
+    {% extends "a2.tmpl" %}
+    {% block body %} Some text {% endblock %}
+
+    #a4.tmpl
+    {% extends "a2.tmpl" %}
+    {% block body %} From A4: {{super.super()}} {% endblock %}
```
jab: This example is a great start. I wonder if it can be made any more minimal in such a way that it increases clarity.

jab:
@scrosby I merged your commit from this PR (b5c6c9c39daa6d758ce75fac1756d92baad510a6) into master in #1015 (after making the touchups discussed above, since it sounded like you'd be too busy to do another round of changes). Thanks for the PR!

PR #1012, opened 2019-05-31
Remind not to put block in if statements in doc

### LINKED ISSUE #423: Blocks inside If/Else and includes
As I had to find out through <a href="https://stackoverflow.com/questions/22655031/jinja2-how-to-put-a-block-in-an-if-statemen">stackexchange</a> blocks ({% block %}) can not be inside a else/if block. This is really confusing -- never did I see anything that is not obeying an if statement -- especially since the documentation does not mention this. 

Another problem that I encountered which is (not clearly enough?) documented: if a block is embedded in a file, which is included, for example into base.html, which than is extended by index.html, the index.html can not modify the content of the mentioned block.

yongbozhang:
Fix #423

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -390,7 +390,9 @@ this template "extends" another template.  When the template system evaluates
 this template, it first locates the parent.  The extends tag should be the
 first tag in the template.  Everything before it is printed out normally and
 may cause confusion.  For details about this behavior and how to take
-advantage of it, see :ref:`null-master-fallback`.
+advantage of it, see :ref:`null-master-fallback`. Also a block will always be
+filled in regardless of whether the surrounding condition is evaluated to be true 
+or false.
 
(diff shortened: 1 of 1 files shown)


docs/templates.rst
```diff
@@ -390,7 +390,9 @@ this template "extends" another template.  When the template system evaluates
 this template, it first locates the parent.  The extends tag should be the
 first tag in the template.  Everything before it is printed out normally and
 may cause confusion.  For details about this behavior and how to take
-advantage of it, see :ref:`null-master-fallback`.
+advantage of it, see :ref:`null-master-fallback`. Also a block will always be
+filled in regardless of whether the surrounding condition is evaluted to be true
```
jab: typo: evaluated

jab:
Thanks for contributing, @yongbozhang!

PR #1013, opened 2019-05-31
Add ability to specify multiple attributes in sort

### LINKED ISSUE #670: Allow multiple attributes in `sort` filter
`foo|sort(attribute='name,id')` is quite useful in some cases.

The current workaround (using the fact that sorting is stable) looks extreme awful and is not intuitive: `foo|sort(attribute='id')|sort(attribute='name')`

mayuropensource:
Added ability to specify multiple attributes in sort:
e.g. foo|sort(attribute='name,id').

This fixes #670

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -84,6 +84,44 @@ def make_attrgetter(environment, attribute, postprocess=None):
     return attrgetter
 
 
+def make_multi_attrgetter(environment, attribute, postprocess=None):
+    """Returns a callable that looks up the given comma separated
+    attributes from a passed object with the rules of the environment.
+    Dots are allowed to access attributes of each attribute.  Integer
+    parts in paths are looked up as integers.
+
+    The value returned by the returned callable is a list of extracted
+    attribute values.
+
+    Examples of attribute: "attr1,attr2", "attr1.inner1.0,attr2.inner2.0", etc.
+    """
+    def _prepare_attribute_parts(attr):
+        if attr is None:
+            return []
+        elif isinstance(attribute, string_types):
+            return [int(x) if x.isdigit() else x for x in attr.split('.')]
+        else:
+            return [attr]
+
+    attribute_parts = attribute.split(',') if isinstance(attribute, string_types) else [attribute]
+    attribute = [_prepare_attribute_parts(attribute_part) for attribute_part in attribute_parts]
+
+    def attrgetter(item):
+        items = [None] * len(attribute)
+        for i, attribute_part in enumerate(attribute):
+            item_i = item
+            for part in attribute_part:
+                item_i = environment.getitem(item_i, part)
+
+            if postprocess is not None:
+                item_i = postprocess(item_i)
+
+            items[i] = item_i
+        return items
+
+    return attrgetter
+
+
 def do_forceescape(value):
     """Enforce HTML escaping.  This will probably double escape variables."""
     if hasattr(value, '__html__'):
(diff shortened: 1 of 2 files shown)


jinja2/filters.py
```diff
@@ -84,6 +84,44 @@ def attrgetter(item):
...
+    attributes from a passed object with the rules of the environment.
+    Dots are allowed to access attributes of each attribute.  Integer
+    parts in paths are looked up as integers.
+
+    The value returned by the returned callable is a list of extracted
+    attribute values.
+
+    Examples of attribute: "attr1,attr2", "attr1.inner1.0,attr2.inner2.0", etc.
+    """
+    def _prepare_attribute_parts(attr):
```
kevin-brown: I see that this logic is basically the same logic from `make_attrgetter`. Would it be possible to abstract this out and use it in both places?

That way if we need to adjust this in the future, we won't have to remember to do it in both places.

artofcomputerprogramming: Sure, will do

mayuropensource: Will do

mayuropensource: @kevin-brown pushed a fix.

jinja2/filters.py
```diff
@@ -84,6 +79,45 @@ def attrgetter(item):
...
+        return items
+
+    return attrgetter
+
+
+def _prepare_attribute_parts(attr):
+    if attr is None:
+        return []
+    elif isinstance(attr, string_types):
+        return [int(x) if x.isdigit() else x for x in attr.split('.')]
```
jab: Drive-by review question: Is `x.isdigit()` correct here? What about integer indexes greater than 9? E.g. '10' should be converted to 10, but `'10'.isdigit() == False`, right? What about negative indexes?

Also, right after http://jinja.pocoo.org/docs/2.10/templates/ explains that the dot syntax can be used for attribute *or* item lookup, it says “Additionally, the attr() filter only looks up attributes.” The docstrings here only mention attribute lookup but in fact support item lookup as well, which is inconsistent, is that right?

ThiefMaster: `isdigit()` checks if a string only contains digits

jab: So it does! Thanks for the correction. (Misleading that it isn’t named something plural like `all_digits`.) in any case, it’s still not a sufficient implementation of “is_integer?” if negatives should be supported: `'-1'.isdigit() == False`. Should they be? Docstring need to mention one way or the other?

jab: So it does! Thanks for the correction. (Misleading that it isn’t named something plural like `all_digits`.) in any case, it’s still not a sufficient implementation of “is_integer?” if negatives should be supported: `'-1'.isdigit() == False`. Should they be? Docstring need to mention one way or the other?

kevin-brown: I'm all in favor of keeping this consistent with the old implementation (supporting keys, and not supporting negative indexes) and looking to support future improvements in later pull requests.

jab: Sounds good. If negative indexes should be unsupported now but supported in future, it does seem worth mentioning that they’re not currently supported in the docstring.

And also clarifying in the docstring that this does attribute and item lookup.

mayuropensource:
Hi, 

Is this pr good to go? Do I need to make doc changes to get this pushed? I can open a separate issue to make doc changes.

Thanks,
Mayur

mayuropensource:
@davidism let me know if I need to make any other changes. 

Thanks,
Mayur

kevin-brown:
Thanks for sticking with us through the process!

I've merged this in since the test failures are against Python 3.8 and are unrelated to these changes (there's an issue in coveragepy).

PR #1015, opened 2019-06-06
Document nested extends and super.super

### LINKED ISSUE #928: Cannot reach multiple levels up in template hierarchy with super

jab:
Subsumes #1011. Fixes #928.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2,6 +2,7 @@ API
 ===
 
 .. module:: jinja2
+    :noindex:
     :synopsis: public Jinja2 API
 
 This document describes the API to Jinja2 and not the template language.  It
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -435,6 +435,36 @@ This gives back the results of the parent block::
     {% endblock %}
 
 
+Nesting template extends.
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the case of multiple levels of %extends, the childmost {% block %} substitutes into the parent template. For instance,
+
+::
+
+    #a1.tmpl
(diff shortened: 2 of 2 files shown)


jab:
All Travis jobs [succeeded](https://travis-ci.org/pallets/jinja/builds/542065273) except Python nightly (CPython 3.8.0a4+), which was already failing in [the latest build of master](https://travis-ci.org/pallets/jinja/builds/539807265), and I don't think these changes are related, so I'm going to merge this.

PR #1016, opened 2019-06-08
Made Cycler.next() act as intended

Mezzomaniac:
Previously, `c = Cycler(‘odd’, ‘even’); c.next(); c.next(); c.current` would return ‘odd’, ‘even’, ‘odd’. Now it returns ‘odd’, ‘even’, ‘even’ as expected.

### DIFF
--- a/jinja2/utils.py
+++ b/jinja2/utils.py
@@ -586,7 +586,7 @@ class Cycler(object):
 
     def reset(self):
         """Resets the cycle."""
-        self.pos = 0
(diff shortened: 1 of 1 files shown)


davidism:
It looks like the docs were already trying to describe the behavior, but were still unclear:

> Goes one item ahead and returns the then-current item.

I've reworded the docs to:

> Return the current item, then advance `current` to the next item.

As well as reworked the rest of the `cycler` docs.

PR #1018, opened 2019-06-10
add default to map filter

### LINKED ISSUE #557: Add default to map filter if attribute is not defined

kbiggers:
added ability to pass default entry to map filter if attribute is not defined. If default is not defined it defaults to an empty string.

fixes #557

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -59,7 +59,7 @@ def ignore_case(value):
     return value.lower() if isinstance(value, string_types) else value
 
 
-def make_attrgetter(environment, attribute, postprocess=None):
+def make_attrgetter(environment, attribute, default=None, postprocess=None):
     """Returns a callable that looks up the given attribute from a
     passed object with the rules of the environment.  Dots are allowed
     to access attributes of attributes.  Integer parts in paths are
@@ -70,6 +70,9 @@ def make_attrgetter(environment, attribute, postprocess=None):
     def attrgetter(item):
         for part in attribute:
             item = environment.getitem(item, part)
+
+            if isinstance(item, Undefined) and default:
+                item = default
 
         if postprocess is not None:
             item = postprocess(item)
@@ -352,8 +355,9 @@ def _min_or_max(environment, value, func, case_sensitive, attribute):
         return environment.undefined('No aggregated item, sequence was empty.')
 
     key_func = make_attrgetter(
-        environment, attribute,
-        ignore_case if not case_sensitive else None
+        environment,
+        attribute,
+        postprocess=ignore_case if not case_sensitive else None
     )
     return func(chain([first], it), key=key_func)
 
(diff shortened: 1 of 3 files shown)


jinja2/filters.py
```diff
@@ -977,10 +984,17 @@ def do_map(*args, **kwargs):
...
     """
+    try:
+        default = kwargs['default']
+    except KeyError:
+        default = ''
     seq, func = prepare_map(args, kwargs)
     if seq:
         for item in seq:
-            yield func(item)
+            if func(item):
```
davidism: This should be a try catch block. Right now, it's replacing falsey values with the default, which it shouldn't do, but it will still fail if the attribute doesn't exist, which is what the issue was about. Also, it causes the function to be called twice.

jinja2/filters.py
```diff
@@ -977,10 +984,17 @@ def do_map(*args, **kwargs):
 
     .. versionadded:: 2.7
     """
+    try:
```
davidism: This should probably be done in `prepare_map`.

kbiggers: See commit below, moved to prepare_map

jinja2/filters.py
kbiggers: See my below commits. Refactored so the function no longer needs to be called twice, also now checking for undefined type instead of falsey values. Added tests as well.

jinja2/filters.py
```diff
@@ -967,6 +969,13 @@ def do_map(*args, **kwargs):
 
         Users on this page: {{ users|map(attribute='username')|join(', ') }}
 
+    If the list of objects does not contain the given attribute
+    you can specify a default value to be returned.
+
+    .. sourcecode:: jinja
+
+        Users on this page: {{ users|map(attribute='username', default='foobar')|join(', ') }}
```
davidism: "Anonymous" might make more sense for this example.

jinja2/filters.py
```diff
@@ -75,6 +75,8 @@ def make_attrgetter(environment, attribute, postprocess=None):
     def attrgetter(item):
         for part in attribute:
             item = environment.getitem(item, part)
+            if type(item) == Undefined and default:
```
davidism: `isinstance(Undefined)` (or `is Undefined` if we actually care about exact match).

davidism:
Sorry, noticed a couple more review items, looks good overall.

jinja2/filters.py
kbiggers: good idea, fixed in most recent commit.

jinja2/filters.py
kbiggers: good catch, I don't think we actually care about an exact match. Updated in most recent commit.

davidism:
OK, I merged this then noticed #985, which came a month before this and looks incredibly similar. Did I miss something where this was a continuation of that?

Issue #973, opened 2019-04-26
Release a version containing #867


flying-sheep:
Hi! In #867, you fixed some `DeprecationWarning`s that were swamping people’s pytest output.

Unfortunately they still do, since you didn’t put the commit into 2.10.1. Could you please release a version with that fix?

davidism:
Closing as this issue won't affect our release schedule. Thank you for the interest.

flying-sheep:
What is your release schedule?

dhimmel:
Also interested in a release that includes #867, since `jinja2` is the only package we use that [causes warnings](https://travis-ci.com/manubot/manubot/jobs/205309499#L541) in Python 3.7. Simplifying the Python output logs are important for our use case, so these warnings are keeping us using Python 3.6.

jbarreneche:
Hi @davidism ! What's the schedule for a new release that contains that simple fix? (neither of the links you provided help answering that)
The fix has been in master for quite a while and hasn't been backported to 2.10 branch. I understand it's not something critical but it's a very annoying warning for projects using new versions of python. I can back port it to the branch 2.10 branch if that helps.

Thanks!

ThiefMaster:
@jbarreneche a PR (with the fix and a changelog entry) for 2.10 would be welcome and indeed help speed it up!

PR #1019, opened 2019-06-18
Backport python3.7 collections warning fix

jbarreneche:
The fix is already in master but probably worth to backport it to the 2.10.x branch (based on this [comment](https://github.com/pallets/jinja/issues/973#issuecomment-503127082)).
Not sure if fixing the warning also makes jinja 2.10.x Python 3.8 compatible, so I only mentioned removing the python 3.7 deprecation warning.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,13 @@ Jinja Changelog
 ===============
 
 
+Version 2.10.2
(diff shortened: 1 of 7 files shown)


jbarreneche:
Thanks for the quick merge! 🙇

PR #1022, opened 2019-06-21
Correct syntax highlighting of jinja block in docs

jdufresne:


jab:
Looks like this reverts #852 contributed by @ngr and merged by @davidism. Not immediately clear to me why #852 was landed but would be good to know if something's changed in the meantime that makes it worth reverting.

davidism:
The earlier patch was a quick fix since Pygments can't parse `%` string formatting in Jinja files (it's been an open issue for 5 years).

davidism:
This is still technically incorrect, since it's being highlighted as Python not Jinja, but it looks fine so I'll merge it.

PR #1025, opened 2019-06-25
Unpin Pytest on master branch

kevin-brown:
The pytest plugin was "temporarily" pinned many months back. Given that this is not pinned on the stable branch we should be able to remove it here.

There are likely a few other changes required to get the nightly build working, since it appears as though pytest broke during the release of Python 3.8.0a4.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,7 +8,7 @@ envlist =
 passenv = LANG
 usedevelop = true
 deps =
-    pytest>=3,<3.7
+    pytest>=3
     coverage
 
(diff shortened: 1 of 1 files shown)


kevin-brown:
Looks like Coverage is just broken in Python 3.8 and they aren't immediately addressing it: https://github.com/nedbat/coveragepy/commit/e26a9237bae636635440061094cd65ffc40c3cf6

davidism:
We need to add nightly to the allowed failures. Preferrably, we need to migrate to Azure like Flask did.

PR #1030, opened 2019-07-04
Fix pytest unknown mark warnings

rillian:
Clean up the test output with recent versions of `pytest`.


rillian:
The travis failure isn't related. It's the same `InvocationError` on nightly with the coverage tool seen elsewhere.

rillian:
Thanks!

PR #1031, opened 2019-07-06
Fix typo

minho42:


jinja2/runtime.py
```diff
@@ -510,7 +510,7 @@ def __call__(self, *args, **kwargs):
         # check here.
         #
         # This is considered safe because an eval context is not a valid
-        # argument to callables otherwise anwyays.  Worst case here is
+        # argument to callables otherwise anyways.  Worst case here is
```
ThiefMaster: should be "anyway"

https://www.grammarly.com/blog/is-anyways-a-word/

jinja2/bccache.py
```diff
@@ -5,7 +5,7 @@
 
     This module implements the bytecode cache system Jinja is optionally
     using.  This is useful if you have very complex template situations and
-    the compiliation of all those templates slow down your application too
+    the compilation of all those templates slow down your application too
```
jab: s/slow/slows

(the subject is the singular “compilation”, not “templates”)

jinja2/bccache.py
```diff
@@ -5,7 +5,7 @@
 
     This module implements the bytecode cache system Jinja is optionally
```
jab: While you’re in here, “that Jinja optionally uses” instead of “Jinja is optionally using”.

minho42:
@jab 
Thanks for your suggestions.

jinja2/bccache.py
```diff
@@ -3,9 +3,8 @@
     jinja2.bccache
     ~~~~~~~~~~~~~~
 
-    This module implements the bytecode cache system Jinja is optionally
-    using.  This is useful if you have very complex template situations and
-    the compilation of all those templates slow down your application too
+    This module implements the bytecode cache system that Jinja optionally uses.  This is useful if you have very complex template situations and
```
jab: This line is too long. Could you please break this up to respect the max line width here?

minho42: Sorry I missed this. Fixed!

Issue #1033, opened 2019-07-15
missing details in 'sort' filter docs


utoddl:
This is an issue with the "sort" filter's docs. (The code itself seems to work fine.) A couple of extra lines could save people a _lot_ of time. Please consider adding (1) that the sort performed by the sort filter is stable, and (2) that sorting by multiple attributes therefore can be accomplished by
```jinja
    sort(attribute='b') | sort(attribute='a')
```
---

## Expected Behavior
Users read the sort docs, then write the above code.

## Actual Behavior
Users read the sort docs, then write code like the following over the course of more hours than I'd like to admit to accomplish the same thing:
```jinja
  sort(attribute='a') | groupby(attribute='a') | map('last') 
              | map('sort', attribute='b') | list | flatten
```
## Your Environment
Not really relevant, but since you asked:
* Python version: multiple (We're using Ansible in a 800+host environment.)
* Jinja version: surprisingly recent, given the hodge-podge of hosts we're running from.

davidism:
PRs welcome

PR #1035, opened 2019-07-18
document that sort is stable

utoddl:
Addresses issue #1033


utoddl:
PR #1035 submitted.

davidism:
Thanks for starting this. After reviewing the current docs, I ended up rewriting them a lot more to hopefully be more descriptive of how everything actually works.

PR #1038, opened 2019-07-22
Wrap undefined with str() call to produce the correct lineno in the traceback

### LINKED ISSUE #276: Wrong line number for UndefinedError exceptions
When using `jinja2.StrictUndefined` and rendering a template, the line numbers seem to be off when an `UndefinedError` exception is triggered:

``` python
Python 2.7.5 (default, Sep  6 2013, 09:55:21)
[GCC 4.8.1 20130725 (prerelease)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import jinja2
>>> template = '''Hello world!
... {{ missing }}
... {{ exists }}
... '''
>>> jinja_env = jinja2.Environment(undefined=jinja2.StrictUndefined)
>>> print jinja_env.from_string(template).render(exists='foo')
jinja2.exceptions.UndefinedError: 'missing' is undefined
```

The `missing` variable is on line 2, not line 3. However, when the variables are properly defined, the value for the `missing` variable is on line 2 like you would expect.

``` python
>>> print jinja_env.from_string(template).render(exists='foo', missing='bar')
Hello world!
bar
foo
>>>
```

The same line number mismatch is seen when loading the template from a file, as well:

``` python
Python 2.7.5 (default, Sep  6 2013, 09:55:21)
[GCC 4.8.1 20130725 (prerelease)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import jinja2
>>> jinja_env = jinja2.Environment(loader=jinja2.FileSystemLoader('/home/erik'), undefined=jinja2.StrictUndefined)
>>> print jinja_env.get_template('template.txt').render(exists='foo')
  File "/home/erik/template.txt", line 3, in top-level template code
    {{ exists }}
jinja2.exceptions.UndefinedError: 'missing' is undefined
```

So, `missing` is undefined, yet the traceback shows the `exists` line.

The above testing was done using version 2.7.1.

kinow:
Hi,

This issue is affecting us (https://github.com/cylc/cylc-flow/issues/2572) and we are getting the wrong line reported in the traceback when loading our templates.

I wrote a small program to confirm what was going on:

```python
class Z(object):
    def __str__(self):
        raise ValueError('Z error')


def raise_generator():
    yield 'three values are: %s %s %s' % (
        'primeiro',
        Z(),  # traceback must point to this lineno 9
        'terceiro'  # but points to this lineno 10 (__str__ only, __eq__ is OK)
    )


print(list(raise_generator()))
```

The traceback printed is

```
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/tests/Z.py", line 14, in <module>
    print(list(raise_generator()))
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/tests/Z.py", line 10, in raise_generator
    'terceiro'  # but points to this lineno 10 (__str__ only, __eq__ is OK)
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/tests/Z.py", line 3, in __str__
    raise ValueError('Z error')
ValueError: Z error
```

The issue appears to be how Python is tracking the line numbers when you have a string with values appended. If you call `undefined(...).__str()__` or `str(undefined(...))`, instead of simply calling `"... %s " % (undefined(...))` then the traceback uses the correct line number.

If you change the code above to use call `__str__()` or `str(undefined(...))` as in this PR, then the right line is reported in the traceback.

Reported a [bug for Python devs](https://bugs.python.org/issue37647) to confirm if that behaviour was intentional, as I would expect it to pick the correct line number.

But in the meantime, this small change looks harmless, and fixes this %+ year old - and very annoying for us at least - issue.

And thanks for Jinja2 :tada: 

This closes #276 

Cheers
Bruno

### DIFF
--- a/jinja2/compiler.py
+++ b/jinja2/compiler.py
@@ -1409,7 +1409,7 @@ class CodeGenerator(NodeVisitor):
             load = frame.symbols.find_load(ref)
             if not (load is not None and load[0] == VAR_LOAD_PARAMETER and \
                     not self.parameter_is_undeclared(ref)):
-                self.write('(undefined(name=%r) if %s is missing else %s)' %
+                self.write('(str(undefined(name=%r)) if %s is missing else %s)' %
                            (node.name, ref, ref))
                 return
 
--- a/tests/res/templates/undefined.html
+++ b/tests/res/templates/undefined.html
@@ -1,0 +1,11 @@
+This is a text with
+one
+{{ two }} (error happens here first!)
+three
+{{ four }} (error here as well)
+five
+{{ six }} (no error)
+and more lines.
+
+The line using the "two" variable must be the one to appear in the
+traceback with a matching 'two' is missing message.
--- a/tests/test_debug.py
+++ b/tests/test_debug.py
@@ -14,7 +14,8 @@ import re
 import sys
 from traceback import format_exception
 
-from jinja2 import Environment, TemplateSyntaxError
+from jinja2 import Environment, TemplateSyntaxError, StrictUndefined
+from jinja2.debug import make_traceback
 
 
 @pytest.fixture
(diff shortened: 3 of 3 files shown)


kinow:
(found this solution when checking the output of the original code with `dis.dis(...)` and comparing with the output of calling `.__str__()` and `str(...)`, and noticing an extra `LOAD_METHOD` + `CALL_METHOD`, that I suspect cause the traceback to locate the correct line numbers).

kinow:
Build against `master` appears to be failing even without this change (from looking at the link in the project `README`). Should I rebase it onto `maintenance-2.9`?

kinow:
ps: the bug on CPython has been closed, as in 3.8+ they changed the behaviour. Instead of returning the last line in that statement, it is now returning the start of the expression. So it is now returning the first line in the `yield` statement of the generated code.

Which means we would still have the bug without the change in this PR, I believe.

kinow:
Actually the error is legit. In some cases the `Undefined` must not be evaluated, or at least not simply wrapped in `str()` (some tests failing due to `int` + `str`). Bigger challenge, as this was the only hack I knew to get Python to output the correct line numbers...

davidism:
I'm not clear what the status of this is. Are you saying that you no longer think this should be merged? I'm not convinced this is something we need to work around, as this is just how traceback reporting works in Python.

kinow:
@davidism 

>I'm not clear what the status of this is. Are you saying that you no longer think this should be merged?

Yes, unfortunately. The workaround I found would force the `undefined` to be evaluated immediately, always raising an exception (which is not always desired).

> I'm not convinced this is something we need to work around, as this is just how traceback reporting works in Python.

The Python devs changed from using the first line number in a multiline expression, to using now the last number, and don't seem likely to adjust the compiler/parser/etc to match the line where the exception actually happened.

That's why I was trying to find a workaround for this issue, as this impact our Cylc workflow when there are errors in with Jinja templates. But I think we will have to just tell users to verify if the number really matches their Jinja template code.

Thanks
Bruno

PR #1040, opened 2019-07-24
Simplify splitting import_name in import_string()

minho42:


jinja2/utils.py
```diff
@@ -136,9 +136,7 @@ def import_string(import_name, silent=False):
         if ':' in import_name:
             module, obj = import_name.split(':', 1)
         elif '.' in import_name:
-            items = import_name.split('.')
-            module = '.'.join(items[:-1])
-            obj = items[-1]
+            module, _, obj = import_name.rpartition('.')
```
davidism: `module, obj = import_name.rsplit('.', 1)`

minho42: I used rpartition because it appears to be slightly faster than rsplit. 
But rsplit looks cleaner.

minho42: @davidism 
Would you like me to change it to rsplit?

PR #1045, opened 2019-08-09
Update PyPI link to point to Jinja2 instead of 1

yuriks:


jab:
Thank you for the contribution!

PR #1053, opened 2019-08-27
Allow truncating from the left side

### LINKED ISSUE #1052: truncate filter can no longer truncate on the right side
Before commit #610 it was possible to do right side truncation, e.g. if one wanted to split a month name into abbreviated (3 char long) and non-abbreviated form it would be possible to achieve with:
```jinja
"November"|truncate(3, True, '')
"November"|reverse|truncate(-3, True, '')|reverse
```

The above would produce "Nov" and "ember" due to the way how string functions work:
```
>>> s='123456789'
>>> s[:3]
'123'
>>> s[:-3]
'123456'
>>>
```

I think that commit removed a very useful functionality that is no longer achievable with the current version of Jinja (due to an explicit `assert()` that requires a positive truncation length to be provided.  I would suggest to remove that assert to give the opportunity to truncate a string on any end using the truncate filter.

galaxy4public:
Now, it is again possible to truncate strings from both left and
right sides by specifying positive and negative values for the
length parameter.

Fixes #1052.

### DIFF
--- a/jinja2/filters.py
+++ b/jinja2/filters.py
@@ -649,7 +649,9 @@ def do_indent(
 @environmentfilter
 def do_truncate(env, s, length=255, killwords=False, end='...', leeway=None):
     """Return a truncated copy of the string. The length is specified
-    with the first parameter which defaults to ``255``. If the second
+    with the first parameter which defaults to ``255``. The negative length
+    will truncate the string on the right side, i.e. chop the specified
+    number of characters from the end of the string. If the second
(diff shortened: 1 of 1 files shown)


kevin-brown:
Is it possible to get a test case for this, so we don't run into this regression again in the future?

galaxy4public:
Fair enough, will see what I can do today.

davidism:
Closing as I think the current behavior is appropriate. https://github.com/pallets/jinja/issues/1052#issuecomment-549957567

PR #1055, opened 2019-08-29
Added new `custom_concat` function to fix #1020 issue

### LINKED ISSUE #1020: Jinja2 NativeEnvironment dropping single quotes between elements in template to render
## Expected Behavior
```
>>> from jinja2 import Template
>>> template = Template("'{{ foo }}', '{{ bar }}', '{{ baz }}'")
>>> template.render(foo='foobar', bar='barbaz', baz='bazfoo')
('foobar', 'barbaz', 'bazfoo')
```

## Actual Behavior
```
>>> from jinja2.nativetypes import NativeEnvironment
>>> e = NativeEnvironment()
>>> t = e.from_string("'{{ foo }}', '{{ bar }}', '{{ baz }}'")
>>> t.render(foo='foobar', bar='barbaz', baz='bazfoo')
'foobar, barbaz, bazfoo'
```

## Your Environment
* Python version: 2.7.5
* Jinja version: 2.10, 2.10.1, current master

Trying to find an explanation of the on the above. Tested this in an interpreter to narrow it down to just Jinja2, and nothing else that may be calling it, such as Ansible with jinja2_native set to True.

When not using jinja2_native, we get the expected behavior above that the three substitutions occur with single quotes around each value. When jinja2_native is set to true (in the NativeEnvironment as shown above), the inner single quotes in the string are disregarded, and a single string is returned with the three substitutions.

This also occurs even with additional characters between the single quotes:
```
>>> e = NativeEnvironment()
>>> t = e.from_string("--foo='{{ foo }}' --bar='{{ bar }}'")
>>> t.render(foo='foobar', bar='barbaz')
u"--foo='foobar --bar=barbaz'"
```

Any help would be appreciated.

ditekunov:
About 40% of our project's ansible roles were broken because of the #1020 issue. This is a possible evaluation for this problem.

### DIFF
--- a/jinja2/nativetypes.py
+++ b/jinja2/nativetypes.py
@@ -6,7 +6,7 @@ from jinja2 import nodes
 from jinja2._compat import text_type
 from jinja2.compiler import CodeGenerator, has_safe_repr
 from jinja2.environment import Environment, Template
-from jinja2.utils import concat, escape
+from jinja2.utils import escape, custom_concat
 
 
 def native_concat(nodes):
@@ -169,7 +169,7 @@ class NativeCodeGenerator(CodeGenerator):
                     arguments.append(item)
 
             self.writeline('yield ')
-            self.write(repr(concat(format)) + ' % (')
+            self.write(repr(custom_contact(format)) + ' % (')
             self.indent()
 
             for argument in arguments:
--- a/jinja2/utils.py
+++ b/jinja2/utils.py
@@ -39,6 +39,15 @@ internal_code = set()
 
 concat = u''.join
 
+def custom_concat(to_concat):
+    def format_args(arg):
+        if "'" in arg:
+            return arg.replace("'", "")
+        else:
+            return "'" + arg + "'"
+
+    return '\"'.join(list(map(lambda arg: format_args(arg), to_concat)))
+
 _slash_escape = '\\/' not in json.dumps('/')
 
 


jinja2/utils.py
```diff
@@ -39,6 +39,15 @@
...
 concat = u''.join
 
+def custom_concat(to_concat):
+    def format_args(arg):
+        if "'" in arg:
+            return arg.replace("'", "")
+        else:
+            return "'" + arg + "'"
+
+    return '\"'.join(list(map(lambda arg: format_args(arg), to_concat)))
```
ThiefMaster: lambda and list are not needed here:

    return '\"'.join(map(format_args, to_concat))

jinja2/utils.py
```diff
@@ -39,6 +39,15 @@
 
 concat = u''.join
 
+def custom_concat(to_concat):
+    def format_args(arg):
+        if "'" in arg:
```
ThiefMaster: this doesn't break anything?! it looks like the kind of code that would easily break things

ditekunov: This is the only way i have found to  parse 'format' list in visit_Output function

jinja2/nativetypes.py
```diff
@@ -169,7 +169,7 @@ def const_finalize(x):
                     arguments.append(item)
 
             self.writeline('yield ')
-            self.write(repr(concat(format)) + ' % (')
+            self.write(repr(custom_contact(format)) + ' % (')
```
ThiefMaster: s/contact/concat/

PR #1057, opened 2019-09-02
Workaround for #1020 issue

ditekunov:
Issue #1020 breaks about 40% of our ansible roles, this can be a solution for this problem


davidism:
Thanks for working on this. My intuition is that we should not be doing this type of string manipulation to patch this. Instead we need to identify and address the real issue causing this.

PR #1059, opened 2019-09-03
Fix simple typo: problemtic -> problematic

timgates42:


jinja2/runtime.py
```diff
@@ -500,7 +500,7 @@ def __call__(self, *args, **kwargs):
         # decide largely based on compile-time information if a macro is
         # safe or unsafe.  While there was a volatile mode it was largely
         # unused for deciding on escaping.  This turns out to be
-        # problemtic for macros because if a macro is safe or not not so
+        # problematic for macros because if a macro is safe or not not so
```
jab:
```suggestion
        # problematic for macros because whether a macro is safe depends not
        # on the escape mode when it was defined, but rather when it was used.
```

jinja2/runtime.py
```diff
@@ -500,7 +500,7 @@ def __call__(self, *args, **kwargs):
...
         # much depends on the escape mode when it was defined but when it
```
jab:
```suggestion
```

jinja2/runtime.py
```diff
@@ -500,7 +500,7 @@ def __call__(self, *args, **kwargs):
...
         # was used.
```
jab:
```suggestion
```

timgates42:
sure thing - will do - Thanks for all these 👍

timgates42:
Thanks @jab - have accepted all your great suggestions.

PR #1064, opened 2019-09-22
Support paths created with pathlib in FileSystemLoader and ModuleLoader

### LINKED ISSUE #870: jinja2 incompatibility with paths created using pathlib
As part of refactoring, I have been converting my application from using `os.path` to `pathlib`.

However, it appears that `jinja2` is incompatible with paths created using `pathlib.Path`.

For example, if `tdir` is a `pathlib` Path object, the following generates an error:

`jinja_environment = Environment(loader=FileSystemLoader(tdir), trim_blocks=False, lstrip_blocks=True)`,

where the bottom of the error traceback is 

```
    jinja_environment = Environment(loader=FileSystemLoader(tdir),
  File ".../python/anaconda3/lib/python3.6/site-packages/jinja2/loaders.py", line 163, in __init__
    self.searchpath = list(searchpath)
TypeError: 'PosixPath' object is not iterable
```

If I convert `tdir` to a string, things work fine.

jinja: v 2.10
python: v3.6.5

alexwlchan:
This patch adds support for passing paths created with `pathlib.Path` in FileSystemLoader and ModuleLoader, both as a single path or one of a list, i.e.,

```python
FileSystemLoader(pathlib.Path('/templates'))
FileSystemLoader([pathlib.Path('/templates'), '/more/templates'])

ModuleLoader(pathlib.Path('/templates'))
ModuleLoader([pathlib.Path('/templates'), '/more/templates'])
```

There's no change in Python 2.7 (which doesn't support pathlib).

I've included tests for all four scenarios, along with a couple of other tests to improve the coverage of FileSystemLoader.

I believe this fixes #870.


alexwlchan:
Well that failed spectacularly quickly. 🤦‍♀ 

I swear I ran the tests before opening this patch, but only on my Mac. Investigating…

alexwlchan:
(Tests are now passing, I forgot I had to support Python 2.7… 😢 )

davidism:
Great PR, thanks!

davidism:
Ah, wait a second, @ThiefMaster's [other review](https://github.com/pallets/jinja/pull/876#issuecomment-402554835) applies to this:

> Just thinking.. would it make more sense to check for:
> 
> - is a string type (like before)
> - has a [PEP519-style `__fspath__` method](https://www.python.org/dev/peps/pep-0519/)
> - otherwise assume iterable, and use fspath if available or the plain value
> 
> Like this it would not only support pathlib paths / objects whose `__str__` is a path, but any object that represents a filesystem path.
> 
> Since pathlib support is a new feature in general I don't think the fact that `__fspath__` is Python3.6-only (probably one could use a pathlib backport for any older Python version) is a problem.

davidism:
pallets/flask#2059 might be helpful

PR #1065, opened 2019-09-22
Add docs, code and issue tracker links to setup.py

alexwlchan:
This will cause the links to show up in the PyPI sidebar, a la click: https://pypi.org/project/click/


jab:
Thank you for the contribution!

PR #1071, opened 2019-10-06
Eliminate dead code by using pytests.raises instead of try/except/else

btharper:
This PR just includes the simplest cases to make review simple as this is my first PR to the jinja project.


davidism:
Thanks, cleaning up the tests and using pytest features is really helpful, keep going if you want to. 😀

btharper:
Getting everything merged that quick was a welcome surprise. Just wanted to make sure the changes were something that made sense for the repo. I'll definitely work on a few of the similar cases when I get the chance then.

Thanks for quick turnaround and the warm welcome 😀

PR #1073, opened 2019-10-09
changed test_join functions to unique names

brtomas0:
the two tests were not being run due to having the same name, so changed the names to better reflect the function inputs.


davidism:
Nice catch, thanks!

PR #1074, opened 2019-10-10
refactored raise snippets into pytest.raise functions

brtomas0:
Refactoring code from try-except blocks to pytest.raise functions


davidism:
Just curious, since you're the second person to open such a PR (#1071) in this repo, and you also fixed exactly two instances of this. Is this related to the Hacktober event? Is there a reason you only addressed two instances of this instead of a more thorough refactor?

I don't mind merging this, it's nice that these are getting addressed, but it would be great to spend some more time thoroughly refactoring. A quick search of the tests folder for `except .*?:` shows there's not too many more.

PR #1075, opened 2019-10-10
Update "test_number" to not return true on bool value

nicbuitr:
Hi!

When evaluating: ```isinstance(True, integer_types + (float, complex, decimal.Decimal))``` returns `True` as `bool` is a subclass of `int`, so `True` is an instance of `int`.

Therefore my suggestion is to use the following instead: ```return type(value) in integer_types + (float, complex, decimal.Decimal))``` which does return `False` when tested with a `bool` value but `True` if it belongs to the number types.

I hope this helps as I use Jinja2 on a daily basis,

Best regards,

Nico

### DIFF
--- a/jinja2/tests.py
+++ b/jinja2/tests.py
@@ -130,7 +130,7 @@ def test_mapping(value):
 
 def test_number(value):
     """Return true if the variable is a number."""
-    return type(value) in integer_types + (float, complex, decimal.Decimal))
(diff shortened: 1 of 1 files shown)


nicbuitr:
Updated the test assertion at this pull request (https://github.com/pallets/jinja/pull/1076) to match this change

Best regards,

Nico

PR #1076, opened 2019-10-10
Update assertion of "is number" to return False on bool value

nicbuitr:
Hello!

This would be to update the assertion to compliment my other pull request (https://github.com/pallets/jinja/pull/1075) as `bool` values evaluated in ```{% <bool> is number %}``` would now return `False` as in ```{% True is number %}``` would be `False`

Best regards,

Nico


davidism:
Please include tests in the same PR as the fix.

Issue #79, opened 2011-12-06
Syntax error in generated python with double {% extends %}


SimonSapin:
Using more than one `{% extends %}` is not allowed, but the error message and location could be more explicit:

```
>>> jinja2.Template('{% extends "a" %}{% extends "b" %}')
  File "<template>", line 13
SyntaxError: 'yield' outside function (<template>, line 13)
```

PR #1077, opened 2019-10-10
Pytest cleanup

kevin-brown:
- Replace `try...catch` statements with `pytest.raises`
  * This was not done for the one in `test_debug` because it relies on the global exception handler to parse out the traceback

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -336,20 +336,12 @@ class TestUndefined(object):
         pytest.raises(UndefinedError, t.render, var=0)
 
     def test_none_gives_proper_error(self):
-        try:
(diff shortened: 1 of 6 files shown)


ThiefMaster:
I was just about to complain about a camelcased `assertRaises` but luckily you meant `pytest.raises` ;)

davidism:
Huh, I think `test_double_extends` is outright wrong, it doesn't raise any error, and the previous version was masking that. A demonstration of why this type of refactor is good. :-)

davidism:
Related issue #79 seems to be about extending *different* templates. This test, added in #123, seems intentionally designed to only fail before the fix for #79 introduced in #83. The solution here seems to be to expect the call to succeed without error.

PR #1079, opened 2019-10-12
Use separate undefined object for else-less conditional expressions

### LINKED ISSUE #710: `{{ 'foo' if false }}` shortcut not usable with StrictUndefined
Using the shortcut version of the ternary operator with no `else` implies `else <undefined>`. This is fine when using the normal `Undefined` but makes this shortcut basically useless with `StrictUndefined` (and to some extent `LoggingUndefined`).

```python
>>> from jinja2 import StrictUndefined, Environment
>>> env = Environment(undefined=StrictUndefined)
>>> env.from_string('{{ "foo" if false }}').render()
  File "/home/adrian/.python-env/lib/python2.7/site-packages/jinja2/environment.py", line 989, in render
  File "/home/adrian/.python-env/lib/python2.7/site-packages/jinja2/environment.py", line 754, in handle_exception
  File "<template>", line 1, in top-level template code
jinja2.exceptions.UndefinedError: the inline if-expression on line 1 evaluated to false and no else section was defined.
```

I'd change the default from `<undefined>` to `''` - like this the behavior of existing templates will not change (empty output) but it would work fine with the stricter `Undefined` variants.

The only breakage I could imagine is someone using it in a function/macro call with an explicit `is defined` check later on. But I think that's rather unlikely and if necessary a new policy option could be added to make it default to `<undefined>` again.

ThiefMaster:
fixes #710

### DIFF
--- a/jinja2/environment.py
+++ b/jinja2/environment.py
@@ -101,9 +101,6 @@ def _environment_sanity_check(environment):
     """Perform a sanity check on the environment."""
     assert issubclass(environment.undefined, Undefined), 'undefined must ' \
         'be a subclass of undefined because filters depend on it.'
-    assert issubclass(environment.cond_expr_undefined, Undefined), \
-        'cond_expr_undefined must be a subclass of undefined because filters ' \
-        'depend on it.'
     assert environment.block_start_string != \
         environment.variable_start_string != \
         environment.comment_start_string, 'block, variable and comment ' \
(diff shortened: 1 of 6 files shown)


jinja2/environment.py
```diff
@@ -283,7 +293,8 @@ def __init__(self,
                  cache_size=400,
                  auto_reload=True,
                  bytecode_cache=None,
-                 enable_async=False):
+                 enable_async=False,
+                 cond_expr_undefined=Undefined):
```
ThiefMaster: For more backwards compatibility this could be set to `None` here and use the value of `undefined` in that case.

davidism:
Do we need this to be configurable? It seems like the contract of `value if condition` is that it returns `Undefined`, maybe it can just be that simple.

ThiefMaster:
That was my question ;) Then I'll change the default to use the `undefined` value if nothing is specified.

ThiefMaster:
updated

ThiefMaster:
oh, I think I misread your comment, i thought you mean that it's supposed to return the `undefined` object by default (which is whatever is configured in the env as `undefined`)

PR #976, opened 2019-05-02
add ImportLibResourceLoader

gsemet:
closes #970

PR #1082, opened 2019-10-16
PackageLoader doesn't depend on setuptools

### LINKED ISSUE #970: Replace pkg_resource by importlib_resources
Hi.

Can you provide a future-proof loader that use `importlib_resources` instead of `pkg_resource`, which does not work pretty well in every case (on my pyinstaller package for instance it fails).

For me the use of `pkg_resource` should be avoided, it is about to be discarded (or maybe one day setuptools will not embed it). Providing a loader compatible with `importlib_resources` (the backport that brings' 3.7 `importlib.resources` to Python <3.7) is the easiest and more future-proof solution.

Thanks

davidism:
closes #970 
closes #976 

Based on pallets/werkzeug#1647. Uses `pkgutil.get_loader` instead of `pkg_resources`. `pkg_resources` required an implicit dependency on setuptools, and was is slow to import.

`importlib.resources` / `importlib_resources` makes some unhelpful decisions, such as requiring every directory to have a `__init__.py` file, and only loading one level deep. Python 3's `loader.get_resource_reader()` API suffers the same issue for the most part. Python's resource API in general is still too limited for us to support arbitrary types of installations, so only directories and zip files are supported, and it's mostly done manually rather than with the loader/reader API.


davidism:
PyPy's `zipimporter` doesn't have the `_files` attribute. :disappointed: I think I just won't return the template list in that case, rather than trying to traverse the zip file manually.

Implementation on Windows have some path issues, I'll have to address those once I can boot into Windows.

gsemet:
I am not sure this method works with pyinstaller's way of packaging jinja, while mine does :)

davidism:
`importlib.resources` is a non-starter as long as it requires `__init__.py` in every directory and doesn't support directory traversal.

davidism:
I'm assuming you mean pyinstaller's way of packaging package data, not packaging Jinja, since how Jinja is packaged does not affect how other things are loaded. I'm not familiar with PyInstaller, if you have a simple example of a package with a templates directory, I can take a look.

gsemet:
that's not something very problematic, you know, it is just a file to add. And I recommend letting people decide which method is the best for them. If you need to package jinja in pyinstaller like i do, I am pretty happy with this __init__.py (it is actually not the more weird thing to do). At least I have the option to choose this loader if I want to.
But it you find a better way to have jinja happy in pyinstaller or similar packaging, I'll be happy to switch

gsemet:
> I'm assuming you mean pyinstaller's way of packaging package data, not packaging Jinja, since how Jinja is packaged does not affect how other things are loaded.

Yes, packaging jinja and a template folder as data inside a pyinstaller executable, so that this executable can render the template. I use importlib resource and yes, it's tricky to do, but at least it works

davidism:
If you need a specific loader for your project, you can always add a loader, we don't need to cover every possible case in the built-in loaders, that's the whole point of it being an interface.

davidism:
Just checked, `PackageLoader` with `pkg_resources` doesn't seem to support PyInstaller either because PyInstaller uses a custom `FrozenImporter`.

This new implementation does support PyInstaller.

```
$ tree example/
example
├── __init__.py
├── __pycache__
│   └── __init__.cpython-37.pyc
└── templates
    └── test.txt

2 directories, 3 files
$ pyinstaller -F --name example --add-data example/templates:example/templates entry.py
$ ./dist/example
Hello

-   1
-   2
-   3

World
```

gsemet:
Great, so it's even simpler that mine :) Will check on it once released !

a-wakeel:
will this be part of the upcoming release?

davidism:
Reverting this in 2.11.2, it broke compatibility with Pytest. However, support will be present when 3.0 is released. See #1168, #1169

PR #1089, opened 2019-10-24
adds strict argument to FileSystemLoader

cav71:
This pacth introduce the strict parameter to the FileSystemLoader.

Eg.
```python
>>> import jinja2 as jj, os.path

>>> relpath = 'templates/test.html'
>>> fullpath = os.path.abspath('tests/res/templates/test.html')

>>> fsl = jj.FileSystemLoader(['tests/res', ])
>>> fsl_nostrict = jj.FileSystemLoader(['tests/res', ], strict=False)

>>> jj.Environment(loader=fsl).get_template(relpath).render()
'BAR'

>>> jj.Environment(loader=fsl).get_template(fullpath).render()
raises TemplateNotFound

>>> jj.Environment(loader=fsl_nostrict).get_template(fullpath).render()
'BAR'

### DIFF
--- a/jinja2/loaders.py
+++ b/jinja2/loaders.py
@@ -180,7 +180,7 @@ class FileSystemLoader(BaseLoader):
         searchpaths = self.searchpath
         if not self.strict and path.isabs(template):
             searchpaths = [ template, ]
-            pieces = (,)
+            pieces = tuple()
         for searchpath in searchpaths:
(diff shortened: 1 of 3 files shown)


davidism:
Thanks for looking at this. However, template names are not filesystem paths, so I don't think this should be added.

PR #1099, opened 2019-11-04
Added Python 3.8 support

RDIL:
Before now, this would throw an error because `collections.abc` was removed, replaced by `collections.abc.abc`.  This should fix it.  

See also: https://github.com/GoogleCloudPlatform/flask-talisman/issues/47

### DIFF
--- a/jinja2/_compat.py
+++ b/jinja2/_compat.py
@@ -13,6 +13,7 @@
 import sys
 
 PY2 = sys.version_info[0] == 2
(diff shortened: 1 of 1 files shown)


davidism:
I don't understand this change. The `try/except` already provides this compat, no need for the if statement and duplication.

davidism:
The deprecations were addressed in #867, released in 2.10.2 on 2019-10-04.

Issue #1102, opened 2019-11-12
Bad comparison parsing for constants


areebbeigh:
Comparisons in jinja don't behave like comparisons in Python.

```python
template = '''{{4 < 2 < 3}}
{{a < b < c}}'''
t = Template(template)
print(t.render(a = 4, b = 2, c = 3))
```
### Expected output: 
```python
False
False
```
### Output: 
```python
True
False
```

Python code:
```python
4 < 2 < 3
```
Output: `False`

The problem lies in https://github.com/pallets/jinja/blob/1af1205a9c9e884c49f74eb3e75dabbcee1b1967/jinja2/nodes.py#L775-L780
where the result of the comparison depends only on the final operator result. 

I would like to submit a PR for this.

* Python version: 3.6.7
* Jinja version: 2.10

davidism:
PRs welcome

PR #1105, opened 2019-11-15
nodes.py: Return on first false in Compare() expr

areebbeigh:
Closes: https://github.com/pallets/jinja/issues/1102


davidism:
Thanks! Please add a test and changelog too.

Issue #408, opened 2015-02-08
Template source code map in jinja2


alecxe:
I'm trying to measure jinja2 template code coverage following Ned Batchelder's [django_coverage_plugin](https://github.com/nedbat/django_coverage_plugin) and [coverage-mako-plugin](https://bitbucket.org/ned/coverage-mako-plugin) coverage plugins.

Since `jinja2` templates are compiled to Python files the approach should be similar to how mako template coverage is currently measured. Though, in mako there is a handy [`ModuleInfo` class](https://bitbucket.org/zzzeek/mako/src/b6d0479034e4f379576caa6eb9ef3e32369c333f/mako/template.py?at=master#cl-575) that provides a template source code map that helps to measure the template code coverage.

Is there anything similar in `jinja2` or should I use the meta API?
Would appreciate any pointers.

Stackoverflow post as a reference: [Jinja2 template compilation information](http://stackoverflow.com/questions/28328283).

citizen-stig:
I've tried to use get_corresponding_lineno, but it seems that it doesn't work
http://stackoverflow.com/questions/37032085/jinja2-template-get-corresponding-lineno-doesnt-work

dongyuwei:
when I Google `jinja template source map`, I got this issue link :) 
But my requirement is different: I just wan to know which partial templates rendered the html segments in browser(one to one mapping).  So I ended up hacking the  `jinja2/loaders.py` source code like this: 
https://gist.github.com/dongyuwei/7f80f0589fbafa0706d741370207796b#file-loaders-py-L178

It works as I expected.

davidism:
Due to some removed behavior in #1109, more lines now have coverage since they aren't compiled into a mult-line format string expression. At this point I think the main issue is that the lexer and parser treat template data across line breaks as a single node. If someone figures out how to record the start and *end* line for each node, that will go a long way towards getting the information we need for full coverage.

PR #1109, opened 2019-11-20
refactor visit_Output

### LINKED ISSUE #276: Wrong line number for UndefinedError exceptions

davidism:
This originally started after I realized that since `NativeCodeGenerator.visist_Output()` is mostly a copy of the code from `compiler.py`, the changes made in #1080 weren't applied to it. To mitigate that in the future, I extracted the differences between them into some internal methods, so `NativeCodeGenerator` only needs to override those instead of the entire `visit_Output` method.

This made me realize that we were doing the same processing to generate the `finalize()` function for every output node, even though it's the same for all of them. Now the compile time function and runtime source are generated once and cached.

Finally, `visit_Output` had two code paths for rendering the source: if there were < 3 nodes, each node would be yielded directly, otherwise a format string was generated to collect all the nodes and yield one big value. The two code paths were almost identical, which made keeping them in sync difficult. I experimented with removing the format string path, and it made the benchmarks in the examples folder about 30% faster. (Admittedly they were already very fast, but the results were consistent across machines).

This also fixes #276, since there are no multi-line format strings anymore. It also helps #408, although since the lexer still combines template data across line breaks some lines still don't have coverage mapping.

### DIFF
--- a/jinja2/compiler.py
+++ b/jinja2/compiler.py
@@ -8,8 +8,8 @@
     :copyright: (c) 2017 by the Jinja Team.
     :license: BSD, see LICENSE for more details.
 """
+from collections import namedtuple
 from itertools import chain
-from copy import deepcopy
 from keyword import iskeyword as is_python_keyword
 from functools import update_wrapper
 from jinja2 import nodes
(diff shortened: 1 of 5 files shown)


kinow:
Thanks @davidism ! Just to confirm, I tested the code from `master`, and instead of having the incorrect line number, it now printed the right line in the context, but without the line number. So is it correct to assume this will give just the line where the error occurred, without the line number now?

PR #1117, opened 2019-12-13
Fix TemplateSyntaxError pickling

nvllsvm:
Implement [`__reduce__`](https://docs.python.org/2/library/pickle.html#object.__reduce__) to fix pickling of TemplateSyntaxError

New test output without the fix (Python 3.8):
```
________________________________________ TestDebug.test_pickleable_syntax_error _________________________________________

self = <test_debug.TestDebug object at 0x109740490>, fs_env = <jinja2.environment.Environment object at 0x1097400a0>

    def test_pickleable_syntax_error(self, fs_env):
        exc = TemplateSyntaxError('wtf', 42)
        pickled_exc = pickle.dumps(exc)
>       unpickled_exc = pickle.loads(pickled_exc)
E       TypeError: __init__() missing 1 required positional argument: 'lineno'

test_debug.py:78: TypeError
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -99,6 +99,8 @@ Unreleased
     ``UndefinedError`` consistently. ``select_template`` will show the
     undefined message in the list of attempts rather than the empty
     string. :issue:`1037`
+-   ``TemplateSyntaxError`` can be pickled. :pr:`1117`
+
 
 Version 2.10.3
 --------------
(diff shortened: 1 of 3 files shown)


davidism:
Interesting.Seems like it would be better to address why `lineno` isn't getting pickled/unpickled.

nvllsvm:
Some findings:
- https://bugs.python.org/issue32696
- https://bugs.python.org/issue1692335#msg310951

nvllsvm:
@davidism Changed the fix to only override `__reduce__`. No other changes to the exception.

Issue #1130, opened 2020-01-24
Template coverage discussion


Thomasb81:
Hello
I probably may not use the issue tracking as expected. Sorry about this.

In reference to:
https://github.com/pallets/jinja/pull/674#issuecomment-549373909
https://github.com/pallets/jinja/issues/408#issuecomment-556501327

Given following template, helloWorld.text:
```
1
2
3  {% block block1 %}
4     {% if poli %}
5  Hello World
6      {% else %}
7
8
9
10     {% endif %}
11  {% endblock %}
12
```

Producing such runtime-code where I report the requirement pointed by @davidism  :
- record the start and end line into node structure
- make it appear in the AST 

```from __future__ import division, generator_stop
from jinja.runtime import LoopContext, TemplateReference, Macro, Markup, TemplateRuntimeError, missing, concat, escape, markup_join, unicode_join, to_string, identity, TemplateNotFound, Namespace, Undefined
name = 'helloWorld.text'

def root(context, missing=missing):
    start_line = 1
    end_line = 12
    resolve = context.resolve_or_missing
    undefined = environment.undefined
    cond_expr_undefined = Undefined
    if 0: yield None
    pass
    yield '1\n2\n3  '
    yield from context.blocks['block1'][0](context)
    yield '\n12'

def block_block1(context, missing=missing):
    start_line = 3
    end_line = 11
    resolve = context.resolve_or_missing
    undefined = environment.undefined
    cond_expr_undefined = Undefined
    if 0: yield None
    l_0_poli = resolve('poli')
    pass
    yield '\n4     '
    if (undefined(name='poli') if l_0_poli is missing else l_0_poli):
        start_line = 4
        end_line = 6
        pass
        yield '\n5  Hello World\n6      '
    else:
        start_line = 6
        end_line = 10
        pass
        yield '\n7\n8\n9\n10     '
    yield '\n11  '

blocks = {'block1': block_block1}
debug_info = '3=14&4=27'
```
Does that match implementation criteria in order to progress on _template coverage support_ in jinja ?
Thanks
Thomas

davidism:
[...]

PR #1132, opened 2020-01-25
Record start_line  end_line into node build with token tagged "data"

Thomasb81:
Reference :
https://github.com/pallets/jinja/pull/674#issuecomment-549373909
https://github.com/pallets/jinja/issues/408#issuecomment-556501327
https://github.com/pallets/jinja/issues/1130


Thomasb81:
test suite is failing ...

PR #1149, opened 2020-02-07
Fix typo

davidbrochart:


davidism:
Thanks for pointing this out. I felt the wording was still confusing, so I just ended up rewriting the whole section.

PR #1152, opened 2020-02-10
Identify an async environment, and return the corresponding strings

angelafrentz:
Initial refactor for async statements in compiler.py cc @davidism


davidism:
Moved to a `Compiler.choose_async` method to avoid having to pass `self.environment` still, makes things more concise.

PR #1161, opened 2020-02-28
Add link to template designer documentation

ldealmei:
For those who were actually looking for the template design documentation putting a link directly to it might prevent them to have to google for it. When on the API page, there is no direct way to find the template design documentation. Hope this can help others.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5,9 +5,10 @@ API
     :noindex:
     :synopsis: public Jinja API
 
-This document describes the API to Jinja and not the template language.  It
-will be most useful as reference to those implementing the template interface
-to the application and not those who are creating Jinja templates.
+This document describes the API to Jinja and not the template language
+(for that, see :doc:`/templates`). It will be most useful as reference
+to those implementing the template interface to the application and not
+those who are creating Jinja templates.
 
 Basics
 ------


docs/api.rst
```diff
@@ -5,7 +5,7 @@ API
     :noindex:
     :synopsis: public Jinja API
 
-This document describes the API to Jinja and not the template language.  It
+This document describes the API to Jinja and not the template language (for that, see [here](https://jinja.palletsprojects.com/en/2.11.x/templates/).  It
```
ThiefMaster: The docs are rst, not markdown. I don't think your syntax works there.

ldealmei: Ah... Let me fix that

docs/api.rst
```diff
@@ -5,7 +5,7 @@ API
     :noindex:
     :synopsis: public Jinja API
 
-This document describes the API to Jinja and not the template language.  It
+This document describes the API to Jinja and not the template language (for that, see `here <https://jinja.palletsprojects.com/en/2.11.x/templates/>`_.  It
```
davidism: This needs to be a reference to the docs page, not a direct link to the hosted docs.

```rst
for that, see :docs:`/templates`
```

PR #1164, opened 2020-03-06
Don't use setlocal in syntax file

oeuftete:
I'm not 100% sure that one should not use `setlocal` in a syntax file, but this breaks other filetypes that may include this syntax file, like the vim-markdown plugin if `jinja` is included in its `g:markdown_fenced_languages`.  The `comments` and `commentstring` end up set in the markdown buffer with the jinja settings.

See https://github.com/dhruvasagar/vim-table-mode/issues/162#issuecomment-595317626 for an example of side effects.

### DIFF
--- a/ext/Vim/jinja.vim
+++ b/ext/Vim/jinja.vim
@@ -82,9 +82,6 @@ syn region jinjaRaw matchgroup=jinjaRawDelim start="{%\s*raw\s*%}" end="{%\s*end
 
 " Jinja comments
 syn region jinjaComment matchgroup=jinjaCommentDelim start="{#" end="#}" containedin=ALLBUT,jinjaTagBlock,jinjaVarBlock,jinjaString,jinjaComment
-" help support folding for some setups
-setlocal commentstring={#%s#}
-setlocal comments=s:{#,e:#}
 
 " Block start keywords.  A bit tricker.  We only highlight at the start of a
 " tag block and only if the name is not followed by a comma or equals sign


oeuftete:
I think the right place for this would be in a filetype plugin, as in https://github.com/Glench/Vim-Jinja2-Syntax/blob/6eff8287c4b61e9fb8657773f36402b33750fc27/ftplugin/jinja.vim#L10-L11

davidism:
I know nothing about vim or its syntax files. These lines seem to refer to entering comments in Jinja, how does removing these lines affect that?

oeuftete:
@davidism Not an expert on what vim's doing here, but I'll give it a try.

The gist of it is that this file defining syntax highlighting is exceeding its scope by setting the detected comment format in the syntax file.

The upshot is that when a vim buffer that pulls in multiple different syntax highlighting files, such as a markdown file with a Jinja code block in this case, evaluates _this_ syntax file, it sets the comment format for the whole buffer.  So the comment format for the entire markdown file ends up as Jinja's comment format.

davidism:
Ok, but if this is removed and we're no editing jinja in markdown, what happens to jinja comments?

davidism:
This behavior seems correct for a jinja template that happens to render a markdown document. Jinja comments will be processed out when rendering jinja before rendering markdown, they're the comments that should be preferred.

oeuftete:
> This behavior seems correct for a jinja template that happens to render a markdown document. Jinja comments will be processed out when rendering jinja before rendering markdown, they're the comments that should be preferred.

Yes, true.  It is not correct for a markdown document (that is not being rendered) that includes a Jinja code block.  That's my case.

> If this is removed and we're not editing jinja in markdown, just plain jinja, what happens to comments?

If you're just using this syntax file alone, I'm not sure.  As far as I understand it, this could/should be set elsewhere, per my earlier comment:

> I think the right place for this would be in a filetype plugin, as in https://github.com/Glench/Vim-Jinja2-Syntax/blob/6eff8287c4b61e9fb8657773f36402b33750fc27/ftplugin/jinja.vim#L10-L11

oeuftete:
cc: @tbnorth who originally added this in #990.

davidism:
Are there other Vim syntax files you can compare this to? How does, for example, a Python or HTML syntax file define comments that work in fenced blocks / multi-language files? Are we sure this is an issue with this file and not with Vim / another plugin that's providing fenced code support?

oeuftete:
> Are there other Vim syntax files you can compare this to? How does, for example, a Python or HTML syntax file define comments that work in fenced blocks / multi-language files? Are we sure this is an issue with this file and not with Vim / another plugin that's providing fenced code support?

This jinja one is the only one that has broken markdown highlighting for me, and I have around 20 languages configured to highlight in the fenced code blocks.

Here's where html and python in the vim repository set their comment strings, in the ftplugin file, and not in the syntax file:

* html: https://github.com/vim/vim/blob/9be61bbb170ed3df0e408c8ac12516e772dc0b75/runtime/ftplugin/html.vim#L16
* python: https://github.com/vim/vim/blob/9be61bbb170ed3df0e408c8ac12516e772dc0b75/runtime/ftplugin/python.vim#L38

Here's where `commentstring` is set in `ftplugin` or `syntax` files in [the vim repository](https://github.com/vim/vim) for languages included with vim.  Only in two syntax files, and I suspect those may not be intended.  In both cases, they are already set in `ftplugin` files.

```
ken@pancakes:~/working/git/github/vim/runtime (master=) $ ack -l '^set(local)? commentstring' | sort
ftplugin/abaqus.vim
ftplugin/ada.vim
ftplugin/awk.vim
ftplugin/bst.vim
ftplugin/cfg.vim
ftplugin/clojure.vim
ftplugin/cmake.vim
ftplugin/cobol.vim
ftplugin/csh.vim
ftplugin/c.vim
ftplugin/dockerfile.vim
ftplugin/dosbatch.vim
ftplugin/dtd.vim
ftplugin/dtrace.vim
ftplugin/eiffel.vim
ftplugin/eruby.vim
ftplugin/gdb.vim
ftplugin/go.vim
ftplugin/groovy.vim
ftplugin/hamster.vim
ftplugin/hog.vim
ftplugin/html.vim
ftplugin/javascript.vim
ftplugin/java.vim
ftplugin/jproperties.vim
ftplugin/json.vim
ftplugin/j.vim
ftplugin/liquid.vim
ftplugin/make.vim
ftplugin/matlab.vim
ftplugin/meson.vim
ftplugin/mma.vim
ftplugin/msmessages.vim
ftplugin/nroff.vim
ftplugin/ocaml.vim
ftplugin/pdf.vim
ftplugin/perl6.vim
ftplugin/perl.vim
ftplugin/php.vim
ftplugin/python.vim
ftplugin/registry.vim
ftplugin/rmd.vim
ftplugin/rrst.vim
ftplugin/ruby.vim
ftplugin/rust.vim
ftplugin/r.vim
ftplugin/sass.vim
ftplugin/scala.vim
ftplugin/sh.vim
ftplugin/tcl.vim
ftplugin/text.vim
ftplugin/tmux.vim
ftplugin/vim.vim
ftplugin/vroom.vim
ftplugin/wast.vim
ftplugin/xml.vim
syntax/dockerfile.vim
syntax/mma.vim
```

PR #1166, opened 2020-03-10
filters.py: do_indent: allow also tab indentions

LarsKollstedt:
Implemenented also indention with tabulator characters instead of
whitespace characters.

This is needed if this is used within a template with a specified tab indention. ;-)

It would be also thinkable to allow to indent with a generic character. This will follow soon, choose the best and close the other one.

Kind regards
     Lars


davidism:
closing, prefer #1167

PR #1167, opened 2020-03-10
indent filter can indent with arbitrary characters

LarsKollstedt:
Implemenented also indention with tabulator characters instead of
whitespace characters. This will also allow the use of other characters or character sequences. It's the more generic way I already promised in #1166, choose the best and close the other one.

This is needed if this is used within a template with a specified tab indention. ;-)

Kind regards
     Lars

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,6 +48,8 @@ Unreleased
     or ``@contextfunction``. :issue:`842`, :pr:`1248`
 -   Support ``pgettext`` and ``npgettext`` (message contexts) in i18n
     extension. :issue:`441`
+-   The ``|indent`` filter's ``width`` argument can be a string to
+    indent by. :pr:`1167`
 
 
 Version 2.11.3
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -748,20 +748,29 @@ def do_urlize(
     return rv
 
 
-def do_indent(s: str, width: int = 4, first: bool = False, blank: bool = False) -> str:
+def do_indent(
+    s: str, width: t.Union[int, str] = 4, first: bool = False, blank: bool = False
+) -> str:
     """Return a copy of the string with each line indented by 4 spaces. The
     first line and blank lines are not indented by default.
 
-    :param width: Number of spaces to indent by.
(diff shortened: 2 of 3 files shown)


davidism:
Let's do what [`json.dump`](https://docs.python.org/3/library/json.html#json.dump) does. If `width` is a string, it's used as the indent, otherwise it's the number of spaces to use as the indent. Makes it a bit weird to call it `width`, but I don't want to deal with a deprecation cycle just to rename it.

LarsKollstedt:
Implemented it as requested in https://github.com/pallets/jinja/pull/1167#issuecomment-612644701. The reuse of 'width' does'nt look like the finest solution to me, but it fit's my need.

matejak:
Why not to introduce a `indent_with` kwarg, to decouple the level of indentation from indentation type?

LarsKollstedt:
The original function has an attribute `width` that can only contain a length. My first suggestion for this was https://github.com/pallets/jinja/commit/08db4a329415c198bc8f4baab0752237f98c416a which was introducing a additional param called indent_char, which can of course also contain something like "&nbsp;". The `width` attribute was still conaining only a length there, and the indention type was decoupled.
I changed this due to https://github.com/pallets/jinja/pull/1167#issuecomment-612644701, but @davidism thought also of renaming the param to `indent`, but discarded this because of legacy reasons. Since this params are optional, theoretically you can have both a param `indent` that contains the whole indention, and a param `width` that contains the length with a param `indent_char`, that contains a single indention sequence. But I think that will cause people to discuss about deprecating some of them.
To rename the length parameter `width` to `indent_with` and to introduce a new `width` parameter with the indention type, will be one of the worst of possible all options.

matejak:
> To rename the length parameter `width` to `indent_with` and to introduce a new `width` parameter with the indention type, will be one of the worst of possible all options.

Sorry, I indeed wrote a nonsense. What I meant was to accompany `width` a kwarg `indent_string` or `indent_char` that would default to space. Arguments `indent`, `width` and `indent_char` are not orthogonal, so I also wouldn't support an interface that uses all three in one macro.

davidism:
> Let's do what json.dump does.

PR #1169, opened 2020-03-10
Use importlib machinery to fix PEP 451 import hooks

### LINKED ISSUE #1148: Regression: package was not installed in a way that PackageLoader understands.
## Expected Behavior
PackageLoader should continue to load from a namespace package even under pytest.

## Actual Behavior
Starting with jinja2 2.11, the tests for pmxbot have [started failing](https://travis-ci.org/yougov/pmxbot/jobs/647234617?utm_medium=notification&utm_source=github_status). They fail when pytest attempts to rewrite the modules for running under test and in doing so, cause the `jinja2.loaders.PackageLoader` not to recognize the 'pmxbot.web' package. Note that `pmxbot` is a setuptools namespace package.

## Full Traceback
```pytb
pmxbot/web/__main__.py:1: in <module>
    import pmxbot.web.viewer
<frozen importlib._bootstrap>:991: in _find_and_load
    ???
<frozen importlib._bootstrap>:975: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:671: in _load_unlocked
    ???
.tox/python/lib/python3.8/site-packages/_pytest/assertion/rewrite.py:143: in exec_module
    exec(co, module.__dict__)
pmxbot/web/viewer.py:24: in <module>
    jenv = jinja2.Environment(loader=jinja2.loaders.PackageLoader('pmxbot.web'))
.tox/python/lib/python3.8/site-packages/jinja2/loaders.py:288: in __init__
    raise ValueError(
E   ValueError: The 'pmxbot.web' package was not installed in a way that PackageLoader understands.
```

The error is not triggered when simply importing the package naturally; it only seems to happen during the pytest rewrite.

## Your Environment
* Python version: 3.8.1
* Jinja version: 2.11.x

Downgrading to Jinja 2.10.3 works around the failure. I suspect #970 is implicated.

### LINKED ISSUE #1168: jinja2 no longer supports the pytest loader
Currently, jinja2 expects either _path defined or get_filenames: https://github.com/pallets/jinja/blob/master/src/jinja2/loaders.py#L262-L281

The pytest assertion rewriter defines neither (see https://github.com/pytest-dev/pytest/blob/master/src/_pytest/assertion/rewrite.py#L48), and as such running a test suite on a source code that has the following global:

```python
from jinja2 import PackageLoader
LOADER = PackageLoader(__name__, "templates")
```

will fail with:

```
    raise ValueError(
E   ValueError: The 'xxx' package was not installed in a way that PackageLoader understands.
```

Not sure if here jinja2 needs to support more ways to get the template rooot, or pytest loader is missing some methods.

asottile:
Resolves #1148
Resolves #1168

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Unreleased
 -   Bump MarkupSafe dependency to >=1.1.
 -   Bump Babel optional dependency to >=2.1.
 -   Remove code that was marked deprecated.
+-   Use :pep:`451` API to load templates with
+    :class:`~loaders.PackageLoader`. :issue:`1168`
 
 
 2.11.2
--- a/src/jinja2/loaders.py
+++ b/src/jinja2/loaders.py
@@ -1,10 +1,11 @@
 """API and implementations for loading templates from different data
 sources.
 """
+import importlib.util
 import os
-import pkgutil
 import sys
 import weakref
+import zipimport
 from collections import abc
 from hashlib import sha1
 from importlib import import_module
@@ -253,21 +254,19 @@ class PackageLoader(BaseLoader):
         # Make sure the package exists. This also makes namespace
         # packages work, otherwise get_loader returns None.
         import_module(package_name)
-        self._loader = loader = pkgutil.get_loader(package_name)
+        spec = importlib.util.find_spec(package_name)
+        self._loader = spec.loader
 
-        # Zip loader's archive attribute points at the zip.
-        self._archive = getattr(loader, "archive", None)
+        self._archive = None
         self._template_root = None
-
-        if hasattr(loader, "get_filename"):
-            # A standard directory package, or a zip package.
-            self._template_root = os.path.join(
-                os.path.dirname(loader.get_filename(package_name)), package_path
-            )
-        elif hasattr(loader, "_path"):
-            # A namespace package, limited support. Find the first
-            # contributor with the template directory.
-            for root in loader._path:
+        if isinstance(spec.loader, zipimport.zipimporter):
+            self._archive = spec.loader.archive
+            pkgdir = next(iter(spec.submodule_search_locations))
+            self._template_root = os.path.join(pkgdir, package_path)
+        elif spec.submodule_search_locations:
+            # this will be one element for "packages" and multiple for
+            # namespace packages
+            for root in spec.submodule_search_locations:
                 root = os.path.join(root, package_path)
 
                 if os.path.isdir(root):
--- a/src/jinja2/loaders.py
+++ b/src/jinja2/loaders.py
@@ -255,17 +255,17 @@ class PackageLoader(BaseLoader):
         # packages work, otherwise get_loader returns None.
         import_module(package_name)
         spec = importlib.util.find_spec(package_name)
-        self._loader = spec.loader
-
+        self._loader = loader = spec.loader
         self._archive = None
         self._template_root = None
-        if isinstance(spec.loader, zipimport.zipimporter):
-            self._archive = spec.loader.archive
+
+        if isinstance(loader, zipimport.zipimporter):
+            self._archive = loader.archive
             pkgdir = next(iter(spec.submodule_search_locations))
             self._template_root = os.path.join(pkgdir, package_path)
         elif spec.submodule_search_locations:
-            # this will be one element for "packages" and multiple for
-            # namespace packages
+            # This will be one element for regular packages and multiple
+            # for namespace packages.
             for root in spec.submodule_search_locations:
                 root = os.path.join(root, package_path)
 
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -352,21 +352,13 @@ def test_package_zip_list(package_zip_loader):
     assert package_zip_loader.list_templates() == ["foo/test.html", "test.html"]
 
 
-def test_pep_451_import_hook(tmp_path):
-    package_name = "_my_pep451_pkg"
-    pkg = tmp_path.joinpath(package_name)
-    pkg.mkdir()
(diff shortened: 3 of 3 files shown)


davidism:
Can this be applied to 2.11.x? That would require it to still support Python 2.7.

asottile:
the importlib machinery is new in python 3 (as are pep 451 loaders) so I'm not sure it would help python2.7?

davidism:
The current implementation does work in 2.7, it avoids the Python 3 specific stuff. It would be nice to fix this in 2.11.x if possible.

More radical would be reverting the removal of `pkg_resources` and making it part of 3.0 instead.

asottile:
it could be backported with version guards probably

gaborbernat:
Here's a small repro:

https://github.com/gaborbernat/jinja-pytest

running tox shows:
```
ERROR:   bug: commands failed
  old: commands succeeded
ERROR:   fix: commands failed
```

gaborbernat:
So forgot to package the templates, which throw the same error. Fixed that and now my report prooves that this PR solves the issue, but latest released is still broken.

gaborbernat:
@davidism can we review/land this? 😊

davidism:
I'm confused, it looks like your output shows this fix didn't work.

davidism:
I think this needs to be in 2.11.x. Maybe something like the following?

```python
try:
    from importlib.util import find_spec

    spec = importlib.util.find_spec(package_name)
    loader = spec.loader
except ImportError:
    spec = None
    loader = pkgutil.get_loader(package_name)
````

gaborbernat:
Fine by me 👍

gaborbernat:
Soft ping on what needs to happen for this 👍

davidism:
Currently in the middle of some work on Werkzeug, then will get to this. If you or Anthony can move this to 2.11.x, that would let it happen sooner.

gaborbernat:
Soft ping on this 👍

davidism:
Working on backporting it.

davidism:
Backporting this makes it super complicated, and doesn't address the issue on Python 2. Debating just requiring importlib_resources for < 3.9.

asottile:
is this a problem in python2? the new style loaders weren't introduced until python3

davidism:
`pkgutil.get_loader` is used for both versions, but pytest doesn't define `get_filename` in either.

asottile:
the 4.6 edition of pytest still uses a PEP 302 loader: https://github.com/pytest-dev/pytest/blob/020831d868319dbfa54046fdcce5fd853e20d281/src/_pytest/assertion/rewrite.py#L321-L325

davidism:
That's not enough for what `PackageLoader` needs. That just allows opening files, not zip files, and doesn't allow listing the files.

@gaborbernat I think regardless of what the fix in Jinja is, Pytest needs to support `get_filename`, just like it added support for PEP 302 `get_data`. It's [only rewriting `SourceFileLoader`](https://github.com/pytest-dev/pytest/blob/211adfc94548e26b64c85d87b6bf793da23e0a7f/src/_pytest/assertion/rewrite.py#L92-L93), so it should be providing the same API, otherwise it may be breaking compatibility with other libraries that assume they're installed as source files.

davidism:
How do we feel about using `importlib_resources` for Python < 3.9? Its `files` API is exactly what's needed here and is compatible with all supported Python versions. (I'm assuming @jaraco will get the new API into 3.9, it's not in the docs yet though.)

asottile:
> Pytest needs to support get_filename, just like it added support for PEP 302 get_data

the 4.6 maintenance branch is frozen (where the PEP 302 importer is implemented) and get_filename is not part of PEP 451 (pytest 5+) -- the import hook on the 4.6 branch is basically what it has been since creation and hasn't caused problems elsewhere -- the dependence on `get_filename` is the bug here as it's specified optional in the PEP

I don't really think importlib-resources solves anything here that the vanilla apis (origin / submodule_search_locations) already give us

davidism:
`get_data` is still present in the modern loader. `get_filename` is optional, but for the only loader that Pytest is rewriting, it *is* implemented, so I don't see why Pytest shouldn't mirror that.

importlib_resources appeals to me because it provides one API for inspecting, opening, and listing folders and zip files, and is compatbile with all versions without a bunch of hacks. The only thing it might not support is namespace packages, but since the `files` API just returns a path-like object, we can return a path-like object when detecting a namespace.

davidism:
I should just revert the removal of pkg_resources in 2.11. The main source of problems is trying to remain compatible with Python 2. Dealing with folder/zip/namespace is minor compared to dealing with not having PEP 451 in Python 2.

asottile:
> I don't see why Pytest shouldn't mirror that.

because (1) it's difficult, pytest changes the module layout entirely (2) it's optional, a consumer can derive the path that they want (and often a better answer than pytest can provide you) from the information present

___

(not my project but) my 2c: I would maybe consider reverting whatever broke this on the 2.11.x branch and then moving forward with this patch in 3.x

davidism:
We're thinking the same thing. :-)

davidism:
OK, reverted in #1182, now I just have to merge that forward to master and rebase this.

jaraco:
> How do we feel about using `importlib_resources` for Python < 3.9? Its `files` API is exactly what's needed here and is compatible with all supported Python versions. (I'm assuming @jaraco will get the new API into 3.9, it's not in the docs yet though.)

I'd recommend this approach. In fact, just rely on `importlib_resources` unconditionally until it lands upstream. It's very [likely to land in 3.9](https://bugs.python.org/issue39791), though there's a non-trivial chance some [later reviews on the design](https://gitlab.com/python-devs/importlib_resources/-/issues/90) might push out the adoption into Python 3.9.

PR #1178, opened 2020-03-27
Fix tokens line number calculation when whitespace stripping is used

avli:
The current version seems to assign wrong line numbers to tokens when whitespace trimming is used. For example this code:

```python

if __name__ == '__main__':
    env = Environment()
    tmpl = env.from_string(
"""\
<html>
<body>
{%- block content -%}
    <hr>
    {{ item }}
{% endblock %}
</body>
</html>"""
    )
    print(tmpl.debug_info)
```

prints `[(2, 12), (4, 23)]` when it should be `[(3, 12), (5, 23)]`. In particular, this causes troubles for debuggers (see [here](https://youtrack.jetbrains.com/issue/PY-40968)).

The suggested patch attempts to fix the line number calculation. Please let me know if I have missed something. Thanks in advance!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,8 @@ Unreleased
     when using Pytest. Due to the difficulty in supporting Python 2 and
     :pep:`451` simultaneously, the changes are reverted until 3.0.
     :pr:`1182`
+-   Fix line numbers in error messages when newlines are stripped.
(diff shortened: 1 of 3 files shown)


davidism:
Thanks for the fix! This will be in 2.11.2.

Issue #1180, opened 2020-03-28
Using namespaces asynchronously raises fatal AttributeError


bartfeenstra:
When asynchronously rendering templates that make use of namespaces, rendering fails on `AttributeError: __class__`.

## Expected Behavior
No error is raised.

## Actual Behavior
Using namespaces when rendering templates asynchronously fails on the `isawaitable()` check performed on `Namespace`.

## Template Code
```jinja
{% set ns = namespace(foo=[]) %}
```

## Reproduction:
```python
import asyncio

from jinja2 import Environment

async def main():
    environment = Environment(
        enable_async=True
    )
    template = '{% set ns = namespace(foo=[]) %}'
    await environment.from_string(template).render_async()

asyncio.get_event_loop().run_until_complete(main())
```

## Full Traceback
```pytb
Traceback (most recent call last):
  File "/home/bart/tmp/jinja2-async-namespacce/venv/lib/python3.8/site-packages/jinja2/utils.py", line 699, in __getattribute__
    return self.__attrs[name]
KeyError: '__class__'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "namespace.py", line 12, in <module>
    asyncio.get_event_loop().run_until_complete(main())
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "namespace.py", line 10, in main
    await environment.from_string(template).render_async()
  File "/home/bart/tmp/jinja2-async-namespacce/venv/lib/python3.8/site-packages/jinja2/asyncsupport.py", line 65, in render_async
    return self.environment.handle_exception()
  File "/home/bart/tmp/jinja2-async-namespacce/venv/lib/python3.8/site-packages/jinja2/environment.py", line 832, in handle_exception
    reraise(*rewrite_traceback_stack(source=source))
  File "/home/bart/tmp/jinja2-async-namespacce/venv/lib/python3.8/site-packages/jinja2/_compat.py", line 28, in reraise
    raise value.with_traceback(tb)
  File "/home/bart/tmp/jinja2-async-namespacce/venv/lib/python3.8/site-packages/jinja2/asyncsupport.py", line 25, in concat_async
    await collect()
  File "/home/bart/tmp/jinja2-async-namespacce/venv/lib/python3.8/site-packages/jinja2/asyncsupport.py", line 22, in collect
    async for event in async_gen:
  File "<template>", line 1, in top-level template code
[...]

PR #1181, opened 2020-03-28
Allow namespaces to be used in asynchronously rendered templates.

bartfeenstra:
This fixes https://github.com/pallets/jinja/issues/1180, by allowing `Namespace` instances to undergo the type checks necessary for async compatibility.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,8 @@ Unreleased
     :pr:`1182`
 -   Fix line numbers in error messages when newlines are stripped.
     :pr:`1178`
+-   The special ``namespace()`` assignment object in templates works in
+    async environments. :issue:`1180`
 
 
 Version 2.11.1
--- a/src/jinja2/utils.py
+++ b/src/jinja2/utils.py
@@ -697,7 +697,8 @@ class Namespace(object):
         self.__attrs = dict(*args, **kwargs)
 
     def __getattribute__(self, name):
-        if name == "_Namespace__attrs":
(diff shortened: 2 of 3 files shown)


tests/test_utils.py
```diff
@@ -188,3 +190,9 @@ def test_consume():
     consume(x)
     with pytest.raises(StopIteration):
         next(x)
+
+
+class TestNamespace:
+    def test_isawaitable(self):
+        """Test async compatibility by asserting the required type check works."""
+        assert not inspect.isawaitable(Namespace())
```
ThiefMaster: Wouldn't it be better to have a test that tests the namespace object in an actual async template context?

bartfeenstra: I wasn't sure how extensive we wanted the test to be, so I decided to keep it small initially. It's now updated with template integration coverage.

tests/test_utils.py
```diff
@@ -188,3 +190,16 @@ def test_consume():
...
+
+class TestNamespace:
+    def test_isawaitable(self):
+        """Test async compatibility."""
+
+        async def _test():
+            environment = Environment(enable_async=True)
+            template = '{% set ns = namespace(foo="Bar") %}{{ ns.foo }}'
+            actual = await environment.from_string(template).render_async()
+            assert "Bar" == actual
```
ThiefMaster: No yoda style please ;) => `assert actual == "Bar"`

bartfeenstra: What you ask, I have done.

PR #1183, opened 2020-04-02
Fix unintended lstrip_blocks behavior

### LINKED ISSUE #1138: lstrip_blocks doesn't preserve space before end block on the same line
Look at the following template code:

```jinja2
set_property verilog_define {
  {%- for k, v in vlogdefine.items() %}{{ k }}={{ v|param_value_str }} {% endfor -%}
  } [get_filesets sources_1]
```

https://github.com/olofk/edalize/blob/bdb6c9ccc666e9f60333279ad53ed09cda88b3dc/edalize/templates/vivado/vivado-project.tcl.j2#L27-L29

Up to Jinja2 2.10.3 this would result in something like

```
set_property verilog_define {vlogdefine_bool=1 vlogdefine_int=42 vlogdefine_str=hello } [get_filesets sources_1]
```
https://github.com/olofk/edalize/blob/bdb6c9ccc666e9f60333279ad53ed09cda88b3dc/tests/test_vivado/test_vivado_0.tcl#L10

[The code](https://github.com/olofk/edalize/blob/bdb6c9ccc666e9f60333279ad53ed09cda88b3dc/edalize/edatool.py#L81) has `lstrip_blocks = True`.

Now with 2.11.0 the code is rendered like this:

```
set_property verilog_define {vlogdefine_bool=1vlogdefine_int=42vlogdefine_str=hello} [get_filesets sources_1]
```

@towoe bisected the behavior change down to https://github.com/pallets/jinja/commit/7d00a40465c89bee141ab5a3db545a20e7d30509 (by @petee-d in https://github.com/pallets/jinja/issues/857)

According to the [documentation](https://jinja.palletsprojects.com/en/2.11.x/api/)

> lstrip_blocks
>    If this is set to True leading spaces and tabs are stripped from the start of a line to a block. Defaults to False.

I don't think the whitespace should be removed before the `endfor`, since that's not whitespace "from the beginning of the line". Am I misunderstanding that, or does the new version introduce an unexpected change in behavior?

Python 3.7.3 (but also happens on 3.5 in CI)

petee-d:
Introduced in #858. Closes #1138

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,8 @@ Unreleased
     :pr:`1178`
 -   The special ``namespace()`` assignment object in templates works in
     async environments. :issue:`1180`
+-   Fix whitespace being removed before tags in the middle of lines when
+    ``lstrip_blocks`` is enabled. :issue:`1138`
 
 
 Version 2.11.1
--- a/src/jinja2/lexer.py
+++ b/src/jinja2/lexer.py
@@ -682,6 +682,7 @@ class Lexer(object):
         balancing_stack = []
         lstrip_unless_re = self.lstrip_unless_re
         newlines_stripped = 0
+        line_starting = True
 
         while 1:
             # tokenizer loop
@@ -731,11 +732,11 @@ class Lexer(object):
                         ):
                             # The start of text between the last newline and the tag.
                             l_pos = text.rfind("\n") + 1
-
-                            # If there's only whitespace between the newline and the
-                            # tag, strip it.
-                            if not lstrip_unless_re.search(text, l_pos):
-                                groups = (text[:l_pos],) + groups[1:]
+                            if l_pos > 0 or line_starting:
+                                # If there's only whitespace between the newline and the
+                                # tag, strip it.
+                                if not lstrip_unless_re.search(text, l_pos):
+                                    groups = (text[:l_pos],) + groups[1:]
 
                     for idx, token in enumerate(tokens):
                         # failure group
@@ -793,6 +794,8 @@ class Lexer(object):
                     if data or tokens not in ignore_if_empty:
                         yield lineno, tokens, data
                     lineno += data.count("\n")
+
+                line_starting = m.group()[-1:] == "\n"
 
                 # fetch new position into new variable so that we can check
                 # if there is a internal parsing error which would result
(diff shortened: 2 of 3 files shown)


petee-d:
I ran a comparison on the resulting parse tree on 5GBs worth of over 100k Jinja templates created by users with various technical backgrounds and saw no difference. So apparently none of these relied on the behavior that this fixes, but it also shows it's unlikely this produces another undesired change of behavior.

I also performance tested this and there is no measurable change in performance. Also including the performance before the first fix that introduced this change.
image

petee-d:
Today I got back to this for a bit and realized that the way I was testing it on my 100k sample templates was wrong, I didn't actually enable lstrip_blocks in my environment. So I repeated the test and the result is great. The generated parse trees after #857 were actually different in many cases from the parse trees generated before #857, and the parse trees after #1183 were in all cases the same as the ones before #857. I made the same mistake when comparing the parse trees when implementing #857 so this shows how the change of behavior went unnoticed and it also confirms that this fix returns to the correct behavior while keeping the much better performance.

davidism:
Awesome work, thanks for following up. I've actually got this checked out right now, rebasing to 2.11.x and adding a changelog. Should be released later today.

davidism:
I'm not sure how you managed to do this, but you opened a PR from a different user (xponea), so I can't force push to the branch.

Can you do the following:

```
git rebase --onto origin/2.11.x origin/master
nano CHANGES.rst  # add changelog entry to bottom of 2.11.2 section, given below
git commit -a --amend --no-edit
git push -f
```

Changelog entry:

```
-   Fix whitespace being removed before tags in the middle of lines with
    ``lstrip_blocks`` enabled. :issue:`1138`
```

davidism:
@petee-d ping on getting this rebased

petee-d:
@davidism Sorry for the delay, I was mostly offline the past few days. Pushed the rebase and change log entry to my fork branch. I don't see the option to let you add commit to this PR created from a fork, as described in https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork. If you need to do more edits, feel free to copy the branch from my fork and recreate the PR.

davidism:
Looks good, thanks!

PR #1185, opened 2020-04-08
Update oudated CutomLoader code example

xinbinhuang:
Simple typo fits on the example doc

### DIFF
--- a/src/jinja2/loaders.py
+++ b/src/jinja2/loaders.py
@@ -56,7 +56,7 @@ class BaseLoader:
                     raise TemplateNotFound(template)
                 mtime = getmtime(path)
                 with open(path) as f:
-                    source = f.read().decode('utf-8')
+                    source = f.read()
                 return source, path, lambda: mtime == getmtime(path)
     """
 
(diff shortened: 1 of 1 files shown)


src/jinja2/loaders.py
```diff
@@ -55,7 +55,7 @@ def get_source(self, environment, template):
                 if not exists(path):
                     raise TemplateNotFound(template)
                 mtime = getmtime(path)
-                with file(path) as f:
+                with open(path) as f:
                     source = f.read().decode('utf-8')
```
davidism, line 59: Decode needs to be removed since read already returns a string, not bytes.

davidism:
It's not a typo, it's just a really, really old example. Needs more updating for Python 3.

xinbinhuang:
> It's not a typo, it's just a really, really old example. Needs more updating for Python 3.

@davidism I see. Thanks for the info. Can I assume that the codebase is all updated to Python 3 (I saw a commit using `pyupgrade`), but some code examples may still be outdated? 

I will make some PRs if I notice more.

Thanks
Bin

davidism:
The code in master is all Python 3 now, but I didn't have the capacity to do all the docs too. https://github.com/pallets/meta/issues/26 is about updating the docs for all projects.

Issue #1187, opened 2020-04-10
Can't handle 'list' and 'dict' typed value in the 'unique' filter


userlocalhost:
I want to let you know that `TypeError` exception would be happened when `list` and `dict` typed value would be passed to the `unique` filter.

This is a test code that passes variety of typed value to the `unique` filter. When it's run, the last two processes would be fault.
```
import jinja2

expression = """
{%- for val in values|unique -%}
  {{ val }}
{%- endfor %}
"""

env = jinja2.Environment()

env.from_string(expression).render(values=[1,2,3,1])                  # PASS
env.from_string(expression).render(values=['a','b','c','a'])          # PASS
env.from_string(expression).render(values=[(1,), (2,), (3,), (1,)])   # PASS
env.from_string(expression).render(values=[[1], [2], [3], [1]])       # NG
env.from_string(expression).render(values=[{'a':1},{'b':2},{'a':1}])  # NG
```

## Expected Behavior
```
>>> env.from_string(expression).render(values=[[1], [2], [3], [1]])
'[1][2][3]'
```
```
>>> env.from_string(expression).render(values=[{'a':1},{'b':2},{'a':1}])
"{'a': 1}{'b': 2}"
>>> 
```

## Actual Behavior
```
>>> env.from_string(expression).render(values=[[1], [2], [3], [1]])
...
TypeError: unhashable type: 'list'
```
```
>> env.from_string(expression).render(values=[{'a':1},{'b':2},{'a':1}])
...
TypeError: unhashable type: 'dict'
```

## Full Traceback
This is an execution result of above code.
```
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import jinja2
>>> 
>>> jinja2.__version__
'3.0.0a1'
>>> 
>>> expression = """
... {%- for val in values|unique -%}
...   {{ val }}
... {%- endfor %}
... """
>>> 
>>> env = jinja2.Environment()
>>> 
>>> env.from_string(expression).render(values=[1,2,3,1])                  # PASS
'123'
>>> env.from_string(expression).render(values=['a','b','c','a'])          # PASS
'abc'
>>> env.from_string(expression).render(values=[(1,), (2,), (3,), (1,)])   # PASS
'(1,)(2,)(3,)'
>>> env.from_string(expression).render(values=[[1], [2], [3], [1]])       # NG
Traceback (most recent call last):
[...]

PR #1188, opened 2020-04-11
Fixed a bug that unique filter couldn't handle 'list', 'dict' and 'set' type values

userlocalhost:
This fixes a problem of #1187.

After applying this patch, I confirmed that unique filter would work as expected as below.
スクリーンショット 2020-04-11 18 33 09

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -200,7 +200,7 @@ class TestConvertToBeHashable:
         for value in self.TEST_DATA:
             assert isinstance(hash(convert_value_to_be_hashable(value)), int)
 
-    def test_nexted_type_of_data(self):
+    def test_nested_type_of_data(self):
         """
         This tests the convert_value_to_be_hashable method could convert
         a value that list and dict are nested.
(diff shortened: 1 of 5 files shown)


tests/test_utils.py
```diff
@@ -177,6 +179,50 @@ def test_lorem_ipsum_max(self):
...
+
+    def test_convert_data(self):
+        """
+        This tests the convert_value_to_be_hashable method could handle
+        variety of built-in types that jinja2 would accept.
+        """
+        for value in self.TEST_DATA:
+            assert isinstance(hash(convert_value_to_be_hashable(value)), int)
+
+    def test_nexted_type_of_data(self):
```
ThiefMaster, line 203: nexted?

userlocalhost: Thank you for your review. Good catch!

davidism:
Thanks for looking at this. However, this isn't a bug, it's a standard behavior of Python. I don't want to add the overhead of recursively creating objects just to throw them away in this filter. If your project needs a unique filter that handles these types, you can add a custom filter.

PR #1189, opened 2020-04-13
Fix some code quality issues

pnijhara:
- Remove unnecessary comprehension
- Remove assert statement from non-test files
- Change methods not using its bound instance to static methods

---

Find the other issues found here - [https://deepsource.io/gh/pnijhara/jinja/issues/?category=all](https://deepsource.io/gh/pnijhara/jinja/issues/?category=all)

This PR also adds `.deepsource.toml` configuration file to run DeepSource analysis on the repo with. Upon enabling DeepSource, the analysis will run on every PR and commit to detect 560+ types of issues in the changes — including bug risks, anti-patterns, security vulnerabilities, etc.

To enable DeepSource analysis after merging this PR, please follow these steps:

1. [Signup](https://deepsource.io/signup/) on DeepSource with your GitHub account and grant access to this repo.
2. Activate analysis on this repo [here](http://deepsource.io/gh/pallets/jinja).

You can also look at the [docs](https://deepsource.io/docs/guides/quickstart.html) for more details. Do let me know if I can be of any help!

### DIFF
--- a/src/jinja2/loaders.py
+++ b/src/jinja2/loaders.py
@@ -90,7 +90,8 @@ class BaseLoader:
             )
         raise TemplateNotFound(template)
 
-    def list_templates(self):
+    @staticmethod
+    def list_templates():
         """Iterates over all templates.  If the loader does not support that
         it should raise a :exc:`TypeError` which is the default behavior.
         """
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -25,7 +25,8 @@ from jinja2.utils import evalcontextfunction
 
 
 class TestExtendedAPI:
-    def test_item_and_attribute(self, env):
+    @staticmethod
+    def test_item_and_attribute(env):
         from jinja2.sandbox import SandboxedEnvironment
 
         for env in Environment(), SandboxedEnvironment():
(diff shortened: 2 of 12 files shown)


src/jinja2/loaders.py
```diff
@@ -90,7 +90,8 @@ def get_source(self, environment, template):
             )
         raise TemplateNotFound(template)
 
-    def list_templates(self):
+    @staticmethod
+    def list_templates():
```
ThiefMaster, line 94: Did you look at the code your PR changes? This is a method that gets overridden in some cases. So making it static in the base version with no functionality is somewhere between pointless and wrong, since a class overriding it will most likely *not* make it static...

tests/test_api.py
```diff
@@ -263,11 +284,13 @@ def test():
...
 
-    def test_undefined_and_special_attributes(self):
+    @staticmethod
+    def test_undefined_and_special_attributes():
         with pytest.raises(AttributeError):
             Undefined("Foo").__dict__
 
-    def test_undefined_attribute_error(self):
+    @staticmethod
+    def test_undefined_attribute_error():
```
ThiefMaster, line 293: Just picking another of these changes randomly: I don't think there's any benefit in making methods static even if they don't use self as long as they usually get called through an instance of the class and not the class itself.

ThiefMaster:
I don't think most (or "any besidess the first"?) of these changes are particularly useful and improve anything.

PR #1190, opened 2020-04-13
native only evals at end of render

### LINKED ISSUE #1186: NativeEnv evals constants twice
Rendering in native mode performs ```literal_eval``` twice on constant values, which leads to unexpected behaviour.

## Expected Behavior
```>>> from jinja2.nativetypes import NativeEnvironment
>>> NativeEnvironment().from_string(r'''0.000{{ a }}''').render({"a":7})
0.0007
>>> NativeEnvironment().from_string(r'''0.000{{ 7 }}''').render({"a":7})
0.0007
```
Templates should behave the same way with constants as they do with dynamic variables.

Nothing should be eval'd before the entire template is finished.

## Actual Behavior
```>>> from jinja2.nativetypes import NativeEnvironment
>>> NativeEnvironment().from_string(r'''0.000{{ a }}''').render({"a":7})
0.07
>>> NativeEnvironment().from_string(r'''0.000{{ 7 }}''').render({"a":7})
0.0007
>>> NativeEnvironment().from_string(r'''{{b}}{{ a }}''').render({"a":7,"b":"0.000"})
0.0007
```

In the first case the constant ```0.000``` gets eval'd at template compile time, and truncated to ```0.0``` before the entire template has been rendered with dynamic data. The template gets eval'd a second time at render time, meaning the constant value has been double eval'd.

In the second case the whole template is constant, and gets eval'd in one go at compile time producing different results.

In the third case the whole template is dynamic, and gets eval'd in one go at render time.

## Your Environment
* Python version: 3.7.6
* Jinja version: 2.11.1

## Suggested solution

```_output_const_repr()``` should not perform any evals, only ```render()``` should perform one eval at the very end. This seems to me like the only sane way to prevent weird double-eval issues like this one.

The ```preserve_quotes``` workaround in ```native_concat``` would no longer be needed with this change. It appears to have been a previous attempt at fixing this class of problems.

Fixing it this way would also fix the same problem in the underlying ```root_render_func``` api:
```
>>> t = NativeEnvironment().from_string(r'''{{ a }}''')
>>> list(t.root_render_func(t.new_context({"a":"7"}))) 
['7']
>>> t = NativeEnvironment().from_string(r'''{{"7"}}''')
>>> list(t.root_render_func(t.new_context({"a":"7"}))) 
[7]
```

davidism:
If each intermediate group was converted, it could change the final value. The example given in the test is that the string `0.000` evaluates to a float, then renders as `0.0` when combined later, so `0.000{{ 7 }}` would result in `0.07`. Instead, intermediate groups are joined as strings, and only the final result is passed to `native_concat`. This also removes the need for `preserve_quotes` from #1086.

submitted by @mkrizek, closes #1186

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,10 @@ Unreleased
     async environments. :issue:`1180`
 -   Fix whitespace being removed before tags in the middle of lines when
     ``lstrip_blocks`` is enabled. :issue:`1138`
+-   :class:`~nativetypes.NativeEnvironment` doesn't evaluate
+    intermediate strings during rendering. This prevents early
+    evaluation which could change the value of an expression.
+    :issue:`1186`
 
 
 Version 2.11.1
--- a/src/jinja2/nativetypes.py
+++ b/src/jinja2/nativetypes.py
@@ -1,4 +1,3 @@
-import types
 from ast import literal_eval
 from itertools import chain
 from itertools import islice
@@ -11,7 +10,7 @@ from .environment import Environment
 from .environment import Template
 
 
-def native_concat(nodes, preserve_quotes=True):
+def native_concat(nodes):
     """Return a native Python type from the list of compiled nodes. If
     the result is a single node, its value is returned. Otherwise, the
     nodes are concatenated as strings. If the result can be parsed with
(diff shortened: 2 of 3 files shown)


davidism:
`math.isclose` is Python 3 only. *Shakes fist at Python 2.*

PR #1195, opened 2020-04-19
Improve and extend urlize

### LINKED ISSUE #522: urlize filter ignores ftp:// URLs

### LINKED ISSUE #827: The `urlize` function does not compose parentheses correctly at the end
## Expected Behavior
Final parentheses should be within the <a> tag and link should be exactly as given before filtering.

## Actual Behavior
Final parentheses is outside the <a> tag and final parentheses is lost in url construction

## Template Code
```python
from jinja2.utils import urlize
example_url = u'https://de.wikipedia.org/wiki/Agilit%C3%A4t_(Management)'
urlize(example_url, 50)
# u'<a href="https://de.wikipedia.org/wiki/Agilit%C3%A4t_(Management">https://de.wikipedia.org/wiki/Agilit%C3%A4t_(Manag...</a>)'
```

## Your Environment
* Python version: 2.7.12
* Jinja version: 2.8

### LINKED ISSUE #1172: Make urlize filter support mailto: scheme
## Expected Behavior
The urlize filter should recognize the `mailto:` URI scheme and create a corresponding link.
```python
>>> env.from_string('{{ "mailto:webmaster@example.com"|urlize }}').render()
'<a href="mailto:webmaster@example.com" rel="noopener">webmaster@example.com</a>'
```

## Actual Behavior
Instead the mail address does not get recognized as an address suitable for an anchor and is not modified by the filter.
```python
>>> env.from_string('{{ "mailto:webmaster@example.com"|urlize }}').render()
'mailto:webmaster@example.com'
```
I already looked into the source code and noticed that urlize is only supported for `http(s)://` and not other protocols. However I think that albeit being rarer than normal web links mail links or other links should also be supported. E.g. `ftp://` and also other schemes like `tel:`, `data:` or `apt:`.
As those would be hard to dynamically locate inside of a space separated string a better alternative would be to add an option to the filter so that it treats the whole passed string as one link naively.

## Template Code
```jinja
{{ "mailto:webmaster@example.com"|urlize }}
```

## Your Environment
* Python version: 3.7
* Jinja version: 2.10

bebleo:
Fixes #522
Fixes #827 
Fixes #1172

The 'urlize' function does not as expected both in terms of some missing support. Among a few a few of the behaviors were a lack of support for 5/8 of the original TLDs, emails prefixed with `mailto:` would not be returned as links, and no support for `ftp://`.

This PR does not "fix" the latter point, but does add a `policy` that can be set so that the function will create the links.  This can also be used to add support for any arbitrary URI scheme.

Specifically:
- Adds mailto: link support (issue #1172)
- Adds an `extra_uri_schemes` policy and parameter to address lack of `ftp:` support and allow for other schemes (for example `tel:`) to be added as needed. (issue #522 and PR #587)
- Parentheses at the end of a URL are now included if they match an opening parenthesis in the URL (issue #827)
- Adds support for all of the original top-level domains when creating links such as example.org
- Uses the same regex used to validate URLs to ensure that ipv4 and ipv6 `https//` and `https://` links are well-formed.
- Sets the default URI scheme for URLs to be `https` instead of `http`

### DIFF
--- a/src/jinja2/utils.py
+++ b/src/jinja2/utils.py
@@ -168,12 +168,6 @@ def pformat(obj):
     return pformat(obj)
 
 
-_word_split_re = re.compile(r"(\s+)")
-_lead_pattern = "|".join(map(re.escape, ("(", "<", "&lt;")))
-_trail_pattern = "|".join(map(re.escape, (".", ",", ")", ">", "\n", "&gt;")))
-_punctuation_re = re.compile(
-    fr"^(?P<lead>(?:{_lead_pattern})*)(?P<middle>.*?)(?P<trail>(?:{_trail_pattern})*)$"
-)
 _http_re = re.compile(
     r"""
     ^
@@ -202,7 +196,7 @@ _http_re = re.compile(
     """,
     re.IGNORECASE | re.VERBOSE,
 )
-_simple_email_re = re.compile(r"^\S+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9._-]+$")
+_email_re = re.compile(r"^\S+@\w[\w.-]*\.\w+$")
 
 
 def urlize(text, trim_url_limit=None, rel=None, target=None, extra_schemes=None):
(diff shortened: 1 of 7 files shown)


davidism:
* updated docstrings
* moved regexes from the top of the module to above the function
* turned the http regex into a verbose one with comments since it's complex
* renamed `extra_uri_schemes` to `extra_schemes`

davidism:
After looking over the code more, I identified a number of optimizations, so I committed another refactor.

src/jinja2/utils.py
```diff
@@ -237,71 +231,85 @@ def urlize(text, trim_url_limit=None, rel=None, target=None, extra_schemes=None)
...
 
-    words = _word_split_re.split(str(escape(text)))
+    words = re.split(r"(\s+)", str(escape(text)))
     rel_attr = f' rel="{escape(rel)}"' if rel else ""
     target_attr = f' target="{escape(target)}"' if target else ""
 
     for i, word in enumerate(words):
-        match = _punctuation_re.match(word)
+        head, middle, tail = "", word, ""
+        match = re.match(r"^([(<]|&lt;)+", middle)
```
jab, line 253: This calls `re.match()` with the same pattern string for each iteration of the loop (and similar is done below). I don’t think the `re` module caches recently- or frequently-compiled patterns, in which case these calls would be needlessly recompiling the same pattern in every iteration, is that right? If so, is it worth compiling these patterns once, and reusing the resulting pattern objects here?

davidism: `re` caches the last 512 patterns used. Compiling every single pattern at import time increases startup time for projects, I'd seen some more (general) discussions about this recently so decided these could be inlined.

jab: Oh interesting, good to know, thanks!

PR #1208, opened 2020-05-16
api.rst: use simpler FileSystemLoader

mhansen:
Down below in this class says FileSystemLoader 'is the preferred way to load them.'

At any rate, it's simpler than PackageLoader, which tripped me up - I wasn't sure what string to use for 'yourapplication'. I'm not the only one - see https://stackoverflow.com/questions/14276829/html-templating-using-jinja2-lost.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -28,18 +28,18 @@ are in use.
 The simplest way to configure Jinja to load templates for your application
 looks roughly like this::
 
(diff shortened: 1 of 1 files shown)


davidism:
Thanks, but I'm happy with package loader here.

mhansen:
OK, no worries, thanks for taking a look!

FWIW I've filed an experience report about some doc issues I hit during my bumpy journey to Hello World at #1211.

PR #1209, opened 2020-05-16
do_dictsort: update example ready to copy/paste

mhansen:
When iterating over a dict you usually want to pull out the keys and values.

### DIFF
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -268,16 +268,16 @@ def do_dictsort(value, case_sensitive=False, by="key", reverse=False):
 
     .. sourcecode:: jinja
 
-        {% for item in mydict|dictsort %}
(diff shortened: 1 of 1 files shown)


davidism:
Thanks for looking at this. I don't find it to be significant enough to require a change.

mhansen:
Thanks for reviewing! Sorry I should have explained ‘why’ / my experience report: it’s because I wasn’t sure if Jinja would be default look over items, keys, or values. I assumed items, but then I wasn’t sure if accessing the key and value from items would be like in normal python ([0] or [1] I think? Or would it be .key and .value? Or is it just looping over keys, like in normal python if you say “for key in dict:”?

Anyway I searched for another example to show me how people are grabbing our keys and values and found this.

PR #1210, opened 2020-05-16
Added Not-A-Number (`float('nan')`) check to `tests.test_number`; i.e…

### LINKED ISSUE #1206: tests.test_number validates to True for NaN
`float('nan')` explicitly represents "**Not** a Number" and is used as such by numpy and pandas, but is a number according to jinja2's `number` test.

## Expected Behavior
`jinja2.tests.test_number(float('nan')) -> False`

## Actual Behavior
`jinja2.tests.test_number(float('nan')) -> True`

## Environment
* Python version: any
* Jinja version: at least 2.11.1 and up, most likely any

## Fix
This will need its own explicit check – append `and not math.isnan(value)` to the boolean expression in `test_number`. *Please note that `==` and `!=` **don't** work as expected with `float('nan')`.*

phryk:
…. made it return False when given NaN.

Fixes #1206

### DIFF
--- a/src/jinja2/tests.py
+++ b/src/jinja2/tests.py
@@ -2,6 +2,7 @@
 import operator
 import re
 from collections import abc
+from math import isnan
 from numbers import Number
 
 from .runtime import Undefined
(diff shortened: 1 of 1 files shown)


davidism:
Despite its name `is number` is checking the type of the object, not the value. Despite its name, `nan` is a number type. I don't think it's good to add value checking to this test. If you're using Jinja in a context where you need this check, it can be added to that project, it doesn't need to be in Jinja.

PR #1214, opened 2020-05-18
use pip-compile to pin dev requirements

davidism:
Sphinx is pinned at `<3` right now because in 3.0 it started raising errors about duplicate references, and the Jinja extension in Pallets-Sphinx-Themes doesn't differentiate between the names of filters and tests like `float` right now.

Added `-e file:.` to ensure the editable install took precedence over the `Jinja2==2.11.1` coming from Sphinx. Switched the order of the requirements and pip lines in the readthedocs config for the same reason.


davidism:
Opened an issue about the names: pallets/pallets-sphinx-themes#29.

PR #1233, opened 2020-06-12
Add required attribute to blocks

### LINKED ISSUE #1147: {% block <name> required %}
This is a feature request for blocks which are required to be implemented by the leaf template. One example of where this would be useful is in title blocks. In the parent template:

```jinja
{% block title required %}{% endblock %}
```

And in the child:

```jinja
{% extends "example.html" %}
{% block title %}
<title>My Page</title>
{% endblock %}
```

If the title block is not present in the latter, an error is rased. If the block has any contents other than whitespace or comments in the former, an error is raised. In effect, this requires all pages to have titles.

amy-lei:
Resolves #1147 

Summary:
 - Added required field to blocks
   - Block nodes now have 4 fields (`scoped` must come before `required` if used together)
   - Must be overridden at some point, although not necessarily by the direct child. `TemplateRuntimeError` raised otherwise.
   - Required block itself cannot contain anything other than comments or whitespace. `TemplateSyntaxError` raised otherwise.
   - With current implementation, rendering the template containing the required block will also raise a `TemplateRuntimeError`.
 
Test:
- Added tests to `test_inheritance.py`
  - Tested up to three levels, made sure correct errors are raised, tested that scoped still works in conjunction with required
- Patched `test_idtracking.py` to include the new field when instantiating Block nodes

Files affected:
- `parser.py`
- `compile.py`
- `nodes.py`
- `test_inheritance.py`
- `test_idtracking.py`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,8 @@ Unreleased
 -   Fix UndefinedError incorrectly being thrown on an undefined variable
     instead of ``Undefined`` being returned on
     ``NativeEnvironment`` on Python 3.10. :issue:`1335`
+-   Add ``required`` attribute to blocks that must be overridden at some
+    point, but not necessarily by the direct child :issue:`1147`
 
 
 Version 2.11.2
--- a/src/jinja2/nodes.py
+++ b/src/jinja2/nodes.py
@@ -340,9 +340,13 @@ class With(Stmt):
 
 
 class Block(Stmt):
-    """A node that represents a block."""
+    """A node that represents a block.
 
-    fields = ("name", "body", "scoped")
+    .. versionchanged:: 3.0.0
+        the `required` field was added.
+    """
+
+    fields = ("name", "body", "scoped", "required")
 
 
 class Include(Stmt):
--- a/src/jinja2/parser.py
+++ b/src/jinja2/parser.py
@@ -255,6 +255,7 @@ class Parser:
         node = nodes.Block(lineno=next(self.stream).lineno)
         node.name = self.stream.expect("name").value
         node.scoped = self.stream.skip_if("name:scoped")
+        node.required = self.stream.skip_if("name:required")
 
         # common problem people encounter when switching from django
         # to jinja.  we do not support hyphens in block names, so let's
@@ -266,6 +267,17 @@ class Parser:
             )
 
         node.body = self.parse_statements(("name:endblock",), drop_needle=True)
+
+        # enforce that required blocks only contain whitespace or comments
+        # by asserting that the body, if not empty, is just TemplateData nodes
+        # with whitespace data
+        if node.required and not all(
+            isinstance(child, nodes.TemplateData) and child.data.isspace()
+            for body in node.body
+            for child in body.nodes
+        ):
+            self.fail("Required blocks can only contain comments or whitespace")
+
         self.stream.skip_if("name:" + node.name)
         return node
 
@@ -924,7 +936,6 @@ class Parser:
         finally:
             if end_tokens is not None:
                 self._end_token_stack.pop()
-
         return body
 
     def parse(self):
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -540,6 +540,40 @@ modifier to a block declaration::
 When overriding a block, the `scoped` modifier does not have to be provided.
 
 
+Required Blocks
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Blocks can be marked as required. They must be overridden at some point,
+but not necessarily by the direct child template. Required blocks can
+only contain whitespace or comments, and they cannot be rendered directly.
+
+For example::
+
+    # parent.tmpl
+    body: {% block body required %}{% endblock %}
+
+    # child.tmpl
+    {% extends "parent.tmpl" %}
+
+    # grandchild.tmpl
+    {% extends "child.tmpl" %}
+    {% block body %}Hi from grandchild.{% endblock %}
+
+
+Rendering ``child.tmpl`` will give
+``TemplateRuntimeError``
+
+Rendering ``grandchild.tmpl`` will give
+``Hi from grandchild.``
+
+When combined with ``scoped``, the ``required`` modifier must be placed `after`
+the scoped modifier.  Here are some valid examples::
+
+    {% block body scoped %}{% endblock %}
+    {% block body required %}{% endblock %}
+    {% block body scoped required %}{% endblock %}
+
+
 Template Objects
 ~~~~~~~~~~~~~~~~
 
--- a/src/jinja2/compiler.py
+++ b/src/jinja2/compiler.py
@@ -797,6 +797,15 @@ class CodeGenerator(NodeVisitor):
         else:
             context = self.get_context_ref()
 
+        if node.required:
+            self.writeline(f"if len(context.blocks[{node.name!r}]) <= 1:", node)
+            self.indent()
+            self.writeline(
+                f'raise TemplateRuntimeError("Required block {node.name!r} not found")',
+                node,
+            )
+            self.outdent()
+
         if not self.environment.is_async and frame.buffer is None:
             self.writeline(
                 f"yield from context.blocks[{node.name!r}][0]({context})", node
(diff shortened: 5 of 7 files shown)


src/jinja2/parser.py
```diff
@@ -254,7 +254,12 @@ def parse_autoescape(self):
     def parse_block(self):
         node = nodes.Block(lineno=next(self.stream).lineno)
         node.name = self.stream.expect("name").value
-        node.scoped = self.stream.skip_if("name:scoped")
+
+        node.scoped = False
+        node.required = False
+        while self.stream.current.type == "name":
+            setattr(node, self.stream.current.value, True)
+            next(self.stream)
```
davidism, line 262: I think this block might allow invalid syntax, such as a typo like "scope" instead of "scoped", or other words entirely. It would assign an unused variable, but instead it should raise an error. Would you confirm whether this happens?

src/jinja2/nodes.py
```diff
@@ -339,7 +339,7 @@ class With(Stmt):
 class Block(Stmt):
     """A node that represents a block."""
 
-    fields = ("name", "body", "scoped")
+    fields = ("name", "body", "scoped", "required")
```
davidism, line 342: Add a ``versionchanged`` entry in the docstring to mention this.

CHANGES.rst
```diff
@@ -11,7 +11,9 @@ Unreleased
 -   Remove code that was marked deprecated.
 -   Use :pep:`451` API to load templates with
     :class:`~loaders.PackageLoader`. :issue:`1168`
-
+-   Add required attribute to blocks that must be
```
davidism, line 14: Make "required" a code item by surrounding it with double backticks.

This can be wrapped at 72 characters, all the other lines just happened to be short.

src/jinja2/nodes.py
davidism, line 342: Need to document this attribute in the template docs.

amy-lei: 👍 added it in as version 3.0.0

amy-lei: I just added a section titled `Required Blocks` under `Block Nesting and Scope`. Was this the right place to put it?

src/jinja2/parser.py
amy-lei: Yup you were right. I added a `test_any` before the assignment and raised an error if it fails to fix this. 

This made me realize that duplicate assignments are also allowed (e.g. `{% block body scoped scoped %}` will still go through. Is this a problem we should be concerned about?

amy-lei: > This made me realize that duplicate assignments are also allowed (e.g. `{% block body scoped scoped %}` will still go through. Is this a problem we should be concerned about?

@davidism Just had an idea to address the concern above:
```python
node.scoped = self.stream.skip_if("name:scoped")
node.required = self.stream.skip_if("name:required")
        
node.scoped = self.stream.skip_if("name:scoped") if not node.scoped else node.scoped
node.required = self.stream.skip_if("name:required") if not node.required else node.required
```
This implementation throws an error for invalid syntax as well as duplicated attributes. Would this be a better approach?

davidism: Maybe we should to add a "one of" function to the lexer?

amy-lei: Ooh that's a good idea! So it'll be like `test_any` but for `skip_if`'s

amy-lei: Hmm once I started implementing it, I ran into the same issue of having to know which attribute to set, and setting the attribute to False if not present. This is given that `one_of` has `next_if`'s functionality but over an iterable:
```python
node.scoped = False
node.required = False

token = self.stream.one_of("name:scoped", "name:required")
if token is not None:
    setattr(node, token.value, True)
token = self.stream.one_of("name:scoped", "name:required")
if token is not None:
    setattr(node, token.value, True)
``` 
It doesn't seem much cleaner unless I interpretted `one_of` differently.



Another idea was to add an `look_and_skip_if` to the lexer which operates like `skip_if` but checking the next token as well if the current does not match. 
```python
node.scoped = self.stream.look_and_skip_if("name:scoped")
node.required = self.stream.look_and_skip_if("name:required")
```
I'm leaning towards the second idea, but would love to get your input first.

src/jinja2/lexer.py
```diff
@@ -352,9 +352,24 @@ def next_if(self, expr):
             return next(self)
 
     def skip_if(self, expr):
-        """Like :meth:`next_if` but only returns `True` or `False`."""
+        """Like :meth:`next_if` but only retuprns `True` or `False`."""
```
davidism: typo

src/jinja2/parser.py
```diff
@@ -266,6 +267,18 @@ def parse_block(self):
             )
 
         node.body = self.parse_statements(("name:endblock",), drop_needle=True)
+
+        # enforce that required blocks only contain whitespace or comments
+        # by asserting that the body, if not empty, is just TemplateData nodes
+        # with whitespace data
+        if node.required:
+            try:
+                assert all(
```
davidism, line 276: Don't use an assert here, it will be optimized out when running `python -O`. flake8 is supposed to catch this.

src/jinja2/nodes.py
```diff
@@ -337,9 +337,13 @@ class With(Stmt):
 
 
 class Block(Stmt):
-    """A node that represents a block."""
+    """A node that represents a block.
 
-    fields = ("name", "body", "scoped")
+    .. versionchanged:: 3.0.0
+       the `required` field was added.
```
davidism, line 343: Use 4 spaces to indent here. (I know this isn't consistent in older docs.)

src/jinja2/parser.py
davidism: `look_and_skip_if` looks good to me.

src/jinja2/parser.py
amy-lei: 👍 fixed. (accidentally resolved without actually pushing the changes before)

davidism:
`look_and_skip_if` will fail if there's more than two optional attributes, since it would be possible to arrange a combination where the first attribute we look for comes last in the list of tokens.

A possible solution is a "look and skip while" method instead, which will keep looking until there are no more of the given type of token. In this case, looking through all names for one that matches the name we're currently looking for, accumulating other names then pushing them all back.

amy-lei:
> `look_and_skip_if` will fail if there's more than two optional attributes, since it would be possible to arrange a combination where the first attribute we look for comes last in the list of tokens.

Is this for other nodes in general? Hmm, yeah `look_and_skip_if` was only designed with Block nodes in mind since they had just 2 attributes.

> A possible solution is a "look and skip while" method instead, which will keep looking until there are no more of the given type of token. In this case, looking through all names for one that matches the name we're currently looking for, accumulating other names then pushing them all back.

This sounds like it'll be more useful. I can give it a try

> `Token` has a `test_any` function, did we already consider using that?

Yeah, one of the earlier iterations tried this. I think the issue I ran into was preventing duplicates like `{% block name scoped scoped %}`, and having to do something like `setattr(node, token.type, True)` since `test_any` won't know which value resulted in the match. 

Do you think it would be better to play around with `test_any` more before trying a `look_and_skip_while`?

davidism:
I'll take a look at this, I need to experiment to see what feels right.

davidism:
Was looking through the docs, and found another case of a tag having two optional keywords. `include` can have `ignore missing` and `with context`, and they must appear in that order if they're both used. https://jinja.palletsprojects.com/en/2.11.x/templates/#include Looks like whoever implemented that didn't want to deal with arbitrary order either, especially for phrases instead of words. :slightly_smiling_face:

I think for now the easiest thing to do is just document that `required` comes last. We can always relax that requirement later if it starts coming up a lot, but I don't want figuring that out to block merging this.

amy-lei:
> Was looking through the docs, and found another case of a tag having two optional keywords. `include` can have `ignore missing` and `with context`, and they must appear in that order if they're both used. https://jinja.palletsprojects.com/en/2.11.x/templates/#include Looks like whoever implemented that didn't want to deal with arbitrary order either, especially for phrases instead of words. 🙂
> 
> I think for now the easiest thing to do is just document that `required` comes last. We can always relax that requirement later if it starts coming up a lot, but I don't want figuring that out to block merging this.

Awesome! Didn't notice `include` was a similar case. This really does simplify the problem :smile: . I'll update the docs and switch them to `skip_if`'s.

PR #1235, opened 2020-06-13
Include instructions for installing docs prerequisites

### LINKED ISSUE #1234: First-time setup instructions missing some requirements
The issue tracker is a tool to address bugs in Jinja itself.
Please use the #pocoo IRC channel on freenode or Stack Overflow for general
questions about using Jinja or issues not related to Jinja.

If you'd like to report a bug in Jinja, fill out the template below and provide
any extra information that may be useful / related to your problem.
Ideally, you create an [MCVE](http://stackoverflow.com/help/mcve) reproducing
the problem before opening an issue to ensure it's not caused by something in
your code.

---

## Expected Behavior
Following the instructions in `CONTRIBUTING.rst` should result in success.

## Actual Behavior
I am able to install requirements using `pip install -e . -r requirements/dev.txt` and `pre-commit install` successfully and run `pytest` as expected. However, when getting to the step "Building the docs" results in errors.

## Template Code
N/A

## Full Traceback
N/A

## Your Environment
* Python version: 3.8.2
* Jinja version: HEAD

The fix is to run `pip install -r requirements/docs.txt`. I will create a PR with this update to the documentation.

sixtyfathoms:
Fixes #1234

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -203,6 +203,12 @@ Read more about `coverage <https://coverage.readthedocs.io>`__.
 Building the docs
 ~~~~~~~~~~~~~~~~~
 
+Install the documentation dependencies if you have not already done so.
+
+.. code-block:: text
+
+    $ pip install -r requirements/docs.txt
+
 Build the docs in the ``docs`` directory using Sphinx.
 
 .. code-block:: text


davidism:
I ended up removing this text again. The issue is that I hadn't been including the docs requirements in the dev requirements. I originally did this because Sphinx depends on Jinja which depends on Sphinx, and pip's resolver was ambiguous about installing both in one command. Pip's *new* resolver is not ambiguous, but chooses to prefer the requirements pin, so I ended up changing it to `pip install -r requirements/dev.txt && pip install -e .`.

PR #1236, opened 2020-06-13
Update wording on dictionary default ordering

### LINKED ISSUE #1184: Update documentation to state that python dicts preserve insertion order
Since Python 3.7 `dicts` must preserve insertion order in all implementations (in Python 3.6 it was only an implementation detail in CPython). Thus I think it would make sense to state this in the [loop section](https://jinja.palletsprojects.com/en/master/templates/#for) of the docs which currently states the following:
> Note, however, that *Python dicts are not ordered*; so you might want to either pass a sorted `list` of `tuple`s [...]

I think it would be helpful to state that since 3.7 these measures are not necessarily needed anymore.

This also works towards https://github.com/pallets/meta/issues/26

sixtyfathoms:
Fixes #1184

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -688,9 +688,17 @@ iterate over containers like `dict`::
     {% endfor %}
     </dl>
 
-Note, however, that **Python dicts are not ordered**; so you might want to
-either pass a sorted ``list`` of ``tuple`` s -- or a
-``collections.OrderedDict`` -- to the template, or use the `dictsort` filter.
+Python dicts may not be in the order you want to display them in. If
+order matters, use the ``|dictsort`` filter.
+
+.. code-block:: jinja
+
+    <dl>
+    {% for key, value in my_dict | dictsort %}
+        <dt>{{ key|e }}</dt>
+        <dd>{{ value|e }}</dd>
+    {% endfor %}
+    </dl>
 
 Inside of a for-loop block, you can access some special variables:
 
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -307,9 +307,8 @@ def do_dictsort(
     by: 'te.Literal["key", "value"]' = "key",
     reverse: bool = False,
 ) -> "t.List[t.Tuple[K, V]]":
-    """Sort a dict and yield (key, value) pairs. Because python dicts are
-    unsorted you may want to use this function to order them by either
-    key or value:
+    """Sort a dict and yield (key, value) pairs. Python dicts may not
+    be in the order you want to display them in, so sort them first.
 
     .. sourcecode:: jinja
 


docs/templates.rst
```diff
@@ -640,9 +640,12 @@ iterate over containers like `dict`::
     {% endfor %}
     </dl>
 
-Note, however, that **Python dicts are not ordered**; so you might want to
-either pass a sorted ``list`` of ``tuple`` s -- or a
-``collections.OrderedDict`` -- to the template, or use the `dictsort` filter.
+Note, however, that **Python dicts preserve insertion order**, which may not
```
septatrix, line 643: May be plausible to remove the _however_ and already say at the start, that this is only guaranteed since 3.7

docs/templates.rst
```diff
@@ -640,9 +640,12 @@ iterate over containers like `dict`::
...
+be the order you expect. If order matters, you can sort data.items() or use
+the dictsort filter. (In Python versions prior to 3.7, ``dict`` objects are
+unsorted, so you might want to either pass a sorted ``list`` of ``tuple`` s
+-- or a ``collections.OrderedDict`` -- to the template, or use the `dictsort`
+filter.)
```
septatrix, line 648: I actually like splitting this thematically into sorting before passing the dict to the template or sorting it inside the template using jinja filters like you did in the second sentence. It just feels like these could be fused into one with a bit rewording. (You could still pass a sorted list of tuples and you could also sort the items before 3.7)

jeremysprofile: You may be right that the wording could be improved, but I think in either case this is still better than the existing documentation, which is still displaying misleading information 5 months later. I guess this is just a polite reminder that this exists and could use some love.

sixtyfathoms:
Oops, fell through the cracks... I will follow up soon.

seawolf42:
I've updated the wording and also added some basic samples.

sixtyfathoms:
I've updated the wording to be cleaner based on the comments above, and I also added some sample code.

docs/templates.rst
```diff
@@ -640,12 +640,30 @@ iterate over containers like `dict`::
...
+sorted ``list`` of ``tuple`` s -- or a ``collections.OrderedDict`` -- to the
+template.
+
+For example::
+
+    # using `sorted`:
+    <dl>
+    {% for key, value in sorted(my_dict.items()) %}
+        <dt>{{ key|e }}</dt>
+        <dd>{{ value|e }}</dd>
```
septatrix, line 656: I would leave out the `|e` as it is not strictly necessary. Otherwise I like the new wording a lot!

docs/templates.rst
```diff
@@ -640,12 +640,30 @@ iterate over containers like `dict`::
...
+    {% endfor %}
+    </dl>
+
+    # using `dictsort`:
```
septatrix, line 660: Maybe use jinja comment syntax here? `{# using dictsort #}`

docs/templates.rst
sixtyfathoms: I left the `|e` in as many other examples in the document also use that; in particular, the example immediately before it does. I would suggest a separate PR to make a pass through the docs for that pattern. Having said that, I'm totally happy to make that change here now if you want.

docs/templates.rst
sixtyfathoms: Done.

docs/templates.rst
septatrix: I like consistency so keep it as is :D

septatrix (approved):

PR #1240, opened 2020-06-17
Allow +%} to ignore trim_blocks

### LINKED ISSUE #1036: Allow +%} to ignore trim_blocks like {%+ do for lstrip_blocks
## Expected Behavior
When trim_block is enabled, if a block ends with +%} this behaviour should be ignored.

## Actual Behavior
As reported in previous issues, defining a block this way results in an error and to force new lines right after a block it's required to put two line breaks.

## Template Code
When trim_blocks is enabled
```jinja
{% if something %}X{% endif +%}
more things
```
Should result in:
```
X
more things
```

## Your Environment
* Python version: any
* Jinja version: 2.10.1

amy-lei:
Resolves #1036 

**Summary**:
- When `trim_block` is enabled and a block ends with `+%}` or `+#}`, the behavior will be ignored.
  - Note: does *not* apply to raw start tags (i.e. `{% raw +%}` will raise an error) because they don't process the inner text
  - `+%}` is still allowed when `trim_block` is disabled but exhibits NOP behavior  
  - So following the example from the original issue, when `trim_block` is enabled:
```jinja
{% if something %}X{% endif +%}
more things
```
results in
```jinja
X
more things
```

**Tests**:
- Added tests to `test_lexnparse.py`
  - Specifically added a class for testing trim called `TestTrimBlocks`
  - Tests majority of the same cases as `TrimLstripBlocks`

**Files affected**:
- `lexer.py`
- `test_lexnparse.py`

### DIFF
--- a/src/jinja2/lexer.py
+++ b/src/jinja2/lexer.py
@@ -509,8 +509,8 @@ class Lexer:
             TOKEN_COMMENT_BEGIN: [
                 (
                     c(
-                        fr"(.*?)((?:\-{comment_end_re}\s*"
-                        fr"|{comment_end_re}){block_suffix_re})"
+                        fr"(.*?)((?:\+{comment_end_re}|\-{comment_end_re}\s*"
+                        fr"|{comment_end_re}{block_suffix_re}))"
                     ),
                     (TOKEN_COMMENT, TOKEN_COMMENT_END),
                     "#pop",
@@ -520,7 +520,10 @@ class Lexer:
             # blocks
             TOKEN_BLOCK_BEGIN: [
                 (
-                    c(fr"(?:\-{block_end_re}\s*|{block_end_re}){block_suffix_re}"),
+                    c(
+                        fr"(?:\+{block_end_re}|\-{block_end_re}\s*"
+                        fr"|{block_end_re}{block_suffix_re})"
+                    ),
                     TOKEN_BLOCK_END,
                     "#pop",
                 ),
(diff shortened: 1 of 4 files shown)


src/jinja2/lexer.py
```diff
@@ -509,7 +509,7 @@ def c(x):
             TOKEN_COMMENT_BEGIN: [
                 (
                     c(
-                        fr"(.*?)((?:\-{comment_end_re}\s*"
+                        fr"(.*?)(\+{comment_end_re}|(?:\-{comment_end_re}\s*"
```
septatrix, line 512: This regex is inconsistent with the additions in other places - I do not know the lexer though is this on purpose?

amy-lei: Are you referring to how this one was not wrapped in a non-capturing group? 
I noticed the original `((?:\-{comment_end_re}` had an outermost capture group unlike the others so I kept it that way for the addition.

davidism: Keeping the outer group is good, but I think they mean that previously `comment_end_re` was never captured, now it would sometimes be captured.

septatrix: Yes, I would have expected
```diff
- fr"(.*?)((?:\-{comment_end_re}\s*"
+ fr"(.*?)(?:\+{comment_end_re}|(?:\-{comment_end_re}\s*"
```
though on a second look at the parentheses It may even be more fitting to use
```diff
- fr"(.*?)((?:\-{comment_end_re}\s*"
+ fr"(.*?)((?:\+{comment_end_re}|\-{comment_end_re}\s*"
```

amy-lei: Ohh I see. Thank you, will make the changes! From your second look @septatrix , this means the parentheses for the other regex additions can also be simplified right?

septatrix: I would assume so, yes - though you might have moved them too far back by one group (- again I did not run the test so this may also work ;D )

davidism: I see the inconsistency here between this block and the others. Even before the change this one captured the suffix, while the others ignored it. The change made it ignored as well. If the tests are still passing, I'd stick with that for consistency.

amy-lei: The tests are still passing. So stick with the changes and have them all ignore the the suffix?

davidism: Yeah, that makes sense to me.

PR #1241, opened 2020-06-22
Fix bug with imported macros and template globals

### LINKED ISSUE #688: Including a macro does not allow access to template globals.
## Expected Behavior
When a template includes a macro and uses it the template globals should be accessible to templates included by the macro.

## Actual Behavior
When including a macro (that includes another template) the nested template does not have access to the template `.globals`. If you force the variable into the `template.globals` after calling `get_template` then it works.

This does not seem to be an issue when the macro is defined and used in the main template, only when the macro is included into the template.

## Example Code
```py
from jinja2 import DictLoader, Environment

loader = DictLoader({
    'test': """
        {% import 'macro' as macros %}
        {{ macros.testing() }}""",
    'macro': """
        {% macro testing() %}
            macro reference to global: {{foo}}
            {% include 'nested' %}
        {% endmacro %}""",
    'nested': 'nested reference to global: {{foo}}'
})
env = Environment(loader=loader, cache_size=0)

# When the template is provided globals and includes a macro
# the global variable is not available to the macro.
template = env.get_template('test', globals={'foo': 'bar'})
# This fails and doesn't show the global foo value:
# macro reference to global:
# nested reference to global:
print template.render()

# If you set the global value after getting the template then the macro
# is able to reference the global variable.
template = env.get_template('test')
template.globals['foo'] = 'bar'
# This works, but I think it is a hack or isn't recommended.
# macro reference to global: bar
# nested reference to global: bar
print template.render()
```

## Your Environment
* Python version: 2.7.10
* Jinja version: 2.9.5

amy-lei:
Fixes #688 
- Added parameter to `_get_default_module` to take in the current context and make a new module if template globals are found
- Added `self.globals` to the Context class to differentiate between template variables from render variables.

### DIFF
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -1120,7 +1120,24 @@ class Template:
         )
 
     @internalcode
-    def _get_default_module(self):
+    def _get_default_module(self, ctx=None):
+        """If a context is passed in, this means that the template was
+        imported.  Imported templates have access to the current template's
+        globals by default, but they can only be accessed via the context
+        during runtime.
+
+        If there are new globals, we need to create a new
+        module because the cached module is already rendered and will not have
+        access to globals from the current context.  This new module is not
+        cached as :attr:`_module` because the template can be imported elsewhere,
+        and it should have access to only the current template's globals.
+        """
+        if ctx is not None:
+            globals = {
+                key: ctx.parent[key] for key in ctx.globals_keys - self.globals.keys()
+            }
+            if globals:
+                return self.make_module(globals)
         if self._module is not None:
             return self._module
         self._module = rv = self.make_module()
--- a/src/jinja2/runtime.py
+++ b/src/jinja2/runtime.py
@@ -97,7 +97,9 @@ def new_context(
         for key, value in locals.items():
             if value is not missing:
                 parent[key] = value
-    return environment.context_class(environment, parent, template_name, blocks)
+    return environment.context_class(
+        environment, parent, template_name, blocks, globals=globals
+    )
 
 
 class TemplateReference:
@@ -179,13 +181,14 @@ class Context(metaclass=ContextMeta):
     _legacy_resolve_mode = False
     _fast_resolve_mode = False
 
-    def __init__(self, environment, parent, name, blocks):
+    def __init__(self, environment, parent, name, blocks, globals=None):
         self.parent = parent
         self.vars = {}
         self.environment = environment
         self.eval_ctx = EvalContext(self.environment, name)
         self.exported_vars = set()
         self.name = name
+        self.globals_keys = set() if globals is None else set(globals)
 
         # create the initial mapping of blocks.  Whenever template inheritance
         # takes place the runtime will update this mapping with the new blocks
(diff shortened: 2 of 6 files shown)


src/jinja2/environment.py
```diff
@@ -1120,7 +1120,13 @@ def make_module_async(self, vars=None, shared=False, locals=None):
         )
 
     @internalcode
-    def _get_default_module(self):
+    def _get_default_module(self, ctx=None):
+        if ctx is not None:
```
davidism, line 1124: It would be helpful to have a comment here describing why this happens. "If a context is passed in, that means {explanation}. If there are new globals, we need to create a new moudle because {explanation}."

src/jinja2/environment.py
```diff
@@ -898,7 +898,7 @@ def from_string(self, source, globals=None, template_class=None):
     def make_globals(self, d):
         """Return a dict for the globals."""
         if not d:
-            return self.globals
+            return self.globals.copy()
```
davidism, line 901: I know this enforces what the docs say, but is it necessary for the fix? It might save a bit of space if the reference was reused for a project with tons of templates that didn't change the globals.

src/jinja2/environment.py
davidism, line 1124: Is it intentional that the module is not cached as `self._module` in this case? If so, the comment should explain why.

src/jinja2/runtime.py
```diff
@@ -179,13 +181,14 @@ class Context(metaclass=ContextMeta):
...
 
-    def __init__(self, environment, parent, name, blocks):
+    def __init__(self, environment, parent, name, blocks, globals=None):
         self.parent = parent
         self.vars = {}
         self.environment = environment
         self.eval_ctx = EvalContext(self.environment, name)
         self.exported_vars = set()
         self.name = name
+        self.globals = set() if globals is None else set(globals)
```
davidism, line 191: This should either be the `globals` dict, or should be called something like `globals_keys`.

src/jinja2/environment.py
amy-lei: that's true. removed it 👍

src/jinja2/environment.py
amy-lei: Yep was intentional. Added it to the comments 👍

src/jinja2/environment.py
```diff
@@ -1121,9 +1121,21 @@ def make_module_async(self, vars=None, shared=False, locals=None):
...
+        module because the cached module is already rendered and will not have
+        access to globals from the current context.  This new module is not
+        cached as :attr:`_module` because the template can be imported elsewhere,
+        and it should have access to only the current template's globals.
+        """
         if ctx is not None:
             globals = {
-                key: ctx.parent[key] for key in ctx.globals - set(self.globals.keys())
+                key: ctx.parent[key]
+                for key in ctx.globals_keys - set(self.globals.keys())
```
davidism, line 1138: `dict.keys()` behaves enough like a set that you don't have to wrap it here.

amy-lei: 😮 did not know that!

davidism:
`_get_default_module` is sort of misnamed now, since it will get a default module, or a new module for a new import. :thinking:

Issue #1088, opened 2019-10-24
Can't directly access loop variable in nested block


EpicWink:
Trying to access the loop variable `loop` inside a nest block (even with the `scoped` argument) causes Jinja to report an undefined-error during rendering.

## Template Code
```jinja
Spam
{% for item in ["a", "list", "of", "items"] -%}
{% block spamblock scoped -%}
Foo {{ item }} {% if loop.last %}Bar{% endif %}
{%- endblock spamblock %}
{% endfor -%}
Eggs
```

## Expected Behavior
Ability to access the loop variable in the nested block. The example above should produce:
```
Spam
Foo a 
Foo list 
Foo of 
Foo items Bar
Eggs
```

## Actual Behavior
Jinja reports an undefined-error

### Full Traceback
```pytb
Traceback (most recent call last):
  File "render.py", line 6, in <module>
    rendered = template.render()
  File "~/.local/lib/python3.7/site-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "~/.local/lib/python3.7/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "~/.local/lib/python3.7/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "~/.local/lib/python3.7/site-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "./template.txt", line 3, in top-level template code
    {% block spamblock scoped -%}
  File "./template.txt", line 4, in block "spamblock"
    Foo {{ item }} {% if loop.last %}Bar{% endif %}
  File "~/.local/lib/python3.7/site-packages/jinja2/environment.py", line 430, in getattr
    return getattr(obj, attribute)
jinja2.exceptions.UndefinedError: 'loop' is undefined
```

## Your Environment
* Python version: 3.7.3 [GCC 8.3.0] on linux
* Jinja version: 2.10.1

## Workaround
Simply setting the loop variable to another, and using the other variable inside the nested block (and overrided blocks in inheriting child templates):
```jinja
Spam
{% for item in ["a", "list", "of", "items"] -%}
{% set loop_ = loop -%}
{% block spamblock scoped -%}
Foo {{ item }} {% if loop_.last %}Bar{% endif %}
{%- endblock spamblock %}
{% endfor -%}
Eggs
```

davidism:
[...]

PR #1242, opened 2020-06-23
Fix bugs with contextfunction and loop variables

### LINKED ISSUE #768: Loop variables in custom python contextfunction
Hi,

I am experiencing an unexpected behavior with Jinja and local variables in a for loop when accessing them from a custom Python context function.

## Expected Behavior
I expected for every variable to be available in the context object in a python `@contextfunction`, even if it is defined/updated in a for loop. 

## Actual Behavior
The actual behavior is that only variables defined/updated outside the loop are available/have the expected value.

## Template Code
Here is a simple template code to demonstrate the behavior:

```python
>>> import jinja2
>>> jinja2.__version__
'2.9.4'
>>> @jinja2.contextfunction
... def myContextFunction(ctx):
...     return "myContextFunction: ctx['myI'] = " + str(ctx['myI'])
...
>>> tmplt = """
... {% for i in range(2): %}
... {%- set  myI = i -%}
...     {{ myContextFunction() }}
...     In template myI = {{ myI }}
... {% endfor %}
... """
>>> template = jinja2.Template(tmplt)
>>> template.globals['myContextFunction'] = myContextFunction
>>> template.render()
  File "/arm/tools/python/jinja2_py2.7.8/2.9.4/rhe6-x86_64/lib/jinja2/environment.py", line 1008, in render
  File "/arm/tools/python/jinja2_py2.7.8/2.9.4/rhe6-x86_64/lib/jinja2/environment.py", line 780, in handle_exception
  File "<template>", line 4, in top-level template code
  File "<stdin>", line 3, in myContextFunction
  File "/arm/tools/python/jinja2_py2.7.8/2.9.4/rhe6-x86_64/lib/jinja2/runtime.py", line 253, in __getitem__
    raise KeyError(key)
KeyError: 'myI'
>>> context = {'myI' : 0}
>>> print template.render(**context)

myContextFunction: ctx['myI'] = 0
        In template myI = 0
myContextFunction: ctx['myI'] = 0
        In template myI = 1
```

My first goal was to access the `i` counter, or also the `loop.index` from my custom `contextfunction`. Also those variable, I expected to be defined in context, where not available.

## Your Environment
* Python version: Python 2.7.8
* Jinja version: '2.9.4'

amy-lei:
Fixes #768  and #1088 (closed as dupe but turned out to be slightly different)

Summary:
- Scoped blocks can now access `loop` and its special variables
- `contextfunction` can access variables set inside of loops and blocks
  - Achieved by adding an attribute to `Frame`, tracking assignments made in loops or blocks, and passing it as `extra_kwargs`

Files Changed:
- `compiler.py`
- `runtime.py`
- `test_regression.py`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,10 @@ Unreleased
     :issue:`522, 827, 1172`, :pr:`1195`
 -   Filters that get attributes, such as ``map`` and ``groupby``, can
     use a false or empty value as a default. :issue:`1331`
+-   Fix a bug that prevented variables set in blocks or loops from
+    being accessed in custom context functions. :issue:`768`
+-   Fix a bug that caused scoped blocks from accessing special loop
+    variables. :issue:`1088`
 
 
 Version 2.11.3
--- a/src/jinja2/compiler.py
+++ b/src/jinja2/compiler.py
@@ -1008,9 +1008,12 @@ class CodeGenerator(NodeVisitor):
 
         # try to figure out if we have an extended loop.  An extended loop
         # is necessary if the loop is in recursive mode if the special loop
-        # variable is accessed in the body.
-        extended_loop = node.recursive or "loop" in find_undeclared(
-            node.iter_child_nodes(only=("body",)), ("loop",)
+        # variable is accessed in the body if the body is a scoped block.
+        extended_loop = (
+            node.recursive
+            or "loop"
+            in find_undeclared(node.iter_child_nodes(only=("body",)), ("loop",))
+            or any(block.scoped for block in node.find_all(nodes.Block))
         )
 
         loop_ref = None
--- a/src/jinja2/compiler.py
+++ b/src/jinja2/compiler.py
@@ -131,6 +131,11 @@ class Frame:
         if parent is not None:
             self.buffer = parent.buffer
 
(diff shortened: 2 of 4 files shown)


amy-lei:
I marked this as a draft for now to get feedback on the approach I took—especially for solving the `contextfunction` bug.

The cause was that variables set inside of loops do not actually alter the context, so the contextfunction had no way of accessing those variables unless they're passed in as args/kwargs. The idea was therefore to store variables from Set blocks along the way (if inside a loop), pass it in as a kwarg, and use it to create a derived context for the func.

Block tags also face the same problem, although an issue hasn't been opened for it yet. So if this approach works, I can mirror it for blocks as well!

davidism:
Are blocks already solved by the `scoped` flag? https://jinja.palletsprojects.com/en/2.11.x/templates/#block-nesting-and-scope

amy-lei:
Having both a `loop_frame` and `block_frame` attribute can be a bit redundant, especially since it's duplicate work until you choose which context to copy. Perhaps it can be reduced something like `frame_type`.

However, at the moment, there doesn't seem to be other block tags with this context access issue. So when one does come up, we can refactor/address this then.

PR #1244, opened 2020-06-23
Fix bug with cached templates not registering new globals

### LINKED ISSUE #295: Template cache, {% extends %} and globals - design issue?
Say I have this template with an extends call:

foo.html:

```
{% extends 'base.html' %}
```

I load it:

```
jinja2_env.get_template('foo.html', globals=my_globals}
```

This causes two calls to `Environment._load_template`: First, for `foo.html`, the for `base.html`.

`Environment._load_template` caches both templates.

The crucial point is that for the second call, the "globals" parameter to `Environment._load_template` is missing `my_globals`.

If I now attempt to load the base template directly:

```
jinja2_env.get_template('base.html', globals=my_globals}
```

The cache will be used, and my globals will not be available.

Using 2.7.2

amy-lei:
Fixes #295
Summary:
- Issue affected templates rendered and cached due to `extend` and `include` tags 
  -  Now updates Template object with new globals, if any, when retrieving from cache 

Files affected:
- `environment.py`
- `test_regression.py`

### DIFF
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -811,6 +811,16 @@ class Environment:
             if template is not None and (
                 not self.auto_reload or template.is_up_to_date
             ):
+                # update globals if changed
+                new_globals = globals.items() - template.globals.items()
+                if new_globals:
+                    # it is possible for the template and environment to share
+                    # a globals object, in which case, a new copy should be
+                    # made to avoid affecting other templates
+                    if template.globals is self.globals:
+                        template.globals = dict(template.globals, **globals)
+                    else:
+                        template.globals.update(dict(new_globals))
                 return template
         template = self.loader.load(self, name, globals)
         if self.cache is not None:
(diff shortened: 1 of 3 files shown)


src/jinja2/environment.py
```diff
@@ -898,7 +902,7 @@ def from_string(self, source, globals=None, template_class=None):
     def make_globals(self, d):
         """Return a dict for the globals."""
         if not d:
-            return self.globals
+            return self.globals.copy()
```
amy-lei, line 905: I remember from a previous PR that we decided this was not necessary since it took up extra space. But when fixing this bug, I ran into unintended side effects as a result.

tests/test_regression.py
```diff
@@ -611,3 +611,37 @@ def test_markup_and_chainable_undefined(self):
...
+        assert tmpl.render() == "bar"
+
+        tmpl = env.get_template("base", globals={"foo": 42})
+        assert tmpl.render() == "42"
+
+        tmpl = env.get_template("main")
+        assert tmpl.render() == "bar"
+
+        tmpl = env.get_template("base")
+        assert tmpl.render() == "42"
```
amy-lei, line 647: The test cases I was referring to in the comment above

davidism (approved):

PR #1248, opened 2020-06-26
allow optional use of a filter based on its existence

### LINKED ISSUE #842: optionally use a filter based on its existence
I propose a feature, where within a template, you can test for the existence of a filter so that you can then use that filter if it exists, and not if it doesn't.

For example,
```jinja
{% if filter is callable %}
  {{ var|filter }}
{% else %}
  {{ var }} {# no use of 'filter' #}
{% endif %}
```

Currently, if the filter does not exist, Jinja would throw a `jinja2.exceptions.TemplateAssertionError: no filter named 'filter'` because of the line `{{ var|filter }}`. The desired output is what is rendered in the `else`. You can test if the filter is callable, but you cannot then optionally use it. At a high level this is similar to how one might commonly use a var, if it exists.

```jinja
{% if myvar is defined %}
  {{ myvar }}
{% else %}
  No myvar {# no use of 'myvar' #}
{% endif %}
```

**Example Use Case:**

I would like to be able to write Lektor themes such that they can take advantage of filters provided by various Lektor plugins if they are installed. I do not want to require those plugins, and thus also their provided filters. I'd want a Lektor theme to flexibly handle either situation, as per the users' desire.

More generally, Jinja templates in any Flask or Django project could be written more flexibly to handle the optional presence filters. Adding custom or 3rd party filters is a common thing to do. Passing templates around between projects would be easier if they didn't force implementation of custom filters.

**Implementation:**

The desired effect could be achieved more than one way. For example, at least

1. not raising an exception when evaluating `{{ var|filter }}` and instead returning (basically anything) would allow the above snippet to be rendered. 

2. change the way the entire `if` is evaluated, such that the expressions contained in the contents of the "true" section are evaluated only if the test passes. Evaluation would resume upon reaching an `else` or `endif`.

My intuition was that the second option was already happening, but it is not. Though the contents of the passing section of the `if` never get to the final rendered product, the expressions are still evaluated anyway. Otherwise no exception would be rendered. If 2. happened, that could mean much nonsense could be included in between the `if` and the `else`, and not trip exceptions. That sounds like a change that's worth some careful thought, but I can't immediately think of a problem with it. I am not intimately familiar with Jinja's internals. Exceptions would still be raised if the test ever actually passes. 
Skipping that section may also mean performance improvements, which is a nice side effect.

amy-lei:
Resolves #842 

- Undefined filters nested in if/elif/else statements and conditional expressions will cause an error only if executed during runtime

Files changed:
- `nodes.py`
- `compiler.py`
- `runtime.py`

### DIFF
--- a/src/jinja2/compiler.py
+++ b/src/jinja2/compiler.py
@@ -136,6 +136,11 @@ class Frame:
         self.loop_frame = False
         self.block_frame = False
 
+        # track whether the frame is being used in an if-statement or conditional
+        # expression as it determines which errors should be raised during runtime
+        # or compile time.
+        self.soft_frame = False
+
     def copy(self):
         """Create a copy of the current one."""
         rv = object.__new__(self.__class__)
@@ -154,10 +159,12 @@ class Frame:
         standalone thing as it shares the resources with the frame it
         was created of, but it's not a rootlevel frame any longer.
 
-        This is only used to implement if-statements.
+        This is only used to implement if-statements and conditional
+        expressions.
         """
         rv = self.copy()
         rv.rootlevel = False
+        rv.soft_frame = True
         return rv
 
     __copy__ = copy
@@ -442,7 +449,22 @@ class CodeGenerator(NodeVisitor):
             for name in getattr(visitor, dependency):
                 if name not in mapping:
                     mapping[name] = self.temporary_identifier()
+                # add check during runtime that dependencies used inside of executed
+                # blocks are defined, as this step may be skipped during compile time
+                self.writeline("try:")
+                self.indent()
                 self.writeline(f"{mapping[name]} = environment.{dependency}[{name!r}]")
+                self.outdent()
+                self.writeline("except KeyError:")
+                self.indent()
+                self.writeline("@internalcode")
+                self.writeline(f"def {mapping[name]}(*unused):")
+                self.indent()
+                self.writeline(
+                    f'raise TemplateRuntimeError("no filter named {name!r} found")'
+                )
+                self.outdent()
+                self.outdent()
 
     def enter_frame(self, frame):
         undefs = []
@@ -1641,7 +1663,7 @@ class CodeGenerator(NodeVisitor):
             self.write("await auto_await(")
         self.write(self.filters[node.name] + "(")
         func = self.environment.filters.get(node.name)
-        if func is None:
+        if func is None and not frame.soft_frame:
             self.fail(f"no filter named {node.name!r}", node.lineno)
         if getattr(func, "contextfilter", False) is True:
             self.write("context, ")
@@ -1679,6 +1701,8 @@ class CodeGenerator(NodeVisitor):
 
     @optimizeconst
     def visit_CondExpr(self, node, frame):
+        frame = frame.soft()
+
         def write_expr2():
             if node.expr2 is not None:
                 return self.visit(node.expr2, frame)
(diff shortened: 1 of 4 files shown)


src/jinja2/runtime.py
```diff
@@ -152,6 +152,8 @@ def resolve_or_missing(context, key, missing=missing):
         return context.vars[key]
     if key in context.parent:
         return context.parent[key]
+    if key in context.environment.filters:
```
amy-lei, line 155: Wasn't too sure about this since this will allow users to use filters as variables, which was previously not a thing. However, i believe this is similar to how python deals with built-ins and overriding them.

src/jinja2/compiler.py
```diff
@@ -1167,18 +1193,38 @@ def visit_If(self, node, frame):
         self.visit(node.test, if_frame)
         self.write(":")
         self.indent()
+
+        # check during runtime that filters used inside of the executed
+        # block are defined, as this step was skipped during compile time
+        def check_filters(node):
+            for filter in node.find_all(nodes.Filter, exclude=nodes.If):
+                self.writeline(f"if {self.filters[filter.name]} is None:")
```
davidism, line 1201: Should this use `get`?

amy-lei: i saw  `__getitem__` used here, but it could be `get` in case https://github.com/pallets/jinja/blob/a127d9f452da9f6bfd2f7295cec907bc8399259f/src/jinja2/compiler.py#L1685-L1688

ThiefMaster:
Are we covering cases like `{{ 'foo'|nosuchfilter if false else 'bar' }}` and `{{ 'foo' if true else 'bar'|nosuchfilter }}` - neither should raise (since the filter is not actually used, but based on the code it looks like it may raise..

amy-lei:
> Are we covering cases like `{{ 'foo'|nosuchfilter if false else 'bar' }}` and `{{ 'foo' if true else 'bar'|nosuchfilter }}` - neither should raise (since the filter is not actually used, but based on the code it looks like it may raise..

Right, I forgot to account for those but probably should (and will) too.

ThiefMaster:
I have the feeling any block-level check is likely to have problematic edge cases - can't we just let it fail when actually accessing the filter?

amy-lei:
> I have the feeling any block-level check is likely to have problematic edge cases - can't we just let it fail when actually accessing the filter?

Hmm that's true. The error won't be as helpful though: `TypeError: 'NoneType' object is not callable`

ThiefMaster:
Good point, that error is awful and filters that are sometimes but not always undefined are very much an edge case for which we probably shouldn't sacrifice error verboseness with other, more common, cases

ThiefMaster:
Could we compile filters to something like this at the beginning of the template?

```python
try:
    filter_tmp_0 = self.filters['filter_name']
except KeyError:
    @internalcode
    def filter_tmp_0(*unused):
        raise TemplateRuntimeError("no filter named 'filter_name'")
```

Then the code where the  filter is actually used just uses `filter_tmp_0` which will either call the filter or raise an exception - without adding extra runtime overhead every time the filter is actually invoked.

amy-lei:
Ooh that's a good idea! I moved it to `pull_dependencies` in the new commit. Let me know if I misunderstood. Will address conditional expressions next.

davidism (approved):

davidism:
I'm working on a separate PR to create an `if 'name' is filter` test.

PR #1247, opened 2020-06-26
Setup mypy

teymour-aldridge:
Fixes #1246 
* Add missing type hints (these are intended as an initial set of type hints, to be added upon and improved later)
* Setup MyPy to run as a Github Action

PR #1255, opened 2020-07-09
add type annotations to filters

jessecooper:
I see in https://github.com/pallets/jinja/pull/1247 the project will be adding mypy support. 
I found it useful to have the `__annotations__` attributes on filters in an effort to organize the filters. 
Please let me know if any of the typing is incorrect.

### DIFF
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -8,7 +8,9 @@ from itertools import chain
 from itertools import groupby
 from typing import Any
 from typing import Iterable
+from typing import Optional
 from typing import Sequence
+from typing import TypeVar
 from typing import Union
 
 from markupsafe import escape
@@ -24,6 +26,8 @@ from .utils import urlize
 
 _word_re = re.compile(r"\w+")
 _word_beginning_split_re = re.compile(r"([-\s({\[<]+)")
+
+T = TypeVar('T')
 
 
 def contextfilter(f):
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -429,7 +429,7 @@ def do_max(environment, value: Sequence, case_sensitive=False, attribute=None) -
     return _min_or_max(environment, value, max, case_sensitive, attribute)
 
 
-def do_default(value: Any, default_value: Any = "", boolean: bool = False):
+def do_default(value: Any, default_value: Any = "", boolean: bool = False) -> Any:
     """If the value is undefined it will return the passed default value,
     otherwise the value of the variable:
 
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -419,7 +419,7 @@ def do_min(
 
 
 @environmentfilter
-def do_max(environment, value: Sequence, case_sensitive=False, attribute=None) -> int:
+def do_max(environment, value: Sequence[T], case_sensitive=False, attribute=None) -> T:
     """Return the largest item from the sequence.
 
     .. sourcecode:: jinja
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -845,7 +845,7 @@ def do_striptags(value):
     return Markup(str(value)).striptags()
 
 
-def do_slice(value: Iterable, slices: int, fill_with: str = None) -> Iterable:
+def do_slice(value: Iterable[T], slices: int, fill_with: str = None) -> List[List[T]]:
     """Slice an iterator and return a list of lists containing
     those items. Useful if you want to create a div containing
     three ul tags that represent columns:
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -6,6 +6,7 @@ from collections import abc
 from collections import namedtuple
 from itertools import chain
 from itertools import groupby
+from typing import Union, Any, Iterable, Sequence, ItemsView
 
 from markupsafe import escape
 from markupsafe import Markup
@@ -128,7 +129,7 @@ def do_forceescape(value):
     return escape(str(value))
 
 
-def do_urlencode(value):
+def do_urlencode(value: Union[str, Iterable, Sequence]):
     """Quote data for use in a URL path or query using UTF-8.
 
     Basic wrapper around :func:`urllib.parse.quote` when given a
@@ -158,7 +159,7 @@ def do_urlencode(value):
 
 
 @evalcontextfilter
-def do_replace(eval_ctx, s, old, new, count=None):
+def do_replace(eval_ctx, s: str, old: str, new: str, count: int = None):
     """Return a copy of the value with all occurrences of a substring
     replaced with a new one. The first argument is the substring
     that should be replaced, the second is the replacement string.
@@ -188,18 +189,18 @@ def do_replace(eval_ctx, s, old, new, count=None):
     return s.replace(soft_str(old), soft_str(new), count)
 
 
-def do_upper(s):
+def do_upper(s: str) -> str:
     """Convert a value to uppercase."""
     return soft_str(s).upper()
 
 
-def do_lower(s):
+def do_lower(s: str) -> str:
     """Convert a value to lowercase."""
     return soft_str(s).lower()
 
 
 @evalcontextfilter
-def do_xmlattr(_eval_ctx, d, autospace=True):
+def do_xmlattr(_eval_ctx, d: dict, autospace: bool = True):
     """Create an SGML/XML attribute string based on the items in a dict.
     All values that are neither `none` nor `undefined` are automatically
     escaped:
@@ -234,14 +235,14 @@ def do_xmlattr(_eval_ctx, d, autospace=True):
     return rv
 
 
-def do_capitalize(s):
+def do_capitalize(s: str) -> str:
     """Capitalize a value. The first character will be uppercase, all others
     lowercase.
     """
     return soft_str(s).capitalize()
 
 
-def do_title(s):
+def do_title(s: str) -> str:
     """Return a titlecased version of the value. I.e. words will start with
     uppercase letters, all remaining characters are lowercase.
     """
@@ -254,7 +255,7 @@ def do_title(s):
     )
 
 
-def do_dictsort(value, case_sensitive=False, by="key", reverse=False):
+def do_dictsort(value: dict, case_sensitive: bool = False, by: str = "key", reverse: bool = False):
     """Sort a dict and yield (key, value) pairs. Because python dicts are
     unsorted you may want to use this function to order them by either
     key or value:
@@ -292,7 +293,7 @@ def do_dictsort(value, case_sensitive=False, by="key", reverse=False):
 
 
 @environmentfilter
-def do_sort(environment, value, reverse=False, case_sensitive=False, attribute=None):
+def do_sort(environment, value: Iterable, reverse: bool = False, case_sensitive: bool = False, attribute: str = None):
     """Sort an iterable using Python's :func:`sorted`.
 
     .. sourcecode:: jinja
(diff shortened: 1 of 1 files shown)


src/jinja2/filters.py
```diff
@@ -128,7 +132,7 @@ def do_forceescape(value):
     return escape(str(value))
 
 
-def do_urlencode(value):
+def do_urlencode(value: Union[str, Iterable, Sequence]):
```
marchingphoenix, line 135:
```suggestion
def do_urlencode(value: Union[str, Iterable, Sequence]) -> str:
```

src/jinja2/filters.py
```diff
@@ -158,7 +162,7 @@ def do_urlencode(value):
 
 
 @evalcontextfilter
-def do_replace(eval_ctx, s, old, new, count=None):
+def do_replace(eval_ctx, s: str, old: str, new: str, count: int = None):
```
marchingphoenix, line 165:
```suggestion
def do_replace(eval_ctx, s: str, old: str, new: str, count: int = None) -> str:
```

src/jinja2/filters.py
```diff
@@ -254,7 +258,9 @@ def do_title(s):
     )
 
 
-def do_dictsort(value, case_sensitive=False, by="key", reverse=False):
+def do_dictsort(
+    value: dict, case_sensitive: bool = False, by: str = "key", reverse: bool = False
+):
```
marchingphoenix, line 263:
```suggestion
) -> dict:
```

src/jinja2/filters.py
```diff
@@ -292,7 +298,13 @@ def sort_func(item):
...
 
 @environmentfilter
-def do_sort(environment, value, reverse=False, case_sensitive=False, attribute=None):
+def do_sort(
+    environment,
+    value: Iterable,
+    reverse: bool = False,
+    case_sensitive: bool = False,
+    attribute: str = None,
+):
```
marchingphoenix, line 307:
```suggestion
def do_sort(
    environment,
    value: Iterable,
    reverse: bool = False,
    case_sensitive: bool = False,
    attribute: str = None,
) -> Iterable:
```

src/jinja2/filters.py
```diff
@@ -342,7 +354,9 @@ def do_sort(environment, value, reverse=False, case_sensitive=False, attribute=N
 
 
 @environmentfilter
-def do_unique(environment, value, case_sensitive=False, attribute=None):
+def do_unique(
+    environment, value: Iterable, case_sensitive: bool = False, attribute: str = None
+):
```
marchingphoenix, line 359: Missing output type

src/jinja2/filters.py
```diff
@@ -188,18 +192,18 @@ def do_replace(eval_ctx, s, old, new, count=None):
...
 
-def do_lower(s):
+def do_lower(s: str) -> str:
     """Convert a value to lowercase."""
     return soft_str(s).lower()
 
 
 @evalcontextfilter
-def do_xmlattr(_eval_ctx, d, autospace=True):
+def do_xmlattr(_eval_ctx, d: dict, autospace: bool = True):
```
marchingphoenix, line 206: Missing Output type

src/jinja2/filters.py
```diff
@@ -384,7 +398,9 @@ def _min_or_max(environment, value, func, case_sensitive, attribute):
 
 
 @environmentfilter
-def do_min(environment, value, case_sensitive=False, attribute=None):
+def do_min(
+    environment, value: Sequence, case_sensitive: bool = False, attribute: str = None
+):
```
marchingphoenix, line 403: missing output type

src/jinja2/filters.py
```diff
@@ -413,7 +429,7 @@ def do_max(environment, value, case_sensitive=False, attribute=None):
     return _min_or_max(environment, value, max, case_sensitive, attribute)
 
 
-def do_default(value, default_value="", boolean=False):
+def do_default(value: Any, default_value: Any = "", boolean: bool = False):
```
marchingphoenix, line 432:
```suggestion
def do_default(value: Any, default_value: Any = "", boolean: bool = False) -> Any:
```

src/jinja2/filters.py
```diff
@@ -562,14 +578,14 @@ def do_filesizeformat(value, binary=False):
...
+def do_pprint(value) -> str:
     """Pretty print a variable. Useful for debugging."""
     return pformat(value)
 
 
 @evalcontextfilter
 def do_urlize(
-    eval_ctx, value, trim_url_limit=None, nofollow=False, target=None, rel=None
+    eval_ctx, value: str, trim_url_limit=None, nofollow=False, target=None, rel=None
 ):
```
marchingphoenix, line 589:
```suggestion
    eval_ctx, value: str, trim_url_limit=None, nofollow=False, target=None, rel=None
) -> str:
```

src/jinja2/filters.py
```diff
@@ -1008,7 +1033,7 @@ def do_mark_unsafe(value):
     return str(value)
 
 
-def do_reverse(value):
+def do_reverse(value: Union[str, Iterable]) -> Union[str, Iterable]:
```
marchingphoenix, line 1036: Missing output type

src/jinja2/filters.py
```diff
@@ -765,7 +788,7 @@ def do_int(value, default=0, base=10):
             return default
 
 
-def do_float(value, default=0.0):
+def do_float(value: Union[int, float, str], default=0.0) -> float:
```
teymour-aldridge, line 791: I'm not sure if `Any` would be more appropriate here.
```suggestion
def do_float(value: Any, default=0.0) -> float:
```

src/jinja2/filters.py
```diff
@@ -815,7 +838,7 @@ def do_striptags(value):
     return Markup(str(value)).striptags()
 
 
-def do_slice(value, slices, fill_with=None):
+def do_slice(value: Iterable, slices: int, fill_with: str = None) -> Iterable:
```
teymour-aldridge, line 841:
```suggestion
def do_slice(value: Iterable[T], slices: int, fill_with: str = None) -> List[List[T]]:
```

src/jinja2/filters.py
```diff
@@ -851,7 +874,7 @@ def do_slice(value, slices, fill_with=None):
         yield tmp
 
 
-def do_batch(value, linecount, fill_with=None):
+def do_batch(value: Iterable, linecount: int, fill_with: str = None) -> Iterable:
```
teymour-aldridge, line 877:
```suggestion
def do_batch(value: Iterable[T], linecount: int, fill_with: str = None) -> List[List[T]]:
```

src/jinja2/filters.py
```diff
@@ -927,7 +950,7 @@ def do_round(value, precision=0, method="common"):
 
 
 @environmentfilter
-def do_groupby(environment, value, attribute):
+def do_groupby(environment, value: Sequence, attribute: str) -> list:
```
teymour-aldridge, line 953:
```suggestion
def do_groupby(environment, value: Sequence[T], attribute: str) -> List[T]:
```

src/jinja2/filters.py
jessecooper: I did it in this manner because you would not run a do_float on a list or dict and if we use Any that might suggest such a thing could/would get a valid return statement.

src/jinja2/filters.py
```diff
@@ -506,7 +522,7 @@ def do_first(environment, seq):
 
 
 @environmentfilter
-def do_last(environment, seq):
+def do_last(environment, seq: Sequence) -> Any:
```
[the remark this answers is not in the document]
teymour-aldridge: The advantage as I see it is that using `Sequence[T] -> T` means "given a sequence of some type `T` this function returns a value of the *same type* `T`

src/jinja2/filters.py
teymour-aldridge: Yeah that's true. I think a type annotation for default (`float`) would also help :D

src/jinja2/filters.py
```diff
@@ -524,15 +540,15 @@ def do_last(environment, seq):
 
 
 @contextfilter
-def do_random(context, seq):
+def do_random(context, seq: Sequence) -> Any:
```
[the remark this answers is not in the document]
jessecooper: I guess I am not clear when to use a TypeVar of T that represents an Any generic type vs just using `Any` and/or a group type like Sequence that represents any type of sequence types. If by `Sequence[T]` could we also just say `Sequence[Any]` instead of using generic type T?

src/jinja2/filters.py
jessecooper: In this case I do not think we would want someone using `do_float` on a list or dict so I think Any would be a little to vague.

src/jinja2/filters.py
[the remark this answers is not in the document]
jessecooper: Ok I can see that making a lot more sense than Any.

src/jinja2/filters.py
```diff
@@ -6,6 +6,13 @@
 from collections import namedtuple
 from itertools import chain
 from itertools import groupby
+from typing import Any
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Sequence
+from typing import TypeVar
+from typing import Union
```
septatrix, line 15: Is this the prefered code style? For so many imports it seems a lot more intuitive to put these all on a single line (`from typing import Any, Iterable, List, Optional, Sequence, TypeVar, Union`)

jessecooper: This is the way the pre commits formatted it. Not my preference at all.

src/jinja2/filters.py
```diff
@@ -442,7 +458,7 @@ def do_default(value, default_value="", boolean=False):
 
 
 @evalcontextfilter
-def do_join(eval_ctx, value, d="", attribute=None):
+def do_join(eval_ctx, value: Iterable, d: str = "", attribute: str = None) -> str:
```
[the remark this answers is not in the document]
Dreamsorcerer: If it's just using `str()`, then it should be `Iterable[object]`, which matches the typeshed annotation for str:
https://github.com/python/typeshed/blob/master/stdlib/2and3/builtins.pyi#L453

jessecooper:
@teymour-aldridge Can I please get an update on this?

src/jinja2/filters.py
```diff
@@ -491,13 +514,13 @@ def do_join(eval_ctx, value, d="", attribute=None):
...
 
-def do_center(value, width=80):
+def do_center(value: str, width: int = 80) -> str:
     """Centers the value in a field of a given width."""
     return str(value).center(width)
 
 
 @environmentfilter
-def do_first(environment, seq):
+def do_first(environment, seq: Sequence) -> Any:
```
Dreamsorcerer, line 523: This should use a TypeVar the same of the below 2 functions, right?

septatrix: The type could even be `Iterable`

src/jinja2/filters.py
```diff
@@ -506,7 +529,7 @@ def do_first(environment, seq):
 
 
 @environmentfilter
-def do_last(environment, seq):
+def do_last(environment, seq: Sequence[T]) -> T:
```
septatrix, line 532: This could use `Reversible` to allow more types

septatrix (commented):
This look great so far. However a few annotations are still missing (for parameters, return values and generic classes).

davidism:
Thanks for working on this. After reviewing it, I noticed that there were still a lot of missing annotations, and many could be made more accurate. #1384 adds annotations to all filters (including async) and tests.

PR #1282, opened 2020-08-29
Mention new Discord server invite link

bittner:
The IRC channel #pocoo doesn't seem to be used anymore. When you login there, a friendly notice mentions the "New Home on Discord".

This change updates the related information, so interested parties don't have to do the unnecessary hop to IRC first.


ThiefMaster:
AFAIK we have the vanity URL https://discord.gg/pallets people can use to join. I'd also keep the mention of the IRC channel since we have a relay bot and I'd like to not exclude people who prefer IRC and don't even have Discord. (but we should be consistent among the different pallets repos)

davidism:
I'll change this for now, but I need to copy over the work we did on the Werkzeug templates.

bittner:
I tried to address your comment. Does that now look like you'd want to have it?

ThiefMaster:
I'd reword it to put the discord server first (it's the preferred place for people to go) and just keep a mention of the IRC channel after that.

cc @davidism in case he prefers some text that's the same everywhere

davidism:
Thanks for working on this, but I should really just copy over the new templates from Werkzeug.

bittner:
Alright, I understand. Copied the templates over from Werkzeug. :checkered_flag:

.github/workflows/tests.yaml
```diff
@@ -22,6 +22,7 @@ jobs:
           - {name: 'PyPy', python: pypy3, os: ubuntu-latest, tox: pypy3}
           - {name: Style, python: '3.8', os: ubuntu-latest, tox: style}
           - {name: Docs, python: '3.8', os: ubuntu-latest, tox: docs}
+          - {name: Typing, python: '3.8', os: ubuntu-latest, tox: mypy}
```
davidism, line 25: This shouldn't be here.

bittner: Okay, done. Anything else? Because, really, I just copied the entire folder and did search+replace for "Werkzeug".

PR #1284, opened 2020-09-01
Provide basic support for sorting list of dictionaries

### LINKED ISSUE #1283: Handling list of dictionaries differs between python versions
Python 2.7 supports dictionaries comparison, so sorting of dict list is supported by default. 
On python 3 `__lt__` is not defined for dictionaries, so it makes harder to migrate projects that uses jinja2 to python 3.

For example for template
```
"{{ items|sort }}"

tmpl.render(items=[{'b': 4}, {'a': 3}, {a': 2}, {'b': 1}])
```
leads to `TypeError: '<' not supported between instances of 'dict' and 'dict'"`

Providing even basic support for dictionaries would make migration process a lot easier.

Istota42:
This commit adds support for sorting list of dictionaries.

fixes #1283

Checklist:

### DIFF
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -103,6 +103,11 @@ def make_multi_attrgetter(environment, attribute, postprocess=None):
             for part in attribute_part:
                 item_i = environment.getitem(item_i, part)
 
+            try:
+                item_i < item_i
+            except TypeError:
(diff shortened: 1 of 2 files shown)


davidism:
Closing, see #1283 for discussion.

PR #1288, opened 2020-09-01
Provide support for grouping iterable with 'None' value in keys

### LINKED ISSUE #1287: Grouping iterable with 'None' in keys differs between python versions
Python 2.7 allows to compare a value with None. It's not possible in python 3 and it makes harder to migrate projects (that uses jinja2) to python3.

Example:
```
{{ [{'foo': 1, 'bar': 2},
            {'foo': 2, 'bar': 3},
            {'foo': 1, 'bar': 1},
            {'foo': None, 'bar': 5},
            {'foo': 3, 'bar': 4}]|groupby('foo') }}
```
executing
```python
tmpl.render()
```
leads to ```TypeError: '<' not supported between instances of 'NoneType' and 'int'```

Istota42:
This commit adds support for such case where ``key`` lambda returns ``None`` for groupby funciton.

- fixes #1287 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ Unreleased
 -   Fix a bug that caused imported macros to not have access to the
     current template's globals. :issue:`688`
 -   Add ability to ignore ``trim_blocks`` using ``+%}``. :issue:`1036`
+-   Add ability to group iterables for ``None`` returned as ``key``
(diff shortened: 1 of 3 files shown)


davidism:
Thank you for working on this. However, see #1287 for discussion.

PR #1309, opened 2020-11-13
Added upper and lower versions of camel and snake case filters

sochoa:
**Problem:**

There is no built-in filter that provides a conversion to camel or snake case.

**Solution:**

Added filters for the following:  

* foo bar baz ==> upper_camel_case:  FooBarBaz
* foo bar baz ==> lower_camel_case:  fooBarBaz
* foo bar baz ==> lower_snake_case:  foo_bar_baz
* foo bar baz ==> upper_snake_case:  Foo_Bar_Baz


ThiefMaster:
I don't think this is particularly useful as a builtin Jinja feature... This is something an application with a need for it can easily add.

ThiefMaster:
And just for the record, converting a string with spaces to camel or snake case is not very common. Something much more common (but also outside Jinja's scope) is to convert *between* snake and camel case.

sochoa:
@ThiefMaster its possible that responses like yours above are why the number of contributors to open source projects in the Python community are so low.  Potential contributors get a ton of "that's wrong", just "no", or just complete silence.  There's zero discussion and very little, if any, collaboration before a pull-request is just closed.

ThiefMaster:
If I closed **and locked** the PR I'd understand your argument - it's incredibly annoying if that happens. But a PR being closed does not prevent discussion. And if the discussion is fruitful, a PR is re-opened with a single click of a button.

Also, unless a contribution is a bug fix or for an existing issue, then it's generally a good idea to open an issue first and mention that you'd be willing to work on it simply to avoid getting the PR closed and having wasted time on it (even though you can copy your implementation to your project and use it there with almost no changes).

Issue #1194, opened 2020-04-16
Create a formal definition of Jinja's grammar


davidism:
Every time there's an issue that requires looking at the lexer and parser, I worry due to the amount of complexity and under-documented behavior. The lexer depends on a bunch of very complex regular expressions. The parser sometimes seems to overlap in functionality. The lexer's state machine is mostly a giant set of nested for-loops in a single function.

There is documentation about the template syntax, but it's informal, it implies the grammar without actually saying what the rules are. There are tests for the lexer, but not all its branches. Even the entire test suite managed to miss something with the recent issue #1138 about stripping whitespace.

For long term maintainability, I'd feel much more confident if there was a document about the grammar to refer to, as well as test coverage for all its behavior.

davidism:
[PEP 617](https://www.python.org/dev/peps/pep-0617/), the new PEG parser in Python 3.9, talks about some similar issues, such as overly broad rules in the parser that are then limited in the compiler.

davidism:
Assigned @kevin-brown since he's been working on this in a branch: https://github.com/pallets/jinja/tree/GH-1194-formal-grammar

Others are still welcome to discuss or work on this too, it's an exploratory issue right now.

septatrix:
@kevin-brown Any specific reasons to go with tatsu? It seems to be only sporadically maintained and not as capable as some of the more popular python parsing solutions like [lark](https://github.com/lark-parser/lark)

davidism:
We had extensive discussion about this in Discord, but I don't plan to add any external dependency for this. We'd want to modify or rewrite our lexer and parser to reflect what we learn here. We could use a library in tests to compare the AST our implementation generates to theirs, but otherwise need to keep the code in Jinja.

septatrix:
Yes not adding an extra dependency sounds like a good idea if it can be prevented though isn't tatsu an external dependency? Or is the owner somehow related to the palletsproject?

davidism:
Kevin's just using that in tests and to help build the grammar, the branch I linked isn't a PR right now.

septatrix:
Okay thanks for clarifying

ThiefMaster:
Have you considered using the parser from Python 3.9 (with a backport for older versions)? No idea if it's a good choice for this or not, just curious.

davidism:
[...]

PR #1311, opened 2020-11-16
TatSu Lexer tests

Boyu1997:
Working progress for issue #1194
Adapt [TatSu](https://github.com/neogeny/TatSu) for syntax lexing (and parsing), this will help formalize the current regex implementation to EBNF grammar rule.

This change will address the current informal grammar issue, improve maintainability for lexer. It will make deriving and maintaining grammar and syntax documentation easier. The grammar documentation can follow the EBNF like grammar rule used by TatSu, moving away from the current array of regular expressions.

Changes:
- created `TatsuLexer` class from `Lexer` class, override `tokeniter` function using TatSu
- added parameter `tatsu_lexer` to `Environment` class for specifying which lexer to use

Note: this is a WIP PR, lexer features like comment, raw, custom string, etc. are not implemented.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,6 @@ with open("src/jinja2/__init__.py", "rt", encoding="utf8") as f:
 setup(
     name="Jinja2",
     version=version,
-    install_requires=["MarkupSafe>=1.1","TatSu>=4.4"],
+    install_requires=["MarkupSafe>=1.1", "TatSu>=4.4"],
     extras_require={"i18n": ["Babel>=2.1"]},
 )
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,6 @@ with open("src/jinja2/__init__.py", "rt", encoding="utf8") as f:
 setup(
     name="Jinja2",
     version=version,
-    install_requires=["MarkupSafe>=1.1"],
+    install_requires=["MarkupSafe>=1.1","TatSu>=4.4"],
     extras_require={"i18n": ["Babel>=2.1"]},
 )
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -293,7 +293,6 @@ class Environment:
         lstrip_blocks=LSTRIP_BLOCKS,
         newline_sequence=NEWLINE_SEQUENCE,
         keep_trailing_newline=KEEP_TRAILING_NEWLINE,
-        tatsu_lexer=False,
         extensions=(),
         optimized=True,
         undefined=Undefined,
@@ -304,6 +303,7 @@ class Environment:
         auto_reload=True,
         bytecode_cache=None,
         enable_async=False,
+        tatsu_lexer=False,
     ):
         # !!Important notice!!
         #   The constructor accepts quite a few arguments that should be
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -36,6 +36,7 @@ from .exceptions import TemplateSyntaxError
 from .exceptions import UndefinedError
 from .lexer import get_lexer
 from .lexer import TokenStream
+from .tatsu import get_tatsu_lexer
 from .nodes import EvalContext
 from .parser import Parser
 from .runtime import Context
@@ -292,6 +293,7 @@ class Environment:
         lstrip_blocks=LSTRIP_BLOCKS,
         newline_sequence=NEWLINE_SEQUENCE,
         keep_trailing_newline=KEEP_TRAILING_NEWLINE,
+        tatsu_lexer=False,
         extensions=(),
         optimized=True,
         undefined=Undefined,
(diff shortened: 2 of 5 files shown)


src/jinja2/tatsu.py
```diff
@@ -0,0 +1,137 @@
...
+        environment.keep_trailing_newline,
+    )
+    tatsu_lexer = _tatsu_lexer_cache.get(key)
+    if tatsu_lexer is None:
+        tatsu_lexer = TatsuLexer(environment)
+        _tatsu_lexer_cache[key] = tatsu_lexer
+    return tatsu_lexer
+
+
+GRAMMAR = r"""
```
flosincapite, line 36: Nitpick: make this private, i.e. `_GRAMMAR`

src/jinja2/tatsu.py
```diff
@@ -0,0 +1,137 @@
...
+        comment_begin:'<!--' ~ {comment_data:comment_content}+ comment_end:'-->'
+        ;
+
+    comment_content = /([^-]|-(?!->)|\\n)/ ;
+"""
+
+
+def tatsu_tokenize(source):
+    # use print for debug
+    pprint.pprint(parse(GRAMMAR, source), indent=2, width=20)
```
flosincapite, line 95: Remove debug statements before submitting (okay to keep for now since it's still a WIP)

src/jinja2/tatsu.py
```diff
@@ -0,0 +1,137 @@
...
+    return parse(GRAMMAR, source)
```
flosincapite, line 96: I think it's cleaner to compile the grammar once at top-level, then use that parse method, i.e., above, where `GRAMMAR` is defined, do

```
_GRAMMAR = """
[...]
"""
_TATSU_MODEL = tatsu.compile(_GRAMMAR)
```

Then, here, you can instead call `_TATSU_MODEL.parse(source)`.

That said, I believe `TatSu` does internally memoize calls to `tatsu.parse` (effectively ensuring you don't recompile every time), but I'd rather control that explicitly here.

tests/test_tatsu.py
```diff
@@ -0,0 +1,13 @@
+from jinja2 import Environment
```
flosincapite, line 1: Cool, this is a really good start for testing.

For the next step, I'd like to see many of the existing lexer's tests duplicated.

In order to do that, I think it will be good to use a customized test case (instead of just copying them here). I'll send you links separately via Discord.

src/jinja2/tatsu.py
```diff
@@ -0,0 +1,137 @@
...
+
+        for token in tatsu_tokens:
+            if token["data"]:
+                data = (data or "") + token["data"]
+            else:  # ast
+                if data:
+                    yield lineno.get(), "data", data
+                    lineno.add(data.count("\n"))
+                    data = None
+                yield from get_ast_tokens(lineno, token)
```
flosincapite, line 134: Will `"\n"` never show up in AST tokens? Is this codified somewhere in the grammar? Please document that somewhere if so -- it's a pretty subtle implementation detail.

src/jinja2/environment.py
```diff
@@ -302,6 +303,7 @@ def __init__(
         auto_reload=True,
         bytecode_cache=None,
         enable_async=False,
+        tatsu_lexer=False,
```
davidism, line 306: We won't add a public API for this.

setup.py
```diff
@@ -9,6 +9,6 @@
 setup(
     name="Jinja2",
     version=version,
-    install_requires=["MarkupSafe>=1.1"],
+    install_requires=["MarkupSafe>=1.1", "TatSu>=4.4"],
```
davidism, line 12: This can't be a required dependency, it must only be in `requirements/tests.in`.

src/jinja2/environment.py
```diff
@@ -433,7 +436,13 @@ def overlay(
 
         return _environment_sanity_check(rv)
 
-    lexer = property(get_lexer, doc="The lexer for this environment.")
+    def lexer_getter(self):
```
davidism, line 439: This change isn't needed. We already have `Environment.lexer`, if you need to inject this during tests override it with a subclass.

```python
class TatsuEnvironment(Environment):
    lexer = TatsuLexer()
```

src/jinja2/tatsu.py
Boyu1997: Updated

src/jinja2/tatsu.py
Boyu1997: Switched to `tatsu.parse` in commit 5710224

src/jinja2/tatsu.py
```diff
@@ -0,0 +1,141 @@
...
+    def get(self):
+        return self.value
+
+    def add(self, number):
+        self.value += number
+
+
+class TatsuLexer(Lexer):
+    def tokeniter(self, source, name, filename=None, state=None):
+        _TATSU_MODEL = get_tatsu_model()
```
flosincapite, line 123: This should live in global scope:

```
_TATSU_MODEL = tatsu.compile(_GRAMMAR)
```

PR #1326, opened 2020-12-22
loaders.py: FileSystemLoader consider cwd (.) as default searchpath

hginzel:
FileSystemLoader should use the [current working directory](https://en.wikipedia.org/wiki/Working_directory) as default searchpath.


davidism:
Thanks for looking at this, but I don't think this needs to change.

hginzel:
The change is backward compatible.
I find it as a unix standard, expected behaviour.
KISS

PR #1337, opened 2021-01-25
Fix "If" example in documentation

adamchainz:
The `sourcecode` directive didn't work.


davidism:
Why change it from `jinja` to `tex`?

adamchainz:
Copying the other near examples in this file

ThiefMaster:
tex != text... (and maybe the other `text` cases that actually contain jinja code should use `jinja` as well?)

davidism:
Yeah, should be `jinja` unless that causes a highlighting issue with Pygments, then use `text` instead (not `tex`).

PR #1343, opened 2021-01-31
backport urlize speedup

davidism:
Match leading and trailing punctuation as separate steps. For trailing punctuation, perform a basic check first to reduce backtracking. Update the email regex to reduce backtracking and also be more accurate: the domain part must start with a word character, and the tld only matches word characters.

This is backported from #1195, but only includes the regex speedups, not the other changes.

Checklist:

- ~Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.~
- ~Add or update relevant docs, in the docs folder and in code.~
- ~Add `.. versionchanged::` entries in any relevant code docs.~

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,15 @@
 .. currentmodule:: jinja2
 
+Version 2.11.3
+--------------
+
+Unreleased
+
+-   Improve the speed of the ``urlize`` filter by reducing regex
+    backtracking. Email matching requires a word character at the start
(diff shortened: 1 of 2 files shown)


davidism:
Confirmed that all 2.11.x supported Pythons pass in Tox locally, since CI isn't configured on this branch anymore.

merrlyne:
Is this change planned to be integrated into the version of Jinja that is bundled with Flask?

davidism:
Jinja is not bundled with Flask, it's a dependency, you just upgrade it like any other.

PR #1349, opened 2021-02-02
Add note about Django testing incompatibility

roysmith:
- fixes #<issue number>

Checklist:


davidism:
Thanks, however this doesn't make things any clearer for me. It has no explanation or links for the statement it makes. This seems very specific to how Django integrates with Jinja, not with how Django templates and Jinja templates are similar or different.

roysmith:
Is the issue that I don't provide specific enough details, or that you don't think this should be mentioned at all in the jinja docs?  This is a section on how to convert your django project from django templates to jinja templates.  The issue I'm describing is something that's going to trip people up when doing the conversion.

davidism:
If you're using Jinja in Django, that's not converting from Django templates to Jinja, that's using Jinja in Django. I don't think this needs to be mentioned here.

PR #1353, opened 2021-02-12
Modify the Contributing Guidelines

BALaka-18:
Description : 

Added additional commands to the Contributing Guidelines, for initializing a virtualenv via Git Bash on Windows.

Checklist:

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -103,9 +103,17 @@ First time setup
 
     On Windows, activating is different.
 
-    .. code-block:: text
+    1. If you are activating a virtualenv from the Windows CMD :
 
-        > env\Scripts\activate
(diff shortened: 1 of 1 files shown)


davidism:
The confusion seems to be that while Git Bash is *running* on Windows, it is Bash not CMD, so you need to use `.` (or `source`) like the Linux instructions. The only wrinkle is that you need to refer to `Scripts` instead of `bin` since the env was created on Windows. A good note in case anyone runs across it again, but I wouldn't recommend anyone use Git Bash for general development on Windows, so I'm going to leave this unmerged.

On Windows, I'd rather people stick with CMD or PowerShell, or use WSL instead. Other things like Git Bash, Cygwin, etc. aren't supported because they have various oddities that can cause issues.

PR #1356, opened 2021-02-22
Map filter can use False as default

### LINKED ISSUE #1331: map filter: can't use a false, '', None, [], {} as default (or anything that is automatically cast to false)
https://github.com/pallets/jinja/blob/e39c87c4d474e36dde3b3576a674980e6f4699e1/src/jinja2/filters.py#L70

So this will work:
```
{{ users|map(attribute="username", default="Anonymous")|join(", ") }}
```
but not this:
```
{{ users|map(attribute="username", default="")|join(", ") }}
```

BALaka-18:
According to Issue #1331, the map filter doesn't allow values like [],{},(),None etc ( that automatically default to False ) to be set as the value to the default argument. The user can now set False values to the default argument of the map filter in Jinja2.

- fixes #1331 

Checklist:

### DIFF
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -67,7 +67,7 @@ def make_attrgetter(environment, attribute, postprocess=None, default=None):
         for part in attribute:
             item = environment.getitem(item, part)
 
-            if default and isinstance(item, Undefined):
+            if default is not None and isinstance(item, Undefined):
                 item = default
 
         if postprocess is not None:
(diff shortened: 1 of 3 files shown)


src/jinja2/filters.py
```diff
@@ -66,8 +66,7 @@ def make_attrgetter(environment, attribute, postprocess=None, default=None):
     def attrgetter(item):
         for part in attribute:
             item = environment.getitem(item, part)
-
-            if default and isinstance(item, Undefined):
+            if (default or not default) and isinstance(item, Undefined):
```
lisongmin, line 69: How about `None` as default value here?  it seems `item` will never be Undefined.

it is better to use `default is not None` if None can not use as default value in `map`.

or change the default value to `default=Undefined()` so that None can be use as default value in `map`

BALaka-18: I have made the change in ```filters.py``` so that it now uses ```default is not None```.

PR #1368, opened 2021-03-10
Re-use existing regex for whitespace detection

mvolfik:
Less regexes = win.
Whitespace regex could also detect newlines, but since we're only
matching the part from last newline, it's safer (and more readable)
to use it here.

Checklist:

### DIFF
--- a/src/jinja2/lexer.py
+++ b/src/jinja2/lexer.py
@@ -507,9 +507,7 @@ class Lexer:
         # block suffix if trimming is enabled
         block_suffix_re = "\\n?" if environment.trim_blocks else ""
 
-        # If lstrip is enabled, it should not be applied if there is any
-        # non-whitespace between the newline and block.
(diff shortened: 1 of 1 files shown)


davidism:
Fairly sure that comment/code you removed was due to a previous PR. Can you check to see if you can find that PR so we can make sure we're not missing something?

mvolfik:
You just added comments, some variables and reformatted in 37249c08101060bf5a4a230e412575aa741989db  There was some change in dbbd082bce92cb21ae3e4f49aa32554e05e5d8fc, but this PR doesn't reintroduce the bug in the linked issue (and I suppose there is a test for that somewhere). The regex was introduced in 7d00a40465c89bee141ab5a3db545a20e7d30509, I don't know what was going on there before that. So I don't think I'm reverting some old PR.

davidism:
#858, #1014, and #1183 seem relevant and important enough to warrant a second look at why the code is the way it is.

PR #1374, opened 2021-03-30
Template globals use ChainMap, more docs about globals

### LINKED ISSUE #1376: TypeError: unhashable type: 'Config' (regression)
When installing the current git master of jinja, with a trivial Flask application:

```python
import flask

app = flask.Flask(__name__)

@app.route('/')
def test():
    return flask.render_template('test.html')
```

and then:

- Creating an empty `template/test.html`
- Running `FLASK_APP=test flask run`
- Loading the page and reloading it in the browser

This happens:

```pytb
  File ".../site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File ".../site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File ".../site-packages/flask/app.py", line 1821, in handle_user_exception
  File ".../site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File ".../site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File ".../site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File ".../test.py", line 7, in test
    return flask.render_template('test.html')
  File ".../site-packages/flask/templating.py", line 138, in render_template
    ctx.app.jinja_env.get_or_select_template(template_name_or_list),
  File ".../site-packages/jinja2/environment.py", line 898, in get_or_select_template
    return self.get_template(template_name_or_list, parent, globals)
  File ".../site-packages/jinja2/environment.py", line 851, in get_template
    return self._load_template(name, self.make_globals(globals))
  File ".../site-packages/jinja2/environment.py", line 815, in _load_template
    new_globals = globals.items() - template.globals.items()
TypeError: unhashable type: 'Config'
```

This is caused by #1244 / 40a312e80f4f1b25f201293c3f1a840a1b88191d - the commit before that (f71f5ebab2e76c5ee6f5f705a2703f3891e762bc) works fine. Reproducible with Python 3.6.13, 3.7.10, 3.8.8 and 3.9.2.

davidism:
Add detail about the intended use of environment and template globals after the comments in #295. Rewrite the docs for `get_template`, `select_template`, `get_or_select_template`, and `from_string`.

Changes `environment.make_globals` to always return a `collections.ChainMap`, even if no template globals are given. The map prefers a dict of template globals and falls back to the environment globals. Changes will only modify the template dict, not the environment. Globals are still global, but changes to the template will only ever affect the template, while changes to the environment will still be reflected in all templates, even those that provide template globals. This allows `_load_template` to do `template.globals.update()` instead of more complex logic. Continues #1244

fixes #1376

### DIFF
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -5,6 +5,7 @@ import os
 import sys
 import typing as t
 import weakref
+from collections import ChainMap
 from functools import partial
 from functools import reduce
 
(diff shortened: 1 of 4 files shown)


src/jinja2/environment.py
```diff
@@ -811,16 +812,11 @@ def _load_template(self, name, globals):
...
-                    # made to avoid affecting other templates
-                    if template.globals is self.globals:
-                        template.globals = dict(template.globals, **globals)
-                    else:
-                        template.globals.update(dict(new_globals))
+                # template.globals is a ChainMap, modifying it will only
+                # affect the template, not the environment globals.
+                if globals:
+                    template.globals.update(globals)
+
```
amy-lei, line 819: The if-statement will always be true since the `ChainMap` is always instantiated with env globals (unless a user empties out the default env globals). 

I think this will only be an issue when the template and environment globals overlap. (I didn't see the docs explicitly mention intersections, but I assume it's possible as long as template globals have higher precedence.) So if allowed, something like this fails:

```python
    def test_globals_overlap(self):
        env = Environment(
            loader=DictLoader({"base": "{{ foo }}", "main": "{% include 'base' %}"})
        )

        env.globals['foo'] = 'env'
        tmpl = env.get_template("main", globals={"foo": 1})
        assert tmpl.render() == "1"

        tmpl = env.get_template("base", globals={"foo": 2})
        assert tmpl.render() == "2"

        tmpl = env.get_template("main")
        assert tmpl.render() == "1" # got env instead

        tmpl = env.get_template("base")
        assert tmpl.render() == "2" # got env instead

```

davidism: Those results seem correct based on your original implementation, where template globals are additive. Loading without globals just means the template's globals aren't changed further. Good point about the if statement though, I'll check the logic in general.

davidism: I think I misread your example, it's saying you're not seeing the template globals anymore, so yeah I'll fix that.

davidism: Changed it so `_load_template` calls `make_globals` instead of `get_template` and `select_template`. Updated the test to demonstrate that.

amy-lei (approved):

PR #1386, opened 2021-04-05
WIP: Allow custom escape

### LINKED ISSUE #1377: Feature Request: Allow custom autoescape functions as result of the autoescape function parameter for the Environment
I am currently trying to build an app that generates PDFs based on LaTeX imports.
For this I use [django extensions](https://github.com/weinbusch/django-tex) that renders the LaTeX file and also provides the possibility to escape single variables.

Every variable that is given to the templates system can be modified by the users, so I would need to escape every single occurrence, which is error prone.

Therefore using the autoescape feature of Jinja would be great. 

Reading #503, which was closed unsuccessfully there are currently two way to implement this kind of autoescape

1. Writing an Extension that overwrites the `filter_stream` and and adds a `| my_custom_filter` add the end of every variable. (used by [JinjaSQL](https://github.com/sripathikrishnan/jinjasql/blob/master/jinjasql/core.py))
2. Overwrite the Jinja default escape function by a custom one (used by [Jinja Vanish](https://github.com/mbr/jinja-vanish/blob/master/jinja_vanish/__init__.py))

Both solution feel hacky and unstable.

That why I would suggest to expand the `autoescape` parameter from the Environment to accept not only functions returning a boolean but also a custom escape function.

Like this

```python
from typing import Union, Callable
from jinja2 import Environment, PackageLoader

def do_latex_escape(value: str) -> str:
    return (
        value.replace("&", "\\&")
        .replace("$", "\\$")
        .replace("%", "\\%")
        .replace("#", "\\#")
        .replace("_", "\\_")
        .replace("{", "\\{")
        .replace("}", "\\}")
    )


def smart_autoescape(template_name: str) -> Union[bool, Callable[[str], str]] :
    if template_name.endswith(('.html', '.htm', '.xml')):
        return True  # Use Default Escape Function
    if template_name.endswith(('.latex', '.tex', '.luatex')):
        return do_latex_escape  # for LaTeX Files escape LaTeX
    return False

env = Environment(autoescape=smart_autoescape,
                  loader=PackageLoader('mypackage'))

```

I would expect the following to happen in the above case:
* Given a HTML template
  *  Autoescape and everything as normal
* Given a template with any latex extension
  * autoescape is enabled`
  * the function `do_latex_escape` is wrapped to it does not escape safe text ([example](https://github.com/mbr/jinja-vanish/blob/0.1/jinja_vanish/__init__.py#L58)
  * the default escape filter (`| e`) and (`| escape`) is replaced by the `do_latex_escape` filter
  * autoescaping is done with the new filter
* Everything else
  *  Autoescape is disabled

What do you think about the suggestion? If you like it, I would try to create a PR. 
If I understand it correctly  [Jinja Vanish](https://github.com/mbr/jinja-vanish/blob/master/jinja_vanish/__init__.py) already provides some good starting point for it.

CarliJoy:
- fixes #1377

Still work in progress as there are still some edge cases that I could think of

Checklist:


src/jinja2/filters.py
```diff
@@ -220,9 +250,18 @@ def do_replace(
...
         and not hasattr(s, "__html__")
     ):
-        s = escape(s)
+        s = escape(eval_ctx, s)
     else:
-        s = soft_str(s)
+        s = soft_str(s)  # type: ignore
+
+    # Special case, if user uses Markup class directly to mark
+    # somethin as safe but uses custom escape function
```
jugmac00, line 258:
```suggestion
    # something as safe but uses custom escape function
```

src/jinja2/filters.py
```diff
@@ -263,6 +302,10 @@ def do_xmlattr(
     As you can see it automatically prepends a space in front of the item
     if the filter returned something unless the second parameter is false.
     """
+    # even so we assume that the user want the HTML escape here
```
jugmac00, line 305:
```suggestion
    # even so we assume that the user wants the HTML escape here
```

src/jinja2/filters.py
```diff
@@ -273,7 +316,11 @@ def do_xmlattr(
         rv = " " + rv
 
     if eval_ctx.autoescape:
-        rv = Markup(rv)
+        # We don't assume that using this filter we will use a custom
+        # escape function. But for the sake of completeness we use
+        # the custom mark safe funtion from the eval context here as
```
jugmac00, line 321:
```suggestion
        # the custom mark safe function from the eval context here as
```

src/jinja2/filters.py
```diff
@@ -273,7 +316,11 @@ def do_xmlattr(
         rv = " " + rv
 
     if eval_ctx.autoescape:
-        rv = Markup(rv)
+        # We don't assume that using this filter we will use a custom
```
jugmac00, line 319:
```suggestion
        # We don't assume that using this filter will use a custom
```

Maybe this way? Sentence made no sense to me.

src/jinja2/filters.py
```diff
@@ -772,12 +833,17 @@ def do_indent(
...
+    if hasattr(s, "__html__"):
+        indention = eval_ctx.mark_safe(indention)
+        newline = eval_ctx.mark_safe(newline)
+        # Make sure the correct MarkupClass is used
+        s = t.cast("HasHTML", s)
+        s = eval_ctx.mark_safe(s.__html__())
 
     s += newline  # this quirk is necessary for splitlines method
 
+    # Markup is a subclass of string, so lets just assume it is a string
```
jugmac00, line 845:
```suggestion
    # Markup is a subclass of string, so let's just assume it is a string
```

src/jinja2/nodes.py
```diff
@@ -62,16 +64,51 @@ def __new__(mcs, name, bases, d):
 class EvalContext:
     """Holds evaluation time information.  Custom attributes can be attached
     to it in extensions.
+
+    .. versionchanged:: 3.0
+
+    - Added ``get_escape_function`` and ``mark_safe`` functions
+    - allow autoescaped to be not only boolean but also an
```
jugmac00, line 71:
```suggestion
    - allow autoescape to be not only boolean but also an
```

Maybe?

src/jinja2/nodes.py
```diff
@@ -62,16 +64,51 @@ def __new__(mcs, name, bases, d):
...
         if callable(environment.autoescape):
-            self.autoescape = environment.autoescape(template_name)
+            self.autoescape = bool(environment.autoescape(template_name))
         else:
-            self.autoescape = environment.autoescape
+            self.autoescape = bool(environment.autoescape)
         self.volatile = False
 
+        # We need to keep  Markup Class if existing as autoescape can be
+        # overwritten by {% autoescape %} environment.
```
jugmac00, line 85: I do not understand this sentence.

src/jinja2/runtime.py
```diff
@@ -615,8 +626,14 @@ def __call__(self, *args, **kwargs):
         # argument to callables otherwise anyway.  Worst case here is
         # that if no eval context is passed we fall back to the compile
         # time autoescape flag.
+
         if args and isinstance(args[0], EvalContext):
             autoescape = args[0].autoescape
+            # If the eval contex is available we use it to determine
```
jugmac00, line 632:
```suggestion
            # If the eval context is available we use it to determine
```

src/jinja2/utils.py
```diff
@@ -593,17 +598,56 @@ def select_autoescape(
     If nothing matches then the initial value of autoescaping is set to the
     value of `default`.
 
+    The `special_extensions` is a dictionary whose keys are the extensions
+    to be considered and the values are the escape function to be used
+    to escape this kind of files.
```
jugmac00, line 603:
```suggestion
    to escape these kind of files.
```

src/jinja2/utils.py
```diff
@@ -593,17 +598,56 @@ def select_autoescape(
     If nothing matches then the initial value of autoescaping is set to the
     value of `default`.
 
+    The `special_extensions` is a dictionary whose keys are the extensions
```
jugmac00, line 601:
```suggestion
    The `special_extensions` is a dictionary which keys are the extensions
```

src/jinja2/utils.py
```diff
@@ -593,17 +598,56 @@ def select_autoescape(
...
+    to be considered and the values are the escape function to be used
+    to escape this kind of files.
+
+    I.e. if you use the `latex package <https://pypi.org/project/latex/>`_
```
jugmac00, line 605:
```suggestion
    I.e. if you use the `latex package <https://pypi.org/project/latex/>`_,
```

src/jinja2/filters.py
```diff
@@ -772,12 +833,17 @@ def do_indent(
 
     newline = "\n"
 
-    if isinstance(s, Markup):
-        indention = Markup(indention)
-        newline = Markup(newline)
+    if hasattr(s, "__html__"):
+        indention = eval_ctx.mark_safe(indention)
+        newline = eval_ctx.mark_safe(newline)
+        # Make sure the correct MarkupClass is used
```
jugmac00, line 839:
```suggestion
        # Make sure the correct Markup class is used
```

At least in other places it is written like that.

docs/api.rst
```diff
@@ -259,6 +268,65 @@ defaults in the future might change.
...
+        """
+        s.replace("war", "peace")
+
+    env = Environment(
+        autoescape=select_autoescape(special_extensions={".world": escape_to_peace}),
+        loader=PackageLoader("mypackage"),
+    )
+
+
+Note that for `.world` files the `{{ var|e }}` and `{{ var | escape }}`
```
ThiefMaster, line 295: I think all this should use double backticks (assuming you want the equivalent to `single-backtick` highlighting on github)

src/jinja2/utils.py
```diff
@@ -653,6 +697,60 @@ def htmlsafe_json_dumps(
     )
 
 
+@lru_cache(100)
+def get_wrapped_escape_class(custom_escape: Callable[[Any], str]) -> Type[Markup]:
+    """
+    Use a simple escape function to generate a wrapped Markup class
+
+    This class uses the given ``custom_escape`` function to escape
+    the value andat the same time makes sure that no already escaped
```
jugmac00, line 706:
```suggestion
    the value and at the same time makes sure that no already escaped
```

docs/api.rst
```diff
@@ -259,6 +268,65 @@ defaults in the future might change.
...
+filters are replaced with the custom escape function.
+
+To mark a string as safe please use the :meth:`Environment.get_markup_class`
+instead of direct :class:`Markup` calls::
+
+
+    template = env.get_template("message_to_the.world")
+    # the content of the template is simply assumed to by
+    """
+    <h1>My Message to the world<h1>
```
ThiefMaster, line 305:
```suggestion
    <h1>My Message to the world</h1>
```

docs/api.rst
```diff
@@ -259,6 +268,65 @@ defaults in the future might change.
...
+
+
+.. admonition:: A word of caution
+
+    Be aware that mixing files that using different custom escape
+    functions set by autoescape within of one render command,
+    can lead to unexpected behavior.
+    In general the ``{% include %}`` directive works fine but especially
+    ``{% extends %}`` commands can have unexpected outcomes as main template
+    overwrites the context of the included one.
```
ThiefMaster, line 323: Can we just make this throw an error? Possibly with a policy setting to disable it - but I'd rather have a sane/safe default...

src/jinja2/utils.py
```diff
@@ -653,6 +697,60 @@ def htmlsafe_json_dumps(
...
+        def escape(cls, s):
+            """
+            Make sure the custom escape function does not escape
+            already escaped strings
+            Also make sure the escaped string is marked as escaped
+            with the correct class
+
+            If the object has an ``__html__`` method, it is called
+            and the return value is assumed to already be safe for HTML
+            / resp what ever is escaped currently.
```
jugmac00, line 740:
```suggestion
            / resp. what ever is escaped currently.
```

src/jinja2/utils.py
```diff
@@ -653,6 +697,60 @@ def htmlsafe_json_dumps(
...
+            The HTML Attribue shall be understood simply as
```
jugmac00, line 741:
```suggestion
            The HTML attribute shall be understood simply as
```

src/jinja2/utils.py
```diff
@@ -653,6 +697,60 @@ def htmlsafe_json_dumps(
...
+            return cls(custom_escape(s))  # noqa: B902
+
+    return MarkupWrapper
+
+
+get_wrapped_escape_class
+
+
 class Cycler:
     """Cycle through values by yield them one at a time, then restarting
```
jugmac00, line 755:
```suggestion
    """Cycle through values by yielding them one at a time, then restarting
```

src/jinja2/compiler.py
```diff
@@ -65,7 +64,12 @@ def has_safe_repr(value):
     if value is None or value is NotImplemented or value is Ellipsis:
         return True
 
-    if type(value) in {bool, int, float, complex, range, str, Markup}:
+    if isinstance(value, Markup):
+        # If a custom Markup Class is used,
```
ThiefMaster, line 68:
```suggestion
        # If a custom Markup class is used,
```

tests/test_custom_escape.py
```diff
@@ -0,0 +1,270 @@
...
+                }
+            ),
+        )
+        # First test simple stuff
+        t = env.get_template("macro.star")
+        assert t.render(foo=chars) == "<star~+>;bar<star~+>"
+        t = env.get_template("inc.plus")
+        # Again not 100% what was expected but we documented it
+        # If you can fix it, that the correct escape functions are used,
+        # dont forget to adopt the documentaion
```
jugmac00, line 224:
```suggestion
        # don't forget to update the documentation
```

tests/test_custom_escape.py
```diff
@@ -0,0 +1,270 @@
...
+                    "{% endblock %}",
+                }
+            ),
+        )
+        # First test simple stuff
+        t = env.get_template("macro.star")
+        assert t.render(foo=chars) == "<star~+>;bar<star~+>"
+        t = env.get_template("inc.plus")
+        # Again not 100% what was expected but we documented it
+        # If you can fix it, that the correct escape functions are used,
```
jugmac00, line 223:
```suggestion
        # If you can fix it so that the correct escape functions are used,
```

src/jinja2/compiler.py
```diff
@@ -623,6 +627,14 @@ def write_commons(self):
         """
         self.writeline("resolve = context.resolve_or_missing")
         self.writeline("undefined = environment.undefined")
+        # make sure in the root render function the correct escape is used
+        self.writeline("escape = context.eval_ctx.get_escape_function()")
+        # make sure in the root render function the coorect Markup
+        # class ise used
```
ThiefMaster, line 633:
```suggestion
        # class is used
```

but maybe these comments here are a bit too verbose?

src/jinja2/compiler.py
```diff
@@ -623,6 +627,14 @@ def write_commons(self):
...
+        self.writeline("Markup = context.eval_ctx.mark_safe")
+        # Custom Wrappers have a differnt naming
```
ThiefMaster, line 635:
```suggestion
        # Custom Wrappers have a different naming
```

tests/test_security.py
```diff
@@ -91,6 +91,31 @@ def test_template_data(self, env):
...
+        t = env.from_string(
+            "{% macro say_hello(name) %}"
+            "$<p>Hello {{ name }}!</p>${% endmacro %}"
+            '{{ say_hello("<$$foo$$>") }}'
+        )
+        my_escape = env.default_markup_class.escape
+        escaped_out = "$<p>Hello <€€foo€€>!</p>$"
+        assert t.render() == escaped_out
+        # Make sure that the same Markup class is used everywhere
+        # somehow even different instances of the escape class are use
```
jugmac00, line 105:
```suggestion
        # somehow even different instances of the escape class are used
```

src/jinja2/environment.py
```diff
@@ -201,14 +204,57 @@ class Environment:
...
+            Use this with care.
+            Not all functions within Jinja get the context
+            that is required to determine the correct escape function.
+            If you run in troubles simply use different environments
+            with custom ``default_escape`` (see below) for
+            each file type.
+
+            .. versionchanged:: 3.0
+                if `autoescape` function returns not True or False but a Callable
+                this is assumed to be a custom escape function
```
ThiefMaster, line 229:
```suggestion
                if the `autoescape` function doesn't return True or False but a
                callable, it is assumed to be a custom escape function
```

CHANGES.rst
```diff
@@ -7,6 +7,7 @@ Unreleased
 
 -   Drop support for Python 2.7 and 3.5.
 -   Bump MarkupSafe dependency to >=1.1.
+-   Allow custom auto escape functions :issue:`1377`
```
jugmac00, line 10:
```suggestion
-   Allow custom autoescape functions :issue:`1377`
```
everywhere else it is written together

src/jinja2/environment.py
```diff
@@ -323,6 +371,14 @@ def __init__(
         self.optimized = optimized
         self.finalize = finalize
         self.autoescape = autoescape
+        if isclass(default_escape):
+            self.default_markup_class = default_escape
+        elif default_escape != html_escape:
+            self.default_markup_class = get_wrapped_escape_class(default_escape)
+        elif callable(self.autoescape) and callable(self.autoescape(None)):
+            self.default_markup_class = get_wrapped_escape_class(self.autoescape(None))
```
ThiefMaster, line 379: We call `self.autoescape` twice here, but since we can't use `:=` yet I don't think there's a pretty way of avoiding it...

src/jinja2/ext.py
```diff
@@ -203,7 +201,6 @@ class InternationalizationExtension(Extension):
 
     tags = {"trans"}
 
-    # TODO: the i18n extension is currently reevaluating values in a few
```
ThiefMaster, line 206: removing that line probably wasn't intended...

docs/api.rst
```diff
@@ -259,6 +268,65 @@ defaults in the future might change.
 Inside the templates the behaviour can be temporarily changed by using
 the `autoescape` block (see :ref:`autoescape-overrides`).
 
+Note that :func:`~jinja2.select_autoescape` offers also a parameter
+`special_extensions` that allows you to define a custom escape function,
+i.e. if you handling LaTeX files
```
jugmac00, line 273:
```suggestion
i.e. if you work with LaTeX files
```

src/jinja2/filters.py
```diff
@@ -152,12 +170,17 @@ def _prepare_attribute_parts(
...
 
-def do_forceescape(value: "t.Union[str, HasHTML]") -> Markup:
-    """Enforce HTML escaping.  This will probably double escape variables."""
+@evalcontextfilter
+def do_forceescape(eval_ctx: "EvalContext", value: "t.Union[str, HasHTML]") -> Markup:
+    """
+    Enforce HTML escaping.  This will probably double escape variables.
+
+    .. versionchanged:: 3.0
+        made function context aware to use context based escape filter
```
ThiefMaster, line 179: Is this really relevant? It's an internal implementation detail; I would not really expect this in the changelog. Or are the filter implementations considered public API?

docs/api.rst
```diff
@@ -259,6 +268,65 @@ defaults in the future might change.
...
+(see description of :func:`~jinja.select_autoescape` above for example).
+
+To write your own custom escape function simply return it for the
+wanted template extension. For instance if you are fan of peace in
```
jugmac00, line 277: This sentence sounds a bit rough, but I can't come up with something better, as I am not sure what you intend to say.

As far as I see you do not return the custom function, but configure the Environment with it, right?

docs/api.rst
```diff
@@ -259,6 +268,65 @@ defaults in the future might change.
...
+the world::
+
+    def escape_to_peace(s):
+        """
+        Your custom escape function. You only have to take care
+        that your escaping is done properly, everything else like
+        preventing multiple escapes and marking the string as safe
+        is done by Jinja itself.
+        """
+        s.replace("war", "peace")
```
jugmac00, line 287: Is there really not return statement necessary?

src/jinja2/utils.py
```diff
@@ -653,6 +697,60 @@ def htmlsafe_json_dumps(
...
+                "it is a safe string"
+            """
+            if hasattr(s, "__html__"):
+                return cls(s.__html__())
+            return cls(custom_escape(s))  # noqa: B902
+
+    return MarkupWrapper
+
+
+get_wrapped_escape_class
```
ThiefMaster, line 751: 👀

docs/api.rst
```diff
@@ -259,6 +268,65 @@ defaults in the future might change.
...
+    # We know that everything that ends on world will use the special
+    mark_safe = env.get_markup_class(".world")
+    my_msg = mark_safe("Make love not war!")
+    reply = "We want war!"
+    template.render(my_msg=my_msg, reply=reply)
+
+
+.. admonition:: A word of caution
+
+    Be aware that mixing files that using different custom escape
```
jugmac00, line 318:
```suggestion
    Be aware that mixing files that use different custom escape
```

docs/api.rst
```diff
@@ -625,6 +693,29 @@ functions to a Jinja environment.
     Genshi.  It's expected that more template engines and framework will pick
     up the `__html__` concept soon.
 
+.. admonition:: Attention
+
+    Especially when using custom escape function do not use the
```
jugmac00, line 698:
```suggestion
    Especially when using a custom escape function do not use the
```

docs/api.rst
```diff
@@ -625,6 +693,29 @@ functions to a Jinja environment.
...
+    :class:`Markup` directly to mark string as safe or to escape it.
```
jugmac00, line 699:
```suggestion
    :class:`Markup` directly to mark a string as safe or to escape it.
```

docs/api.rst
```diff
@@ -625,6 +693,29 @@ functions to a Jinja environment.
...
+    Instead use :meth:`Environment.get_markup_class` to get the
+    correct class.
+    Usage ``Markup = get_markup_class("mytemplate.ext")``
+
+
+This is required as the :class:`Markup` class implements calls for its
+``Markup.escape`` method i.e. when using the ``join`` or
+the modulo ``%`` operator.
+So it is important that correct :class:`Markup` subclass is used
```
jugmac00, line 708:
```suggestion
So it is important that the correct :class:`Markup` subclass is used
```

docs/api.rst
```diff
@@ -625,6 +693,29 @@ functions to a Jinja environment.
...
+when using custom escaping defined by the :ref:`autoescaping` settings.
+
+The correct Markup class from a custom escape function is generated
+using the helper class:
+
+..
+    comment:: Somehow using the lru_cache wrapper the autodoc does not get the correct parameters
```
jugmac00, line 715:
```suggestion
    comment:: Somehow when using the lru_cache wrapper the autodoc does not get the correct parameters
```

src/jinja2/compiler.py
```diff
@@ -623,6 +627,14 @@ def write_commons(self):
         """
         self.writeline("resolve = context.resolve_or_missing")
         self.writeline("undefined = environment.undefined")
+        # make sure in the root render function the correct escape is used
+        self.writeline("escape = context.eval_ctx.get_escape_function()")
+        # make sure in the root render function the coorect Markup
```
jugmac00, line 632:
```suggestion
        # make sure in the root render function the correct Markup
```

src/jinja2/compiler.py
jugmac00, line 633:
```suggestion
        # class is used
```

src/jinja2/compiler.py
jugmac00, line 635:
```suggestion
        # Custom wrappers have a different naming
```

src/jinja2/environment.py
```diff
@@ -201,14 +204,57 @@ class Environment:
...
+
+            As of Jinja 2.4 this can also
             be a callable that is passed the template name and has to
             return ``True`` or ``False`` depending on autoescape should be
             enabled by default.
 
+            As of Jinja 3.0 the autoescape can be even smarter.
+            If the given function does not return a boolean but a
+            function again, this function is considered to be the
+            escape function that shall be used. So you can used the
```
jugmac00, line 217:
```suggestion
            escape function that shall be used. So you can use the
```

src/jinja2/environment.py
```diff
@@ -201,14 +204,57 @@ class Environment:
...
+            same enviroment to autoescape LaTeX and HTML files.
```
jugmac00, line 218:
```suggestion
            same environment to autoescape LaTeX and HTML files.
```

src/jinja2/environment.py
```diff
@@ -201,14 +204,57 @@ class Environment:
...
+
+            Use this with care.
+            Not all functions within Jinja get the context
+            that is required to determine the correct escape function.
+            If you run in troubles simply use different environments
```
jugmac00, line 223:
```suggestion
            If you run in trouble simply use different environments
```

src/jinja2/environment.py
```diff
@@ -201,14 +204,57 @@ class Environment:
...
+            with custom ``default_escape`` (see below) for
+            each file type.
+
+            .. versionchanged:: 3.0
+                if `autoescape` function returns not True or False but a Callable
```
jugmac00, line 228:
```suggestion
                if the `autoescape` function does not return True or False but a Callable
```

src/jinja2/environment.py
```diff
@@ -349,6 +405,27 @@ def __init__(
...
         _environment_sanity_check(self)
 
+    def get_markup_class(
+        self, template_name: t.Optional[str] = None
+    ) -> t.Type["Markup"]:
+        """
+        Get the correct :class:`Markup` for the given template name.
+
+        Use this instead of the default :class:`Markup` to mark a string as
+        safe, especially when using custom escpaing.
```
jugmac00, line 415:
```suggestion
        safe, especially when using custom escaping.
```

jugmac00 (changes requested):
Hi @CarliJoy,

I suggested some improvements regarding spelling and phrasing. Take my suggestions with a grain of salt as I am neither a Flask maintainer nor a native English speaker.

As I found many typos, I strongly suggest that you consider installing  spell checker for your IDE or editor.

If that is not possible, please give me a ping after your merged /considered my suggestions, then I will download your branch and review it locally - I just through the PR in the GitHub UI.

That said - wow, what a massive PR - hats off!

CarliJoy:
Hi guys,

thanks for your efforts, will work you comments in tomorrow or the days after tomorrow.
Actually PyCharm has Spell Check enabled but especially for grammar stuff it's not that good yet.
Probably I also have to change some settings as I guess I just missed some spell underlines 🤷🏻 🤦🏻 
As I wrote in #1377, I am bit of (maybe not only a bit ;-) ) dyslexic, even in my native language (the one I probably share with @jugmac00). 

So thanks for taking the effort in checking and correcting -> much appreciated!

Also is the first time for me that an PR in github actually got reviewed, so I also need to get started on this :)

src/jinja2/utils.py
CarliJoy: Aehm, how did that ended up there...
Strange that no linter found it, as it is completely without any function.

tests/test_security.py
CarliJoy: Actually I thought you would only correct the documentation and not the comments as well.
But I guess the point about comments is that people understand them, so it makes sense. But well yeah so MANY misstakes... sorry :(

src/jinja2/filters.py
ThiefMaster: > don't know at all. I thought you are the maintainer so you tell me

I'm not sure if we consider it public API. In the end those things are meant to be used in templates; using them outside Jinja is probably pretty close to using undocumented functions / internals directly...

> Is there something as too many comments ;-) ?

Yes, absolutely. things like `# iterate over the list` for example. ;)

Anyway, that's not the case here - my point was more that purely internal changes would just add noise to the end user docs, and thus not be useful. But since this does change the way the function is called, it could be worth mentioning if we consider someone using it from outside a template a valid usecase...

src/jinja2/utils.py
CarliJoy: Not part of my PR but I still add it

docs/api.rst
CarliJoy: done, have a look. I mark this resolved so far

src/jinja2/utils.py
```diff
@@ -593,17 +598,56 @@ def select_autoescape(
...
+    }
 
     def autoescape(template_name):
         if template_name is None:
             return default_for_string
         template_name = template_name.lower()
+        # create dictionary so that have lower case keys sorted from longest
+        # to shortest, so we can easily find the best match
+        # i.e. if there are .la.tex and .tex endings in the dictionary,
+        # .la.tex ending should be found if the filename if .la.tex is given
```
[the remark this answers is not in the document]
jugmac00: I understand the problem, but the sentence still makes no sense to me, especially the `if the filename if .la.tex is given`

Maybe something like this?

```suggestion
        # and given a filename like e.g. overview.la.tex the .la.tex ending should be found
```

src/jinja2/utils.py
```diff
@@ -593,17 +598,56 @@ def select_autoescape(
...
+        for key, func in sorted(
+            special_extensions.items(), key=lambda x: len(x[0]), reverse=True
+        ):
```
ThiefMaster, line 648: The commend above is overly verbose IMO. Something like "Lookup autoescape function using the longest matching suffix" explains it just as well.

That aside, do we really need this complexity? The only file type where double extensions are somewhat common are archives (`.tar.{gz,xz,bz2}`). So I'd rather just match using the actual extension (as determined by `os.path.splitext`) instaed of making this more complex than necessary.

src/jinja2/environment.py
```diff
@@ -855,7 +949,62 @@ def _load_template(self, name, globals):
...
+        """
+        Raise an Error if we expect different custom escape functions
+        or Markup classes used in a way that a user could get
+        unexpected results.
+
+        :param name: The name of first template if known
+        :param parent: the name or names of the templates that
+        :param caller: the name of the function the
+
+        :raises TemplateConfigurationError
```
ThiefMaster, line 967:
```suggestion
        :raise: TemplateConfigurationError
```

(or `:raises:`, see what's more common in the codebase)

src/jinja2/environment.py
```diff
@@ -855,7 +949,62 @@ def _load_template(self, name, globals):
...
+            return
+
+        # The dynamic nature of get_wrapped_escape_class does not allow
+        # for direct compares, so we need to compare the uid
+        if (
+            hasattr(base_class, "get_class_uid")
+            and hasattr(parent_class, "get_class_uid")
+            and base_class.get_class_uid == parent_class.get_class_uid  # type: ignore
+        ):
+            return
```
ThiefMaster, line 997: Why so much complexity? What's wrong with something as simple as `if base_class != parent_class: raise ...`

src/jinja2/utils.py
CarliJoy: Well I can imagine end users that want a custom escape function for `.special.html` and also `i.dont.like.spaces.but.points.html`

src/jinja2/utils.py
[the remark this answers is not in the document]
CarliJoy: like Adrians formulation and used it.

src/jinja2/utils.py
ThiefMaster: What would be the use case for this? I get `.html` vs `.tex`, but anything `.html` should probably use the standard HTML escaper...

CarliJoy: People could (arguably maybe misuse) use the escape function to escape all bad words like "fuck" from endusers.
Or people use a javascript framework that replaces a specific part of the HTML, so the end users should never be allow to write this special string.

IMO libraries should be enablers and if I see a potential bug/edge case I rather fix it right away as frustrate an end user/library user in the end.

src/jinja2/environment.py
CarliJoy: Because I tested it and it won't work. Because the classes given by two calls of get_wrapped_escape_class actually aren't the same even using the lru_cache!

ThiefMaster: ah I didn't notice we're not dealing directly with the custom Markup classes (or escape functions) here..

src/jinja2/utils.py
ThiefMaster: > arguably maybe misuse

indeed, which is why I'm -1 on this

> Or people use a javascript framework that replaces a specific part of the HTML, so the end users should never be allow to write this special string.
> IMO libraries should be enablers and if I see a potential bug/edge case I rather fix it right away as frustrate an end user/library user in the end.

I get your point, but when people go into *that* kind of obscure territory they should probably provide their own `select_autoescape` implementation... In the end, libraries/framework should do their main jobs well, and if someone wants to do obscure things they can do that by subclassing/overriding things (or in this case providing their own autoescape function)

CarliJoy: Sure but honestly I don't know any better solution :dizzy:. 
Using splitext simply *feels* wrong for me:
```
from os.path import splitext
splitext('foobar.ver1.html')
('foobar.ver1', '.html')
splitext('foobar.gz.html')
('foobar.gz', '.html')
splitext('foobar.tar.gz')
('foobar.tar', '.gz')
```

So please provide a change request with your preferred solution . Already spend much more time on this PR then intended and really would not like to invest time implementing something that *feels* like implementing a 🐞 bug  for me. :lotus_position:

src/jinja2/environment.py
CarliJoy: resolved?

CarliJoy:
@davidism is okay for your to take care of the conflict in `docs/api.rst` that was introduced through: #1391?

src/jinja2/environment.py
CarliJoy: Ok did some further digging. The lru_cache solves the issue so we don't need the complexity. I removed it, as I could write a test that calls 997.
In my original testing I extracted the Markup classes from the template context that is generated through exec().
That was probably the reason why the lru_cache wasn't working.

Pushing the changes soon.

CarliJoy:
the moment when you worked 4hours to resolve a merge conflict and you realise that in the meanwhile the next PR was merge that causes the next merge conflict 😭

CarliJoy:
@ThiefMaster I fixed everything so far. Documentation is up to date, tests are as well. Currently no Merge Conflicts.
How likely is that you will give your ok for the PR? Because I rather would not to spend any more work into something that goes into the bin.

Also because this change touches code on so many places, that other PR most likely will introduce other Merge Conflicts.

ThiefMaster:
> Because I rather would not to spend any more work into something that goes into the bin.

Don't worry about that. If we believed this was something we don't want we'd have told you from the beginning.

However, this is a large and complex PR, and while I will have another look, I don't think I'm going to be the one that will take the final decision on it and merge it; I'll leave that up to @davidism once he has the time for it.

I wouldn't worry too much about merge conflicts - most of them can probably be resolved right before merging.

docs/api.rst
```diff
@@ -220,6 +220,98 @@ useful if you want to dig deeper into Jinja or :ref:`develop extensions
...
+``&``, ``<``, ``>``, ``'``, and ``"`` in string `s`
+to HTML-safe sequences.  It is intended to be used  if you need to
+display text that might contain such characters in HTML.
+
+The result of a call of the ``escape(s: str)`` function is a
+``Markup`` class.
+This class provides an ``.__html__()`` method, which is used internally
+as an indicator that the string returned by this method is safe.
+This way of defining a save/HTML string is also used by other Template
+System or things like widget in Jupter Notebook.
```
jugmac00, line 252:
```suggestion
System or things like widget in Jupiter Notebook.
```

CarliJoy:
@davidism once you check the code, have a lock especially on the async stuff. I have no experience with async coding what so ever, so my changes are more like an educated guess.
I still tried to write some tests but not sure if I covered all the cases.
Maybe you also find a way make EvalContext known to the async Macros....

CarliJoy:
@davidism can I support/help you in any way with the review?

davidism:
We won't have time to review something of this magnitude before release, so we're marking this for 3.1. I'm currently adding type annotations to the entire code base, so this is likely to need a rebase after that. Rebasing in general to squash commits would also be helpful, as 71 is too many to review.

CarliJoy:
Sorry I am not yet that experienced with rebasing.
What would be the workflow in an existing PR like this?
I would assume:
1. After your added the Type Annotations create a new branch like `custom_escape_squashed` base on main
2. Merge squash the existing  `custom_escape` in the new branch
3. Create a new PR

Would that work/help?

jugmac00:
I think @davidism meant, that you should squash (https://levelup.gitconnected.com/how-to-squash-git-commits-9a095c1bc1fc) your 71 commits into one (and maybe force push the result to this branch).

After @davidism applied the type annotations, and merged them into master/main, you can rebase onto master/main - but only this one commit (https://stackoverflow.com/questions/7929369) - which makes it much easier.

Finally, force-push to this branch.

davidism:
Hold off on squashing until I merge the type annotations. It might actually be easier to recreate after the fact instead of trying to resolve the merge conflicts, which will be significant.

CarliJoy:
@davidism  just tell me once you finished  the merge of the type annotations...

CarliJoy:
> I think @davidism meant, that you should squash (https://levelup.gitconnected.com/how-to-squash-git-commits-9a095c1bc1fc) your 71 commits into one (and maybe force push the result to this branch).
> 
> After @davidism applied the type annotations, and merged them into master/main, you can rebase onto master/main - but only this one commit (https://stackoverflow.com/questions/7929369) - which makes it much easier.
> 
> Finally, force-push to this branch.
> 
> So, it is neither required to create a new branch or a new PR.

Thanks for the detailed answer

davidism:
They're merged

CarliJoy:
@davidism Did the rebase and combination as requested.
One thing: When are you planning to release 3.0?
Are you following SemVer?
Because if you do it would be good not only to deprecate Markup but directly make people aware of the functions that will return the correct Markdown class?
Otherwise there will be one braking change between 3.0 and 3.1.

My suggestion would be to create a small PR that only provides the functions to get the (then only hardcoded) Markdown and escape functions.

Is this understandable?
What do you thing?

davidism:
We don't follow semver.

I don't understand what the breaking change is, the only thing this PR should be doing is adding customization capabilities, not changing existing behavior.

CarliJoy:
First of all. Sorry I missed that release. Big thing, updating all libraries at once. Congratulations 🎉 
> We don't follow semver.
Good to know.
> 
> I don't understand what the breaking change is, the only thing this PR should be doing is adding customization capabilities, not changing existing behavior.

Braking change is for all creators of custom extensions and people that use Markdown directly in normal Python code to define a string as safe.
If they combine it with custom escape it could lead to unexpected behaviour (i.e. when using the `%` operator).
Tried to explain that in the documentation of the PR as good as possible.

davidism:
That's not a breaking change though, that use case will continue to work, but they'll need to update if they want to take advantage / make correct use of customization that this enables. That's what a good changelog message and docs are for.

src/jinja2/utils.py
```diff
@@ -846,8 +961,10 @@ def __new__(cls, base, encoding=None, errors="strict"):  # type: ignore
 
 def escape(s: t.Any) -> str:
     warnings.warn(
-        "'jinja2.escape' is deprecated and will be removed in Jinja"
-        " 3.1. Import 'markupsafe.escape' instead.",
+        "'jinja2.Markup' is deprecated and will be removed in Jinja"
```
CarliJoy, line 964: What to do with this warning? Simply remove it, as it is supposed to be removed in 3.1 anyway?

src/jinja2/utils.py
```diff
@@ -837,7 +950,9 @@ class Markup(markupsafe.Markup):
     def __new__(cls, base, encoding=None, errors="strict"):  # type: ignore
         warnings.warn(
             "'jinja2.Markup' is deprecated and will be removed in Jinja"
-            " 3.1. Import 'markupsafe.Markup' instead.",
+            " 3.1. Use Environment.get_markup_class and "
```
CarliJoy, line 953: What to do with this warning? Simply remove it, as it is supposed to be removed in 3.1 anyway?

src/jinja2/utils.py
davidism: Yes, this will be removed in 3.1, so you can omit any changes to it from your PR.

src/jinja2/runtime.py
```diff
@@ -50,19 +48,31 @@ def identity(x):
...
+        added optional parameter escape_function to make
+        use the contex based escape function
+    """
     buf = []
     iterator = map(soft_str, seq)
     for arg in iterator:
         buf.append(arg)
         if hasattr(arg, "__html__"):
-            return Markup("").join(chain(buf, iterator))
+            return "".join(map(escape_func, chain(buf, iterator)))
```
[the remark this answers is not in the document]
CarliJoy: True. But then both the original type annotations and the function description are wrong.
It would return Markup if input has any Markup like element but string if all is string.
Is this intended?
If you say yes then I will change it so it return Markup (and also change the description and type annotation)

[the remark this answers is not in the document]
CarliJoy: I know that Markup is a subclass of string so it is formal correct. But as Markup has a different meaning IMO the typing should state this (as this is done in other functions as well).
Looking at the code a bit more in detail I see that functionality is changed even more, as I escaped the strings, which wasn't done before.
I have to replace it. Hope I can do this, this week or next.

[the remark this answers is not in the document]
CarliJoy: Just realized that this function is called by no test at all currently.
Only `test_ex.py` `TestExtensions.test_do` is actually calling `CodeGenerator.visit_Concat` that would lead to a call possible call of `runtime.py`s  `markup_join`.

I will add a test especially for the custom escape now but thought it might be good to know that there a number of not catched test cases considering the logic of `visit_Concat`.

This comment is only FYI @davidism

CHANGES.rst
```diff
@@ -26,6 +26,8 @@ Released 2021-05-11
 
 -   Drop support for Python 2.7 and 3.5.
 -   Bump MarkupSafe dependency to >=1.1.
+-   Allow custom autoescape functions :issue:`1377`
```
CarliJoy, line 29: As this won't be a part of 3.0 but 3.1 I will change this and all related docstrings to 3.1

CarliJoy:
@davidism I changed things as requested. If there are still any issues, let me know.

CarliJoy:
@davidism and @ThiefMaster will this ever be merged?

davidism:
Yes, I do plan to eventually merge this. It is a huge set of changes, and I just have not been able to sit down and properly review it. I can't block the release of 3.1 on this, so I'm moving it to 3.2 for now.

CarliJoy:
@davidism just another ping. Anything I can do to support you on this MR?

CarliJoy:
@davidism And here the yearly ping again 😉 . Anything I could do to finally finally get this merged?

CarliJoy:
Happy 3rd birthday MR :-)

@davidism I know this is a huge MR, due to the way how autoescape is handled currently.

In the meanwhile I gained a lot more experience, also with async. I could review the MR again, rebase it and see if I am able to maybe reduce the number of changes lines.

But I would only invest this time if there are resources on pallets side to review and merge the MR. Otherwise I will check in a year again.

Feel free to reach out with tasks and questions related to this MR 🤗

docs/api.rst
```diff
@@ -250,6 +347,8 @@ works roughly like this::
         if template_name is None:
             return False
         if template_name.endswith(('.html', '.htm', '.xml'))
+            return True
+        return False
```
MinestoPix, line 351:
```suggestion
        return template_name.endswith(('.html', '.htm', '.xml'))
```

CarliJoy:
Almost forgot. Happy 4th Birthday Pull Request...

davidism:
Please stop pinging me, it just adds stress to an already full time maintainer job. You'll know it's being worked on or merged because you'll see additional activity and notifications.

PR #1391, opened 2021-04-10
Markup and escape should be imported from markupsafe

davidism:
These were still left over from Jinja's early days. Noticed they were still exported and documented, so now the Jinja versions show deprecation warnings pointing to markupsafe.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -609,27 +609,9 @@ functions to a Jinja environment.
 
 .. autofunction:: jinja2.environmentfunction
 
-.. function:: escape(s)
-
-    Convert the characters ``&``, ``<``, ``>``, ``'``, and ``"`` in string `s`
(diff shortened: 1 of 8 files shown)


CarliJoy:
I used this in paragraphs in #1386 to describe how and when to use ``env.get_markup_class`` 

Also I used it to refer to the Markup class as link.
The external link didn't work.
So I kindly ask if you could fix the merge conflict #1386 cause I am kind out of ideas considering this change...

CarliJoy:
The problem is also that now the explaination of what Markup does and what the `.__html__` means is out of the documentation.
And I kind of referred to it in other parts of the Jinja documentation. Now it is all a bit loose an and unconnected :-/

The supposed links to Markup (i.e. in the current [explanation of the parameter](https://jinja.palletsprojects.com/en/master/api/#jinja2.Environment) does not work.

Issue #1191, opened 2020-04-15
Require async/await usage in templates


pgjones:
The current async mode happily allows coroutines to be awaited without requiring the `await` keyword. Whilst this is nice, it is actually quite a confusing and a non-obvious behavior. In addition it doesn't always work, for example `{% if coroutine %}`, and I'm not sure `{% async for ...` or `{% async with ...` work. 

I propose that templating engine understand and require the usage of the `async` and `await` keywords dropping the current support for non `await`ed coroutines. 

It would also be nice if the engine did not use any `asyncio` features so as to be event loop agnostic.

davidism:
Based on the docs, the intention seems to be that the template designer doesn't need to understand what type of function was passed, they could continue to write templates as they always had. Whether this has actually been beneficial is impossible to measure.

The main benefit I see is a sort of re-wording of that. By not requiring the `await` keyword, you don't have to care if your templates will be rendered in an async environment or not. For example, a Flask extension could provide a blueprint with some templates that loop over a database query, and they'd continue to work even if the project using the extension was using an async query library.

I can envision making `enable_async` accept the value `"strict"` instead of just `True`. This would enable async behavior but the compiler wouldn't output the `auto_await` wrapper. Instead, a new `await` token and node would be added to tell the compiler to output the `await` keyword. This would probably be more performant since the wrapper wouldn't be called on literally everything, but would not be as portable. The compiler could skip the `await` in sync mode, but this wouldn't be completely equivalent to the current behavior, as the template author would have to decide what would and would not have to be async in async mode.

---

As to using asyncio features, I'm not clear what we'd have to do not to, but I'm open to it. It looks like all it's used for right now is to get an event loop when patching some functions.

Dreamsorcerer:
This also appears to be a blocker for async support in aiohttp-jinja2:
https://github.com/aio-libs/aiohttp-jinja2/pull/154

pgjones:
I think an additional `"strict"` mode is a better solution, as I can see the value in the original intention - although I personally would like to avoid this.

I'm also not sure if asyncio features are required, but aiming to avoid them would certainly help support other event loops.

nirvana-msu:
I'm also in favour of adding explicit/strict mode, at least as an option. Perhaps someone maintaining a library with templates might think differently, but as far as your own code where you control both front-end and back-end is concerned, being explicit with `await` calls is significantly better than being implicit.

davidism:
[...]

PR #1392, opened 2021-04-10
async support doesn't require patching

### LINKED ISSUE #1390: remove async patching system
It's no longer needed since we only support Python 3.6+.

#765 observed that `import asyncio` takes a long time, so keep those imports at function scope to avoid longer startup time.

A future task is to figure out how to remove the use of `asyncio` so that other libraries such as Trio could be used instead.

davidism:
The patching is not required, `async def` is valid in all supported Python versions. Moves the async/patched sync functions from `asyncsupport` into the regular code. Moves the async filters from `asyncfilters` to `filters`. This does mess with the signature arguments in the docs for the filters, I think that can be fixed in Pallets-Sphinx-Theme's Jinja support.

`import asyncio` is done inside `render` and `generate` to avoid importing it at the top, which would slow down import time.

- fixes #1390 

Checklist:


davidism:
@pgjones related to #1191:

`asyncio` is imported in two methods on `Environment`, to call `asyncio.get_event_loop()` and `loop.run_until_complete(coroutine)`. We can probably make those two separate, overridable methods on `Environment` in order to allow using Trio, etc.

pgjones:
As I understand it the imports are in `render` and `generate` the sync versions of `render_async` and `generate_async`. In which case I don't think this is an issue (as `trio.run(environment.render_async, ...)` is valid).

I wonder though if these sync methods should instead raise runtime/not implemented errors in an async environment. I think this would happen naturally anyway as calling `run_until_complete` from within an event loop already raises a runtime error.

PR #1402, opened 2021-04-19
Update make.bat

Assimilationstheorie:
Add correct url.


davidism:
Thanks, but I'll update the full scripts when needed, this isn't really needed.

PR #1419, opened 2021-05-11
Install sphinx 4.1-dev from git

ThiefMaster:
Otherwise docs builds are broken due to sphinx-doc/sphinx#9216

The Any type is needed to avoid this odd sphinx warning:

> jinja2/nativetypes.py:docstring of jinja2.nativetypes.NativeTemplate::more than one target found for cross-reference 'Template': jinja2.Template, jinja2.environment.Template, jinja2.nodes.Template

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -36,12 +36,10 @@ imagesize==1.2.0
     # via sphinx
 iniconfig==1.1.1
     # via pytest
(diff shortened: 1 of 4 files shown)


davidism:
Might be something to report to Sphinx, maybe they're not resolving the annotations strictly enough.

PR #1422, opened 2021-05-12
Update CHANGES.rst to reflect breaking change in 2.9

ngandhy:
Updating CHANGES.rst to reflect breaking behavior to help folks track down issues on update. 
Otherwise old projects that get updated just start failing with no clear reason from the CHANGES.rst

Nothing to test as this is just a documentation update

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -425,6 +425,7 @@ Version 2.9
 
 Released 2017-01-07, codename Derivation
 
+-   BREAKING - Behavior for variables set inside a loop have changed. See issue:`641`
 -   Change cache key definition in environment. This fixes a performance
     regression introduced in 2.8.
 -   Added support for ``generator_stop`` on supported Python versions


davidism:
Sorry, I'm not sure why this needs to be added now four years later. Since it doesn't match our log style or describe the actual issue or fix, I'm going to close this.

ngandhy:
Seems like best practice for libraries to announce or warn of breaking changes to help developers instead of wasting everyone's time tracking down unannounced changes / breaks.

You're right that pallets/jinja has never properly warned of a breaking change in the changelog, might be worth doing so moving forward for the community.

ngandhy:
@davidism - Seems like best practice for libraries to announce or warn of breaking changes to help developers instead of wasting everyone's time tracking down unannounced changes / breaks.

You're right that pallets/jinja has never properly warned of a breaking change in the changelog leaving developers on their own to track down issues as shown in multiple tickets, might be worth doing so moving forward for the community.

PR #1465, opened 2021-06-14
Add case_sensitive parameter to groupby() filter

### LINKED ISSUE #1463: Add case_sensitive parameter for groupby() filter
`groupby()` filter performs auto sorting under the hood, and it should support specifying case sensitivity for the sort order, the same way `sort()` filter does.

Current `groupby()` implementation always performs case-sensitive sorting under the hood, which makes it inflexible, and behave inconsistently with how the `sort()` filter works (case-insensitive by default). When used in the same template for the same dataset, `groupby()` and `sort()` filters produce different sort order on the same fields.

I think, I would personally prefer to not have any implicit sorting in `groupby()`, and chain it with `sort()` explicitly, but I do understand that auto sorting was, probably, added to `groupby()` for convenience.

Implementing an option to control case sensitivity seems to be pretty straightforward:

- A `case_sensitive` parameter can be added to `do_groupby()` function, similar to `do_sort()`.
- `make_attrgetter()` inside `do_groupby()` should be called with `postprocess=ignore_case` if `case_sensitive==False`.

If this feature request looks reasonable, I should be able to submit a pull request.

wombatonfire:
This is a quick implementation test to add support for case-insensitive sorting inside the `groupby()` filter, intended to verify the approach.

Changes:

- Add `case_sensitive` parameter to `sync_do_groupby()`.
- Ignore case when sorting in `sync_do_groupby()` if `case_sensitive=False`

Open questions:

1. Is this approach correct?
2. Should async groupby (`async def do_groupby`) also be changed?
3. Should the default value for `case_sensitive` parameter be `False` for backwards compatibility, or `True` to align with the `sort()` filter behavior?

- fixes #1463 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,9 @@ Unreleased
 -   Add ``items`` filter. :issue:`1561`
 -   Subscriptions (``[0]``, etc.) can be used after filters, tests, and
     calls when the environment is in async mode. :issue:`1573`
+-   The ``groupby`` filter is case-insensitive by default, matching
+    other comparison filters. Added the ``case_sensitive`` parameter to
+    control this. :issue:`1463`
 
 
 Version 3.0.3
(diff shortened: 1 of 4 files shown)


davidism:
Yeah, I'm going to make it so the first key, whatever case that is, is returned as the key in the output.

PR #1516, opened 2021-10-18
Reduce async overhead due to auto_await

### LINKED ISSUE #1514: Async is slow due to isawaitable use
I was benchmarking a render and I noticed that the use of `inspect.isawaitable` really slows things down:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    22920    0.024    0.000    0.059    0.000 .../python3.10/inspect.py:348(isawaitable)
```

Almost 60 milliseconds is spent there, more than any other Python builtin in the profile.

Environment:

- Python version: 3.10
- Jinja version: 3.0.1

mohd-akram:
The types of most rendered values tend to be primitives, so we can avoid a costly call to `inspect.isawaitable` for them. This improves performance by at least 15% in my tests.

- fixes #1514

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Unreleased
     :issue:`1535`
 -   Fix how the native environment treats leading and trailing spaces
     when parsing values on Python 3.10. :pr:`1537`
+-   Improve async performance by avoiding checks for common types.
+    :issue:`1514`
 
 
 Version 3.0.2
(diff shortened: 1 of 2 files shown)


mohd-akram:
@davidism Have you had a chance to look at this?

davidism:
No not yet. When I do, I'll leave a review, comment, or merge and you'll get a notification.

PR #1530, opened 2021-11-05
test Python 3.10

davidism:


ThiefMaster:
What about including the upcoming dev version as well (possibly with "allowed to fail" if such an option exists on github actions)?

davidism:
For now I'm just going to merge this. Specifying prereleases for setup-python is annoying because you have to write "3.11.0-alpha.1" instead of "3.11", then remember to keep updating it. It also downloads and unpacks rather than using a cached build, which takes longer. Maybe once it gets into beta or RC we can start looking at it.

PR #1334, opened 2021-01-19
Add 'linetable' to the list of preserved CodeType attributes.

thmo:
As part of https://github.com/python/cpython/pull/23113 partly
implementing PEP 626 (see https://bugs.python.org/issue42246), the
co_lnotab member of struct PyCodeObject was replaced by co_linetable,
implementing a new line number table.

This commit therefore adds linetable to the list of attributes that
are copied over into the final CodeType executed by fake_traceback()
to ensure proper line numbers and contents in fake stack traces
generated when running under Python 3.10.

Fixes #1333.

PR #1536, opened 2021-11-09
simplify CodeType rewriting

davidism:
Use `CodeType.replace` to greatly simplify replacing `co_name` on Python 3.8+. Only Python 3.7 needs to use all positional arguments. No longer need to replace `co_filename`, it's already replaced by `compile`. Remove `try/except` block to avoid hiding errors with this code in the future; need to test that modern GAE works with this.

fixes #1334

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,14 @@
 .. currentmodule:: jinja2
 
+Version 3.0.3
+-------------
+
+Unreleased
+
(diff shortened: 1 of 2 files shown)


davidism:
From what I can see from https://cloud.google.com/appengine/docs/standard/runtimes and other docs, the modern GAE runtime is unmodified, unlike the legacy Python 2 runtime. Based on that, I'm going to merge this. If a bug report comes up during the RC phase or later, we can always add a more specific check in a bugfix release.

Issue #1578, opened 2022-02-03
Document that `get_template()` template names should use `/` even on Windows


simonw:
I got caught out by this recently. The only mention I could find of this was in this comment: https://github.com/pallets/jinja/issues/1378#issuecomment-812410922

> Jinja template names/paths are not always filesystem paths. So regardless of the OS Jinja always uses forward slashes.

simonw:
I think the best place to document it is here: https://github.com/pallets/jinja/blob/077b7918a7642ff6742fe48a32e54d7875140894/src/jinja2/environment.py#L965-L976

Issue #1572, opened 2022-01-15
Content outside of block affected by whitespace


BeyondEvil:
Switching off new lines using `-` in the block, affects the content _after_ the block.

Use below example to repro:

`main.py`:
```python
from jinja2 import Environment, FileSystemLoader


rendered_report = (
    Environment(loader=FileSystemLoader("."))
    .get_template("index.jinja2")
    .render(
        scripts=["script1.js", "script2.js"],
    )
)

with open("report.html", "w", encoding="utf-8") as f:
    f.write(rendered_report)
```

`index.jinja2`:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8"/>
    <title>Test Report</title>
  </head>
  <body onLoad="init()">
    <h1 id="title">Test Report</h1>
  </body>
  <footer>
    {% for script in scripts -%}
    <script src="scripts/{{ script }}"></script>
    {% endfor -%}
  </footer>
</html>
```

Expected output:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8"/>
    <title>Test Report</title>
  </head>
  <body onLoad="init()">
    <h1 id="title">Test Report</h1>
  </body>
  <footer>
    <script src="scripts/script1.js"></script>
    <script src="scripts/script2.js"></script>
  </footer>
</html>
```

Actual output (`</footer>` is incorrectly indented):
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8"/>
    <title>Test Report</title>
  </head>
  <body onLoad="init()">
    <h1 id="title">Test Report</h1>
  </body>
  <footer>
    <script src="scripts/script1.js"></script>
    <script src="scripts/script2.js"></script>
    </footer>
</html>
```

Environment:

- Python version: 3.9.9
- Jinja version: 3.0.3

What am I doing wrong?

davidism:
This is intended.

BeyondEvil:
> This is intended.

Umm, ok. Mind sharing the rationale or if there's a way to keep the correct indentation?

davidism:
Remove the dash from the endfor tag.

BeyondEvil:
> Remove the dash from the endfor tag.

Thanks, but if I do that, I get a trailing new line:
```html
  <footer>
    <script src="scripts/script1.js"></script>
    <script src="scripts/script2.js"></script>
    
  </footer>
```

davidism:
[...]

PR #1579, opened 2022-02-03
get_template() uses / separators even on Windows

simonw:
Refs #1578.


simonw:
Oops, bad commit message on that commit - it should have said `refs #1578` not `refs #1572`.

davidism:
I think something like "When loading templates from the filesystem, only "/" is used as the path separator, even on Windows."

User confusion seems to come from the Jinja can load templates from lots of places, there is no requirement for template names to be paths, they only happen to look like paths when using the filesystem loader.

simonw:
Yes that's much better.

Issue #1605, opened 2022-03-01
2.11.3 broken due to outdated dependency


3nprob:
One of our existing applications still on `2.11.3` will no longer run on a fresh install due to use of removed function `soft_unicode`, which was removed from [`MarkupSafe v2.1.0`](https://github.com/pallets/markupsafe/blob/main/CHANGES.rst#version-210).

Could be solved either by:
  * Inserting dependency restriction for `MarkupSafe<2.10`
  * Backporting the appropriate set of changes from 148a19138c197233cbd3a2cb1f543d468479f1bc (changing `soft_unicode` to `soft_str`)

As flask v1.1.4 (released May 2021) still depends on the jinja 2.x branch, I suspect there are many projects hit by this.

cf https://github.com/pallets/flask/issues/4455

```
  File "/usr/local/lib/python3.10/site-packages/mypackage/web/web.py", line 26, in <module>
    import flask
  File "/usr/local/lib/python3.10/site-packages/flask/__init__.py", line 14, in <module>
    from jinja2 import escape
  File "/usr/local/lib/python3.10/site-packages/jinja2/__init__.py", line 12, in <module>
    from .environment import Environment
  File "/usr/local/lib/python3.10/site-packages/jinja2/environment.py", line 25, in <module>
    from .defaults import BLOCK_END_STRING
  File "/usr/local/lib/python3.10/site-packages/jinja2/defaults.py", line 3, in <module>
    from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
  File "/usr/local/lib/python3.10/site-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
```

Environment:

- Python version: 3.10.2
- Jinja version: 2.11.3

ThiefMaster:
You need to pin all your dependencies, not just the direct ones... See https://github.com/pallets/jinja/issues/1585

3nprob:
@ThiefMaster Sure - but given that https://github.com/pallets/jinja/blob/bb7f3a51091d91010deb1e01e69f4f255d3b3eb6/setup.py#L6 pulls in an incompatible version, wouldn't it be prudent to at least release a jinja 1.1.5 which restricts the version further?

Arguably this is an issue of MarkupSafe breaking semver and releasing a breaking change in a minor version bump - but as-is, 2.11.3 will not work independently without changes.

ThiefMaster:
[...]

Issue #1601, opened 2022-02-22
ImportError: cannot import name 'soft_unicode' from 'markupsafe'


diolk49:
Error when trying to install jinja : 
`   File "/usr/local/lib/python3.9/site-packages/jinja2/filters.py", line 13, in <module>
     from markupsafe import soft_unicode
 ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/usr/local/lib/python3.9/site-packages/markupsafe/__init__.py)`

With new version of markup safe (2.1.0), `soft_unicode` has been removed (it was previously deprecated). We need to "Use `soft_str` instead" : https://markupsafe.palletsprojects.com/en/2.1.x/changes/#version-2-1-0

Environment:

- Python version: 3.9
- Jinja version: ^2.11.3

ThiefMaster:
https://github.com/pallets/jinja/issues/1585

Issue #1600, opened 2022-02-21
markupsafe version 2.1.0 removed 'soft_unicode' function


vikingUnet:
Hello!
Today we have a big problem with installing requirement dependency as markupsafe >= 0.23
https://github.com/pallets/jinja/blob/737a4cd41d09878e7e6c584a2062f5853dc30150/setup.py#L53
and we forced to installversion markupsafe==2.0.1 in handly mode.
Please, change this dependency for backward compatibility,
problam describe at this link https://githubhot.com/repo/pallets/markupsafe/issues/284
Thanks!!!

ThiefMaster:
https://github.com/pallets/jinja/issues/1585

davidism:
Duplicate of #1585 

You are using an unsupported version of Jinja, please update to the latest version if possible. Additionally, please read https://hynek.me/articles/semver-will-not-save-you/, then use a tool like [pip-tools](https://pypi.org/project/pip-tools/) to pin your dependencies and control when you get updates. Be sure to run your tests with deprecation warnings treated as errors so that you get notified of these types of changes early.

Issue #1599, opened 2022-02-19
Jinja 2.11 broken against markupsafe==2.1.0


RouquinBlanc:
Environment:

- Python version: 3.9
- Jinja version: 2.11.3

Hi,

Yesterday, `markupsafe` was bumped to 2.1.0 but it breaks jinja `2.11.3` because of the use of ``soft_unicode`` which was just removed...

Is there a plan to continue supporting the 2.11 branch for a while still? In which case it would be nice to either:

* limit the version of markupsafe to be used to pre-2.1.0
* or on the contrary, fix the use of that function toward ``soft_str``?

I've seen this on some of our legacy code, but I'm pretty sure many people have legacy code and would rather not update in a rush 😅 

```
$ python
Python 3.9.5 (default, May 25 2021, 18:01:27)
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from jinja2 import Environment
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jmartin/workspace/nidar-apps/apps/nidar-l3-proxy/venv/lib/python3.9/site-packages/jinja2/__init__.py", line 12, in <module>
    from .environment import Environment
  File "/Users/jmartin/workspace/nidar-apps/apps/nidar-l3-proxy/venv/lib/python3.9/site-packages/jinja2/environment.py", line 25, in <module>
    from .defaults import BLOCK_END_STRING
  File "/Users/jmartin/workspace/nidar-apps/apps/nidar-l3-proxy/venv/lib/python3.9/site-packages/jinja2/defaults.py", line 3, in <module>
    from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
  File "/Users/jmartin/workspace/nidar-apps/apps/nidar-l3-proxy/venv/lib/python3.9/site-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/Users/jmartin/workspace/nidar-apps/apps/nidar-l3-proxy/venv/lib/python3.9/site-packages/markupsafe/__init__.py)
```

Thanks

ThiefMaster:
You may want to look at the pinned issue (or large amount of closed issues) before opening your own in the future ;)
https://github.com/pallets/jinja/issues/1585

TL;DR: Don't use an outdated Jinja version...

[...]

PR #1606, opened 2022-03-01
[2.11.x] Restrict MarkupSafe < 2.1

3nprob:
#1605 #1585 #1601 #1600 #1599 #1598 #1595 #1596 #1597 #1594 #1590 #1591 #1587 #1585

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ setup(
     include_package_data=True,
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
     install_requires=["MarkupSafe>=0.23"],
+    install_requires=["MarkupSafe>=0.23,<2.1"],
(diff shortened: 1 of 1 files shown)


3nprob:
CI seems to assume `master` as base - this should be mergable without conflicts

ThiefMaster:
It's an old and unsupported version... we do not plan to make a release adding upper bounds; people would have to update their pin anyway, so they can simply add the pin on markupsafe instead..

3nprob:
Those who did not add an explicit pin (very much like jinja itself) have had things not breaking without intervention. Those are now failing. Publishing this change will make it work againa without depending projects adding explicit pins.

ThiefMaster:
People who did not pin Jinja would have gotten updates for both Jinja and Markupsafe -> no problem. So the only people are affected are these where *only* some of the deps are pinned...

PR #1613, opened 2022-03-03
Fix pickleability of filters decorated with @async_variant

### LINKED ISSUE #1612: Filters decorated with @async_variant are not pickleable
The [`jinja2.async_utils.async_variant`][async_variant] decorator puts the wrong `__name__` and `__qualname__` on the decorated function.  One result of this is that functions which it decorates are not pickleable.

Here's a doctest script which illustrates the issue.
```pycon
>>> import pickle
>>> import jinja2.filters

>>> pickle.dumps(jinja2.filters.do_first)
    ...
_pickle.PicklingError: Can't pickle <function sync_do_first at 0x...>: it's not the same object as jinja2.filters.sync_do_first

# Here's the crux:
>>> jinja2.filters.do_first.__name__
'sync_do_first'
```

[async_variant]: https://github.com/pallets/jinja/blob/09907ef37891be76ab0a1a3dc1a3f7df667db959/src/jinja2/async_utils.py#L11

Environment:

- Python version: 3.10 — but I don't think that's important
- Jinja version: git main (09907ef37891be76ab0a1a3dc1a3f7df667db959) — but I think this affects all versions >= 3

I will submit a PR with one possible fix shortly.

dairiki:
- fixes #1612 

Checklist:

- [n/a] Add or update relevant docs, in the docs folder and in code.
- [n/a] Add `.. versionchanged::` entries in any relevant code docs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,8 @@ Unreleased
     ``{% trans "title" %}``. :issue:`1430`
 -   Update valid identifier characters from Python 3.6 to 3.7.
     :pr:`1571`
+-   Filters and tests decorated with ``@async_variant`` are pickleable.
+    :pr:`1612`
 
 
 Version 3.0.3
--- a/src/jinja2/async_utils.py
+++ b/src/jinja2/async_utils.py
@@ -1,5 +1,6 @@
 import inspect
 import typing as t
+from functools import WRAPPER_ASSIGNMENTS
 from functools import wraps
 
 from .utils import _PassArg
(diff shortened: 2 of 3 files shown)


dairiki:
A more intrusive but cleaner way to fix the `async_variant` decorator would be the swap the order of decoration.

The decorator is used, really, to wrap two functions — the async and sync variants.  However, for ... reasons ... we want most of the dunder attributes of the wrapper to be copied from (and the wrapper's `__wrapped__` attribute to point to) the non-async function.  That being the case, it might be cleaner to have the net syntax be that we "decorate" the non-async version rather than decorating the async version as is done now.

I.e, the current API is:
```py
def sync_do_foo(...):
    "Docstring which gets copied to decorated async version..."
    ...

@async_variant(sync_do_foo)  # type: ignore
async def do_foo(...):
    # async version of sync_do_foo
    ...
```

Despite the fact that `do_foo` is the decorated function here, the under attributes of the wrapped `do_foo` are copied from `sync_do_foo`.  That seems confusing.

Instead, what about

```py
async def async_do_foo(...):
    # async version of sync_do_foo
    ...

@has_async_variant(async_do_foo)
def do_foo(...):
    """The non-async version.  Decorated function gets this docstring."""
    ...
```

(Another option, copying the way [property setters](https://docs.python.org/3/library/functions.html?highlight=property#property) can be decorated might be)

```py
@has_async_variant2
def do_foo(...):
    """Non-async version.  Wrapper gets this docstring."""
    ...

@do_foo.async_version
async def do_foo(...):
    ...
```

Is `async_variant` a public-enough API that changing it requires deprecating it, but leaving it in place?  If so, maybe not worth the effort.  If not, then...

PR #1663, opened 2022-05-02
`FileSystemLoader` include paths in error

### LINKED ISSUE #1661: Add searchpath to TemplateNotFound exception message
This was from https://github.com/apache/airflow/issues/23333 where trying to load a template that doesn't exist in the searchpath raises `TemplateNotFound` exception. In this case an absolute path is used which causes issue over concatenation but including `searchpath` might be a useful addition so that user can know under the paths under which the template was searched for and will be helpful during debugging.

https://github.com/pallets/jinja/blob/6f79daafe868b8ef20baecdcb45d12299529ed2e/src/jinja2/loaders.py#L218

Current exception :

```
jinja2.exceptions.TemplateNotFound: /tmp/a/python_template_exts.txt
```

Proposed exception to include searchpath :

```
jinja2.exceptions.TemplateNotFound: /tmp/a/python_template_exts.txt not found in search path ['/opt/airflow/tests/operators', '/']
```

Yourun-proger:
- fixes #1661

Checklist:

### DIFF
--- a/src/jinja2/loaders.py
+++ b/src/jinja2/loaders.py
@@ -204,7 +204,10 @@ class FileSystemLoader(BaseLoader):
             if os.path.isfile(filename):
                 break
         else:
-            raise TemplateNotFound(template)
+            raise TemplateNotFound(
+                f"{template} not found in the following search path(s):"
+                f" {self.searchpath}"
+            )
 
         with open(filename, encoding=self.encoding) as f:
             contents = f.read()
--- a/src/jinja2/loaders.py
+++ b/src/jinja2/loaders.py
@@ -204,9 +204,11 @@ class FileSystemLoader(BaseLoader):
             if os.path.isfile(filename):
                 break
         else:
+            plural = "path" if len(self.searchpath) == 1 else "paths"
+            paths_str = ", ".join(repr(p) for p in self.searchpath)
             raise TemplateNotFound(
-                f"{template} not found in the following search path(s):"
-                f" {self.searchpath}"
+                template,
+                f"{template!r} not found in search {plural}: {paths_str}",
             )
 
         with open(filename, encoding=self.encoding) as f:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,8 +35,8 @@ Unreleased
 -   Fix `copy`/`pickle` support for the internal ``missing`` object.
     :issue:`2027`
 -   ``Environment.overlay(enable_async)`` is applied correctly. :pr:`2061`
--   Paths where the loader searched for the template were added
-    to the error message. :issue:`1661`
+-   The error message from ``FileSystemLoader`` includes the paths that were
+    searched. :issue:`1661`
 
 
 Version 3.1.4
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,8 @@ Unreleased
 -   Fix `copy`/`pickle` support for the internal ``missing`` object.
     :issue:`2027`
 -   ``Environment.overlay(enable_async)`` is applied correctly. :pr:`2061`
+-   Paths where the loader searched for the template were added
+    to the error message. :issue:`1661`
 
 
 Version 3.1.4
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -179,6 +179,24 @@ class TestFileSystemLoader:
         t = e.get_template("foo/test.html")
         assert t.filename == str(self.searchpath / "foo" / "test.html")
 
+    def test_error_includes_paths(self, env, filesystem_loader):
+        env.loader = filesystem_loader
+
+        with pytest.raises(TemplateNotFound) as info:
+            env.get_template("missing")
+
+        e_str = str(info.value)
+        assert e_str.startswith("'missing' not found in search path: ")
+
+        filesystem_loader.searchpath.append("other")
+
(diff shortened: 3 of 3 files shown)


ThiefMaster:
I don't find this particularly useful to be honest. Especially in a larger Flask app with many blueprints (which IIRC. all contribute to the search path) this would make the error message extremely bloated with information that's not particularly useful in most cases...

src/jinja2/loaders.py
```diff
@@ -215,7 +215,9 @@ def uptodate() -> bool:
 
             # Use normpath to convert Windows altsep to sep.
             return contents, os.path.normpath(filename), uptodate
-        raise TemplateNotFound(template)
+        raise TemplateNotFound(
+            f"{template} not found in the following search path(s): {self.searchpath}"
```
Yourun-proger, line 219: I don't know, maybe add a starred expression here

Yourun-proger:
Probably. I saw now that in airflow has already come up with a patch. Maybe the truth is that everyone should change on their own, not at the library level.
In short, I can close this.

Yourun-proger:
Well, a parameter like *ignore* or *verbose* is definitely stupid....

uranusjr:
A middle ground for this would be to include `searchpath` on the exception object as an attribute, but not shown in the error message. This should be good enough for most usages; they can simply do

```python
try:
    template = get_template(filename)
except TemplateNotFound as e:
    raise MyCustomException(f"{filename} not found in {e.searchpath}")
``

instead of having to dig through the template loaders to find the context.

**Comment by Yourun-proger** [comment]
It's me again. I made a patch that looks like this:
```diff
diff --git a/src/jinja2/exceptions.py b/src/jinja2/exceptions.py
index 082ebe8..6031bee 100644
--- a/src/jinja2/exceptions.py
+++ b/src/jinja2/exceptions.py
@@ -1,4 +1,5 @@
 import typing as t
+import os

 if t.TYPE_CHECKING:
     from .runtime import Undefined
@@ -18,6 +19,9 @@ class TemplateError(Exception):
 class TemplateNotFound(IOError, LookupError, TemplateError):
     """Raised if a template does not exist.

+    .. versionchanged:: 3.1.3
+        Added new parameter "search path"
+
     .. versionchanged:: 2.11
         If the given name is :class:`Undefined` and no message was
         provided, an :exc:`UndefinedError` is raised.
@@ -31,6 +35,7 @@ class TemplateNotFound(IOError, LookupError, TemplateError):
         self,
         name: t.Optional[t.Union[str, "Undefined"]],
         message: t.Optional[str] = None,
+        searchpath: t.Optional[t.Union[str, os.PathLike, t.Sequence[t.Union[str, os.PathLike]]]] = None,
     ) -> None:
         IOError.__init__(self, name)

@@ -45,6 +50,8 @@ class TemplateNotFound(IOError, LookupError, TemplateError):
         self.message = message
         self.name = name
         self.templates = [name]
+        # Transforms of searchpath are not required here.
+        self.searchpath = searchpath

     def __str__(self) -> str:
         return str(self.message)
@@ -55,6 +62,9 @@ class TemplatesNotFound(TemplateNotFound):
     are selected.  This is a subclass of :class:`TemplateNotFound`
     exception, so just catching the base exception will catch both.

+    .. versionchanged:: 3.1.3
+        Added new parameter "search path"
+
     .. versionchanged:: 2.11
         If a name in the list of names is :class:`Undefined`, a message
         about it being undefined is shown rather than the empty string.
@@ -66,6 +76,7 @@ class TemplatesNotFound(TemplateNotFound):
         self,
         names: t.Sequence[t.Union[str, "Undefined"]] = (),
         message: t.Optional[str] = None,
+        searchpath: t.Optional[t.Union[str, os.PathLike, t.Sequence[t.Union[str, os.PathLike]]]] = None,
     ) -> None:
         if message is None:
             from .runtime import Undefined
```
and
```diff
diff --git a/src/jinja2/loaders.py b/src/jinja2/loaders.py
index d2c39c1..e398c2d 100644
--- a/src/jinja2/loaders.py
+++ b/src/jinja2/loaders.py
@@ -215,9 +215,7 @@ class FileSystemLoader(BaseLoader):

             # Use normpath to convert Windows altsep to sep.
             return contents, os.path.normpath(filename), uptodate
-        raise TemplateNotFound(
-            f"{template} not found in the following search path(s): {self.searchpath}"
-        )
+        raise TemplateNotFound(template, searchpath=self.searchpath)

     def list_templates(self) -> t.List[str]:
         found = set()
```
In principle, now you can achieve the desired behavior and do something like:
```py
try:
    template = my_file_system_loader.get_source(some_env, filename)
except TemplateNotFound as e:
    raise MyCustomException(f"{filename} not found in {e.searchpath}")
```
As you may have understood, I now need to add the keyword argument `searchpath` wherever `TemplateNotFound` and `TemplatesNotFound` are raised, which is feasible in principle, but seems like some kind of bad decision? What do you think about this? Can you advise me something?

davidism:
Flask won't be affected by this, as it raises a simple `TemplateNotFound(name)` rather than any particular error message from the combined loaders. I don't want to add special attributes to the errors, as `searchpath` would only be relevant to `FileSystemLoader`. There's another PR for improving the `PackageLoader` error message as well, and `TemplateNotFound` already includes a way to specify the template name and a detailed message, so this seems fine.

PR #1717, opened 2022-09-07
Fix custom filter example

bzeba:
Filter name used in templates must match a key in the filters map.


davidism:
Fix the name in the template instead.

bzeba:
@davidism Forgot to ping you originally. It should be as requested now.

davidism:
fixed by #1911, sorry saw that one first even though this was made first, thanks for looking into it

PR #1723, opened 2022-09-19
Fix pow and unary operators order

### LINKED ISSUE #1720: Negative number raised to even power gives negative result if exponent is a variable
Unexpected behavior of the power operator with a variable of even value as exponent. It returns the negative value of the expected result.

## mcve

```python

env = Environment()
env.globals = {"x": 2}
env.from_string("{{ -1 ** x }}").render()
```
From which I get `'-1'`

## Expected behavior
I expected the positive value `'1'` as a result

It instead works as expected in a Sandboxed environment when intercepting the `**` operator:

```python
from jinja2.sandbox import SandboxedEnvironment

sandbox_env = SandboxedEnvironment()
sandbox_env.globals = {"x": 2}
sandbox_env.intercepted_binops = frozenset(["**"])
sandbox_env.from_string("{{ -1 ** x }}").render()
```
From which I get `'1'`

## Environment:

- Python version: 3.8.10
- Jinja version: 3.1.2

### LINKED ISSUE #1722: Bug with right-left associativity of exponent
Discovered it when working on #1720. Looks like the fix will fix both issues.
### Minimal reproducible example
```python
env = Environment()
tmpl = env.from_string('{{ 2 ** 3 ** 2 }}')
assert eval('2 ** 3 ** 2') == 512 
assert tmpl.render() == '512' # it's actualy 64 because of wrong order
```

### Expected behaviour 
I expected exact behaviour as in [pure python](https://docs.python.org/3/reference/expressions.html#evaluation-order). We need to start evaluation from left to right.

Environment:

- Python version: 3.9.12
- Jinja version: 3.1.x

krepysh:
Change order of evaluating expressions with pow by Parser.

- fixes #1720 
- fixes #1722 

Checklist:


davidism:
The changelog entry should be much clearer on what is changing since this affects how these expressions are evaluated.

bart-nouveau:
There is still one bug pointed out in #1720 that would not be fixed by this pull request:

expression: `'(-1) ** x'`
jinja_output: `'-1'`
python_output: `1`

## minimal reproducible example
```python
env = Environment()
expression = '(-1) ** x'
env.globals = {"x": 2}
jinja_output = env.from_string(f"{{{{ {expression} }}}}").render()
python_output = str(eval(f"{expression}",env.globals))
assert python_output == jinja_output
```

## suggested fix

One possible fix could be to modify the `_make_binop` function [here](https://github.com/pallets/jinja/blob/9fde7eb82021da0bb016e1990753afd6b2d39386/src/jinja2/compiler.py#L71) as follows:

```python
            self.write("(")
            self.visit(node.left, frame)
            self.write(f" {op} ")
            self.visit(node.right, frame)
```
changed into:

```python
            self.write("((")
            self.visit(node.left, frame)
            self.write(")")
            self.write(f" {op} ")
            self.visit(node.right, frame)
```

tests/test_lexnparse.py
```diff
@@ -575,6 +575,14 @@ def test_parse_unary(self, env):
...
         assert tmpl.render(foo={"bar": 42}) == "42"
 
+    def test_negative_pow(self, env):
+        tmpl = env.from_string("{{ - 1 ** 2 }}")
+        assert tmpl.render() == "-1"
+
+    def test_pow_associative_from_left_to_right(self, env):
+        tmpl = env.from_string("{{ 2 ** 3 ** 2 }}")
+        assert tmpl.render() == "512"
+
```
bart-nouveau, line 585: In order to verify if issue #1720 is fixed, I would add the following tests:

```python
    def test_pow_negative_base(self, env):
        tmpl = env.from_string("{{ (- 1) ** 2 }}")
        assert tmpl.render() == "1"

    def test_pow_negative_base_variable_exponent(self, env):
        tmpl = env.from_string("{{ (- 1) ** x }}", {"x": 2})
        assert tmpl.render() == "1"

    def test_negative_pow_variable_exponent(self, env):
        tmpl = env.from_string("{{ - 1 ** x }}", {"x": 2})
        assert tmpl.render() == "-1"
```

docs/templates.rst
```diff
@@ -1369,7 +1369,10 @@ but exists for completeness' sake.  The following operators are supported:
 
 ``**``
     Raise the left operand to the power of the right operand.
-    ``{{ 2**3 }}`` would return ``8``.
+    ``{{ 2**3 }}`` would return ``8``
+
+    .. versionchanged:: 3.2
+        From 3.2, this operator has the same evaluation order as in Python. ``{{ 2 ** 3 ** 2 }}`` will be evaluated as ``2 ** (3 ** 2)``.
 
     Unlike Python, chained pow is evaluated left to right.
```
krepysh, line 1377: @davidism, should I remove this outdated explanation? Or it will be better to leave it here in order to keep history in docs?

bart-nouveau (commented):
It's ok for me, it fixes issues :issue:`1720` :issue:`1722

PR #1735, opened 2022-10-17
Document how to make dynamically an include

jetm:
The document is not clear how to make dynamically an include. Add
examples for clarification.


davidism:
Neither of these examples are correct, and the line above already says that the target can be a variable with a name or template object.

PR #1753, opened 2022-11-04
Ensure that name and filename are strings

invisibleroads:
Prevent "name 'PosixPath' is not defined" when the argument to `get_template` is a `pathlib.Path` instance.

```python
from pathlib import Path
env.get_template(Path('t.html'))
```

`f'{name!r}'` compiles to

```python
name = PosixPath('t.html')
```

which results in the exception "name 'PosixPath' is not defined"

### DIFF
--- a/src/jinja2/compiler.py
+++ b/src/jinja2/compiler.py
@@ -310,8 +310,8 @@ class CodeGenerator(NodeVisitor):
         if stream is None:
             stream = StringIO()
         self.environment = environment
-        self.name = name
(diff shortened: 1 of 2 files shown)


invisibleroads:
I moved the type coercion to `BaseLoader` in case the bytecode cache needs strings for unique keys.

invisibleroads:
Fixes https://github.com/pallets/jinja/issues/1754

davidism:
See closure reason on #1754.

Issue #1752, opened 2022-11-04
Content of if-else block is computed even if condition is false


Louis-DR:
Hello,

I've noticed filters inside an if-else loop are triggered regardless of the condition result. However I have two issues with this :

1. We spend resources and performance rendering all blocks, which in certain cases can increate the total rendering time by a lot.
2. When using extensions or filters that don't just return text, they are executed regardless.

For instance, when using the following custom filters

``` python
import time

def log(txt):
  print(txt)
  return txt

def sleep(txt):
  time.sleep(5)
  return txt
```

And this template

``` jinja2
{% set condition = true %}

{% if condition %}
{{ "Hello" | log   }}
{{ ""      | sleep }}
{% else %}
{{ "World" | log   }}
{{ ""      | sleep }}
{% endif %}
```

The resulting rendered file will only contain `Hello`, but rendering will take 10 seconds (instead of only 5), and the console will print `Hello` and `World`.

In my case I use custom filters to generate warnings or errors when checking the configuration in the template, and also I have some special filters that write/append to other files and change environment settings/variables.

I think this could also cause issues with the `do` expression.

Is this expected behaviour or perhaps something I configured incorrectly ?

Python version 3.10.7
Jinja2 version 3.1.2

Louis-DR:
I've investigated this issue more, and I have to admit I don't understand how Jinja2 parses a template. I've tried disabling optimization (`optimized=False`).

With the following template :

``` jinja2
{% set condition = true %}

{% set ns = namespace(foo="✔one") %}
{{ ("✔1 Should see this at the start : "~ns.foo) | log }}
{{  "✔2 Should see this" | log }}
{{ ns.foo | log }}

{% if condition %}
  {{ "✔3 Should see this" | log }}
  {% set ns.foo = "✔two" | log %}
{% else %}
  {{ "✘1 Should not see this !" | log }}
  {% set ns.foo = "✘three" | log %}
{% endif %}

{% if not condition %}
  {% set ns.foo = "✘four" | log %}
  {{ ns.foo | log }}
  {{ "✘2 Should not see this either !" | log }}
{% endif %}

{{ ("✔4 Should see this : "~ns.foo) | log }}
{{  "✔5 Should see this at the end" | log }}
```

With optimization, the console prints :
```
✔2 Should see this
✔3 Should see this
✔two
✘1 Should not see this !
✘three
✘four
✘2 Should not see this either !
✔5 Should see this at the end
✔1 Should see this at the start : ✔one
[...]

PR #1759, opened 2022-11-11
Render time only filter decorator

Louis-DR:
Added the `@render_time_only` decorator for filters and tests. It marks filters to not be resolved at compile time even if the arguments are constants. This can be useful as described in #1752 for time or memory intensive filters and tests, filters with global counting variables and changing global states, or filters that cause external actions such as printing, file operations or server requests.

The feature is tested by a new pytest, documented in changelog and the API documentation, passes pytest and tox (with a small tweak because mypy doesn't register set attributes).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Version 3.2.0
 
 Unreleased
 
+-   Add function decorator ``@render_time_only`` for filters and tests.
+    :issue:`1752`
+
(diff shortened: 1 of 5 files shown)


Louis-DR:
Relating to issue #1752, @davidism can you review this MR ? The checks are good, the changelog and documentation are also aligned. Thanks

PR #1764, opened 2022-11-17
ext.i18n policy to fallback to untranslated string when interpolation fails

### LINKED ISSUE #1763: Configurable behavior when translated string interpolation fails.
Similar to #421 when translation strings are not kept perfectly up to date, changes to the names of variables in strings can break behavior on a site. I'd like to see some way to manage this behavior via a policy similar to the `ext.i18n.trimmed` that would fallback to the untranslated string in scenarios where KeyError is raised.

PyPI.org's codebase (pypi/warehouse) is translated by volunteers, so occasionally when we change our templates to use new or different variable names translations fall out of sync and raise errors when trying to render them.

Our work around is pretty fragile: https://github.com/pypi/warehouse/commit/94d9503551002fd7806bb27a675c80e747513841#diff-8125a22a065160a5ac02ff19f3dc404a6d92ecfdf31b496dc9446120f59eb6ff

ewdurbin:
Implements `ext.i18n.newstyle_fallback_interpolation` policy which can be set to `True` to gracefully fallback to untranslated strings if an interpolation error occurs due to incorrectly or out-of-date named variables in translations.

- fixes #1763 

Checklist:

### DIFF
--- a/src/jinja2/defaults.py
+++ b/src/jinja2/defaults.py
@@ -45,4 +45,5 @@ DEFAULT_POLICIES: t.Dict[str, t.Any] = {
     "json.dumps_function": None,
     "json.dumps_kwargs": {"sort_keys": True},
     "ext.i18n.trimmed": False,
+    "ext.i18n.newstyle_fallback_interpolation": False,
 }
--- a/src/jinja2/ext.py
+++ b/src/jinja2/ext.py
@@ -167,21 +167,30 @@ def _gettext_alias(
     return __context.call(__context.resolve("gettext"), *args, **kwargs)
 
 
-def _make_new_gettext(func: t.Callable[[str], str]) -> t.Callable[..., str]:
+def _make_new_gettext(
+    func: t.Callable[[str], str], newstyle_fallback_interpolation: bool
+) -> t.Callable[..., str]:
     @pass_context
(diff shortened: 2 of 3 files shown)


ewdurbin:
If this is even remotely in a direction acceptable to maintainers, I'm happy to address the remaining items to make it mergeable.

davidism:
When I was looking at this earlier, I expected you'd be able to either pass custom functions to `env.install_gettext_callables` or `env.install_gettext_translations`. But it looks like the problem is that this only matters for the "new style" wrappers, so overriding the functions being wrapped isn't going to be able to get at the error that the fallback needs.

I think this fallback makes a lot of sense. It seems like it would almost always be preferable to show/format the original string rather than failing. The downside is that you no longer get errors, so it's harder to catch and fix the bad translation. Maybe this shouldn't be a configuration, and should show a warning when substitution fails.

ewdurbin:
Thanks for considering this @davidism! So you're suggesting it might be better to make this the default behavior and fire a [warning](https://docs.python.org/3/library/warnings.html) when the fallback occurs? That certainly simplifies things from my decision to implement it as a policy so that it wouldn't change without opt in. Biased, since this behavior is in line with my concern, but I would happily update this PR to that end.

PR #1776, opened 2022-12-15
add an alternative method for getting a template's dependencies

### LINKED ISSUE #1775: Support accurate dependency tracking including dynamic inheritance or inclusion
`jinja2.meta.find_referenced_templates` does not support dynamic inheritance or inclusion.

infinity0:
Add `Environment.extract_parsed_names` to support tracking dynamic inheritance or inclusion.

Compared to `jinja2.meta.find_referenced_templates()`, it:

a. works on dynamic inheritance and includes
b. does not work unless and until you actually render the template

Many buildsystems are unable to support (b), but some do e.g. [1], the point being that if the target file does not exist, dependency information is not needed since the target file must be built anyway. In such cases, you may prefer this function due to (a).

[1] https://make.mad-scientist.net/papers/advanced-auto-dependency-generation/

- fixes #1775

Checklist:

- [X] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [X] Add `.. versionchanged::` entries in any relevant code docs.


infinity0:
pre-commit.ci is complaining about these two files even though I didn't touch them in this PR:

~~~~
src/jinja2/nodes.py:138:30: B905 `zip()` without an explicit `strict=` parameter.
src/jinja2/compiler.py:1350:29: B905 `zip()` without an explicit `strict=` parameter.
~~~~

PR #1807, opened 2023-02-19
Additional test coverage for async rendering of native type templates

### LINKED ISSUE #1806: Missing test coverage from pull request #1511?
As part of pull request #1511, the changes may have fixed a bug related to native type rendering for iterator; it could be worth adding test coverage related to that.

Before #1511, the code was calling `itertools.islice(values, 2)` to retrieve the head elements from a `values` collection (potentially of an iterator type) and then using `itertools.chain(head, values)` to subsequently produce output.

In the case where `values` _is_ an iterator, I think that works correctly.  But in cases where `values` is _not_ an iterator, it can result in output duplication (something that I think #1511 indirectly resolves with [this check](https://github.com/pallets/jinja/pull/1511/commits/2903565262e6c8971e612a2717feb430e284e4ac#diff-25a3ba2093b8517416a58a44407a2580496610a92a90b6d6550da54ce4f3220cR35-R37)).

For example:

```pycon
>>> from itertools import islice
>>> values = [1,2,3,4]
>>> head = islice(values, 2)
>>> from itertools import chain
>>> list(chain(head, values))
[1, 2, 1, 2, 3, 4]
```

jayaddison:
- fixes #1806

Checklist:

- [ ] ~~Add or update relevant docs, in the docs folder and in code.~~
- [ ] ~~Add `.. versionchanged::` entries in any relevant code docs.~~
- [x] Run `pytest` ~~and `tox`~~, no tests failed.
  - `PYTHONPATH=src pytest tests/test_nativetypes.py`


tests/test_nativetypes.py
```diff
@@ -122,6 +128,18 @@ def test_string_top_level(env):
...
+def test_string_concatenation(async_native_env):
+    async def async_render():
+        t = async_native_env.from_string(
+            "{%- macro x(y) -%}{{ y }}{%- endmacro -%}{{- x('not') }} {{ x('bad') -}}"
+        )
+        result = await t.render_async()
+        assert isinstance(result, str)
+        assert result == "not bad"
+
+    asyncio.run(async_render())
```
graingert, line 140: this should use

```python
def _asyncio_run(async_fn, *args):
    return asyncio.run(async_fn(*args))


@pytest.fixture(params=[_asyncio_run, trio.run], ids=["asyncio", "trio"])
def run_async_fn(request):
    return request.param


def test_string_concatenation(async_native_env, run_async_fn):
    async def async_render():
        t = async_native_env.from_string(
            "{%- macro x(y) -%}{{ y }}{%- endmacro -%}{{- x('not') }} {{ x('bad') -}}"
        )
        result = await t.render_async()
        assert isinstance(result, str)
        assert result == "not bad"

    run_async_fun(async_render)
```

jayaddison: Thanks!  That suggested `run_async_fn` fixture is used in two locations at the moment (in the `main` branch of the Jinja test suite, that is) - perhaps it'd make sense to promote it to be a shared `conftest.py` fixture?

jayaddison: > Thanks! That suggested `run_async_fn` fixture is used in two locations at the moment (in the `main` branch of the Jinja test suite, that is) - perhaps it'd make sense to promote it to be a shared `conftest.py` fixture?

I've pushed a few changes that make that suggestion: re-using a single definition of a `run_async_fn` fixture from `conftest.py`.

CHANGES.rst
```diff
@@ -9,6 +9,8 @@ Unreleased
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`1793`
 -   Use ``flit_core`` instead of ``setuptools`` as build backend.
+-   Additional test coverage for async rendering of native type templates
```
davidism, line 12: Typically we don't add docs/test changes to the change log.

jayaddison:
Thank you both!

PR #1817, opened 2023-03-02
Clarify what operations the default 'Undefined' supports

Varriount:
This slightly clarifies the documentation regarding what operations the default `Undefined` implementation supports. Before this change, it was only clear from the given example that boolean expressions were supported.


davidism:
Please make documentation fixes against the latest release branch (3.1.x). You can retarget this PR, then rebase onto that branch and force push.

PR #1836, opened 2023-04-07
Refactor: factor-out usage of itertools.groupby from generate_identifier_pattern.py script

### LINKED ISSUE #1835: flake8 warning: B031: repeat usage of generator
I'm not complete certain whether this is a true-positive or a false-positive, but it looks like the [`pre-commit` `flake8` checks in continuous integration](https://results.pre-commit.ci/run/github/994551/1680527637.RBDpIf3CRH6i1TPxShMr6A) are warning about [generator-reuse here](https://github.com/pallets/jinja/blob/85e5ad85d10eaf119dc0f17f245118c9f78bb864/scripts/generate_identifier_pattern.py#L33-L34).

```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

scripts/generate_identifier_pattern.py:33:18: B031 Using the generator returned from `itertools.groupby()` more than once will do nothing on the second usage. Save the result to a list, if the result is needed multiple times.
scripts/generate_identifier_pattern.py:34:15: B031 Using the generator returned from `itertools.groupby()` more than once will do nothing on the second usage. Save the result to a list, if the result is needed multiple times.
scripts/generate_identifier_pattern.py:34:24: B031 Using the generator returned from `itertools.groupby()` more than once will do nothing on the second usage. Save the result to a list, if the result is needed multiple times.
```

Environment:

- Python version: Unknown
- Jinja version: 85e5ad85d10eaf119dc0f17f245118c9f78bb864

jayaddison:
Refactors the `generate_identifier_pattern.py` script; initially this was done in order to experiment with removal of `itertools.groupby` code that `flake8` pre-commit checks were complaining about (see #1835).

- Resolves #1835 

Checklist:

  - Please note: I think that the best way to confirm that these changes are correct is to use the script to regenerate `jinja`'s [Unicode identifier regex pattern](https://github.com/pallets/jinja/blob/85e5ad85d10eaf119dc0f17f245118c9f78bb864/src/jinja2/_identifier.py) 

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Unreleased
 
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`1793`
+-   Refactor the ``generate_identifier_pattern.py`` script. :issue:`1654`
 
 
 Version 3.1.2
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ Unreleased
 
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`1793`
--   Refactor the ``generate_identifier_pattern.py`` script. :issue:`1654`
+-   Refactor the ``generate_identifier_pattern.py`` script. :issue:`1835`
 
 
 Version 3.1.2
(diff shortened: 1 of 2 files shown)


davidism:
The check is getting confused by the name reuse, this does not require a huge refactor.

jayaddison:
Thanks; that's fair reason, yep.

FYI: there is a single-character difference in the output produced by the refactor compared to the script as-is when I run it locally; the existing script emits a trailing minus symbol (range indicator).

That doesn't appear to produce any change-in-behaviour based on my testing in Py3.11 (I was briefly worried that it could allow the `-` character to appear in identifiers, or to exclude the preceding character from successfully matching, but those don't appear to be the case).

PR #1851, opened 2023-05-22
Configurable Parser

### LINKED ISSUE #1850: Configurable Parser
It's not a common use case, but I would find it very useful to be able to tweak how the `Parser` works. For my use case, I'd like to tweak the keywords the `Parser` recognises (`for`, `if`, `block`...) in order to restrict them as necessary (but not extend them, as that would involve changing the actual behaviour of the `Parser`, which I'm not aiming to tackle here).

Currently the `Parser` recognises the keywords from a module-level constant, which is not easily editable just for 1 `Parser` class.
https://github.com/pallets/jinja/blob/953acd65b2be5428482dcb60f5b8481b66252ac9/src/jinja2/parser.py#L18-L33
Besides, even if it were possible to specifically change a `Parser`, say by subclassing it and dynamically overwriting some of the methods, we would need to further subclass an `Environment` and override some of its methods, as the `Parser` is hardcoded in them.
https://github.com/pallets/jinja/blob/953acd65b2be5428482dcb60f5b8481b66252ac9/src/jinja2/environment.py#L615-L619

I would find much easier to implement this behaviour if the `Parser` had a class-level variable that could be swapped in an inheriting class, and the same with a class-level variable for the `Parser` in the `Environment` class. This is a pattern that's already followed for other internals, such as the `CodeGenerator` class via the `code_generator_class` class-variable.
https://github.com/pallets/jinja/blob/953acd65b2be5428482dcb60f5b8481b66252ac9/src/jinja2/environment.py#L284

hasier:
Adds a class-variable to `Parser` so that subclasses can override which keywords the parser will recognise.
Similarly, adds a class-variable to `Environment` so that subclasses can override which `Parser` they'll use.

- fixes #1850

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Unreleased
 
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`1793`
+-   Make ``Parser`` statement keywords configurable and make ``Parser`` configurable
+    for ``Environment``. :issue:`1850`
 
 
 Version 3.1.2
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -138,6 +138,15 @@ useful if you want to dig deeper into Jinja or :ref:`develop extensions
        template variables are handled.  For details, see
        :class:`~jinja2.runtime.Context`.
 
+    .. attribute:: parser_class
+
+       The Parser used internally.  This should not be changed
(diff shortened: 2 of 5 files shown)


hasier:
The test failing in `3.12-dev` seems to also be [failing in `main`](https://github.com/pallets/jinja/actions/runs/4853250663/jobs/8649197618), so it does not seem related to the changes here.

PR #1852, opened 2023-05-26
fix f-string syntax error in code generation

### LINKED ISSUE #1792: f-string sytax error when importing macro in a template which filename is also a template
With this template structure:
```
/templates/
+- {% jinja expr %}.jinja
+- macro_file.jina
```
When `{% jinja expr %}.jinja` imports a macro using 'from "macro_file.jinja" import my_macro' syntax, Environment.render fails with `SyntaxError: f-string: invalid syntax` error message.

Internally jinja concatenates the importing template filename into an f-string, which results with invalid f-string:
```python
if l_0_print_helo is missing:
        l_0_print_helo = undefined(f"the template {included_template.__name__!r} (imported on line 1 in '{% if True %}a{%endif%}.txt.jinja') does not export the requested name 'print_helo'", name='print_helo')
```

Couldn't reproduce the error using `import "macro_file.jinja" as m` syntax.

Environment:

- Python 3.10.7
- Jinja 3.1.2

sisp:
I've fixed a bug that caused an f-string syntax error in the template compilation when a template, that imports a macro, contains curly braces in its name. The bug was caused by the code-generation of an f-string in which the template name is inserted, and when the template name contains curly braces and the generated code gets executed, Python treats the curly braces as f-string braces. I've changed the code-generation to construct the string differently, such that the affected substring is not an f-string. Instead, the different substrings are concatenated using the `+` operator.

- fixes #1792

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@ Unreleased
 -   The sandboxed environment handles indirect calls to ``str.format``, such as
     by passing a stored reference to a filter that calls its argument.
     :ghsa:`q2x7-8rv6-6q7h`
+-   Escape template name before formatting it into error messages, to avoid
+    issues with names that contain f-string syntax.
+    :issue:`1792`, :ghsa:`gmj6-6f8f-6699`
 -   Sandbox does not allow ``clear`` and ``pop`` on known mutable sequence
     types. :issue:`2032`
 -   Calling sync ``render`` for an async template uses ``asyncio.run``.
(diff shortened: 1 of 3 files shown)


sisp:
I've rebased this PR onto the latest state of the `3.1.x` branch. All tests are passing except `pre-commit.ci - pr`, but the reported errors originate from `3.1.x` and weren't introduced by this PR.

@davidism (or any other maintainer) &ndash; may ask you for a review? :pray:

sisp:
I've performed another rebase onto the latest state of the `3.1.x` branch. All tests are passing now.

davidism:
I've refactored the fix so that the curly braces are escaped rather than trying to keep track of nested quote levels and string concatenation. I also fixed the test, as it didn't seem to trigger an error regardless of the fix.

PR #1873, opened 2023-07-25
Add Python version classifiers

### LINKED ISSUE #1872: Add explicit Python version classifiers in pyproject.toml
It would be beneficial to explicitly add the Python version classifiers in the `pyproject.toml` file, such that this package shows correctly its supported Python versions on the ['PyReadiness' website](https://pyreadiness.org/3.8/). I will create a PR for this.

JeremyKeusters:
This change explicitly adds Python version classifiers, such that this package shows correctly its supported Python versions on the ['PyReadiness' website](https://pyreadiness.org/3.8/)

- fixes #1872 

Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,12 @@ classifiers = [
     "License :: OSI Approved :: BSD License",
     "Operating System :: OS Independent",
     "Programming Language :: Python",
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.7',
+    'Programming Language :: Python :: 3.8',
+    'Programming Language :: Python :: 3.9',
+    'Programming Language :: Python :: 3.10',
+    'Programming Language :: Python :: 3.11',
     "Topic :: Internet :: WWW/HTTP :: Dynamic Content",
     "Topic :: Text Processing :: Markup :: HTML",
 ]


JeremyKeusters:
Based on the information I have, I added support for versions `3.7` up until `3.11`, but I see there are no tests for `3.11`. Can a maintainer confirm these are the right versions?

ThiefMaster:
https://github.com/pallets/jinja/blob/main/tox.ini#L3

so I'd say you can do 3.7 to 3.12

JeremyKeusters:
Thx for the reply @ThiefMaster , I added 3.12.

davidism:
See my comments on the pyreadiness issue tracker: https://github.com/di/pyreadiness/issues/18

JeremyKeusters:
@davidism I understand your reasoning, but I don't agree. I think that `python_requires >= 3.7` is way too broad. When Python 3.13 is in prerelease in a few months, this tag (in absence of classifiers) suggests that your package will work with that version from day 1, even though it was maybe not explicitly tested.

By including explicit classifiers, we can significantly enhance the user experience. Potential users will be able to quickly determine whether the package is deemed compatible through testing with their Python version. Moreover, classifiers provide a more comprehensive view of the supported versions, including both minimum and maximum versions, offering clarity for the community and external dependency managers.

I acknowledge that maintaining classifiers can be an additional responsibility, but I believe the benefits mentioned above outweigh the effort required. Especially given that these classifiers will likely only need to be updated once a year, maybe twice if you want to drop a Python version.

Ultimately, it is up to you as one of the project maintainers to make the final decision, and I respect that. However, I sincerely hope you will reconsider the change and recognise the value it can bring to the jinja2 community.

davidism:
Thanks, but I don't agree with your disagreement, for the reasons I clearly stated in the pyreadiness tracker.

PR #1880, opened 2023-08-23
improve annotations for methods returning copies

viccie30:
These methods always return a new object of the same type as the object they are bound to, even if this is a subclass. Change the annotations to reflect that.

I have not opened an issue, because it's not a functional bug. It just causes issues with static type checkers.

Checklist:

- [ ] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change. (N/A)
- [ ] Add or update relevant docs, in the docs folder and in code.  (N/A)
- [ ] Add `.. versionchanged::` entries in any relevant code docs. (N/A)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,6 +39,7 @@ Unreleased
     searched. :issue:`1661`
 -   ``PackageLoader`` shows a clearer error message when the package does not
     contain the templates directory. :issue:`1705`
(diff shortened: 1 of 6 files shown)


viccie30:
The pre-commit failure is not related to the changes in the PR and also happens with the current state of the branch.

PR #1903, opened 2023-10-25
ISSUE-1902 Fix Template::is_up_to_date contract

### LINKED ISSUE #1902: Property Template::is_up_to_date breaks type hint contract
environment.py::Template::is_up_to_date (https://github.com/pallets/jinja/blob/main/src/jinja2/environment.py#L1496) tries to call self._uptodate even if this correctly is a boolean and not callable, disallowing to correctly implement BaseLoader::get_source with an alway up_to_date get_source like here: https://github.com/peering-manager/peering-manager/blob/main/peering_manager/jinja2/loaders.py#L48

Overwrite a BaseLoaders get_source method returning a bool as last value as type hints allow and include template twice with cache enabled like here: https://github.com/peering-manager/peering-manager/blob/main/peering_manager/jinja2/loaders.py#L6

Untested example code snippet:
```
from jinja2 import BaseLoader


class MyLoader(BaseLoader):
    def get_source(self, environment, template):
        source = ...
        return source, template, True  # true should be allowed but jinja does not handle it here correctly according to type hints
...
---
...
{% include myTemp %}
{% include myTemp %}
```

Example trace

````
  File "/opt/peering-manager/peering_manager/jinja2/__init__.py", line 30, in render_jinja2
    return jinja2_template.render(**context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/peering-manager/.venv/lib/python3.11/site-packages/jinja2/environment.py", line 1301, in render
    self.environment.handle_exception()
  File "/opt/peering-manager/.venv/lib/python3.11/site-packages/jinja2/environment.py", line 936, in handle_exception
  File "<template>", line 5, in top-level template code
  File "/opt/peering-manager/.venv/lib/python3.11/site-packages/jinja2/environment.py", line 1494, in is_up_to_date
    return self._uptodate()
           ^^^^^^^^^^^^^^^^
TypeError: 'bool' object is not callable
```

<!--
Describe the expected behavior that should have happened but didn't.
-->

Respect bools value and do not call it.

Environment:

- Python version: Python 3.11.5
- Jinja version: Jinja2==3.1.2

MR incoming...

rwielk:
https://github.com/pallets/jinja/issues/1902

Template::_uptodate type hints get supporting implementation

- fixes #1902

Checklist:

- [X] Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.
- [X] Add `.. versionchanged::` entries in any relevant code docs. (Opted against any change since this just fixes already expected behaviour from my POV)
- [X] Run `pre-commit` hooks and fix any issues.
- [X] Run `pytest` and `tox`, no tests failed.

Please hint me on where and how you want a test to get implemented for this, did not see an obvious place. (Example test in regressions added, perhaps it's good enough...)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Version 3.2.0
 
 Unreleased
 
+-   Fix Template's type hint contract :pr: `1903`
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`1793`
 -   Use ``flit_core`` instead of ``setuptools`` as build backend.
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -1493,7 +1493,7 @@ class Template:
         """If this variable is `False` there is a newer version available."""
         if self._uptodate is None:
             return True
-        return self._uptodate()
+        return self._uptodate() if callable(self._uptodate) else self._uptodate
 
     @property
     def debug_info(self) -> t.List[t.Tuple[int, int]]:
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -11,6 +11,28 @@ from jinja2.utils import pass_context
 
 
 class TestCorner:
+    def test_uptodate_loader(self, env):
+        class UpToDateDictLoader(DictLoader):
+            def get_source(self, environment: "Environment", template: str):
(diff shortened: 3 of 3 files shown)


davidism:
See discussion in #1902 for close reason.

PR #1911, opened 2023-11-16
Fix typo on filter name

vbuxbaum:
Hi everyone!

I was studying Jinja and Flask with @JefersonViana , and noticed a typo in the docs about custom filters.

I hope there is no need for creating an issue for this.

Let me know if you think that there is a better solution.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -666,8 +666,8 @@ Now it can be used in templates:
 
 .. sourcecode:: jinja
 
-    {{ article.pub_date|datetimeformat }}
-    {{ article.pub_date|datetimeformat("%B %Y") }}
+    {{ article.pub_date|datetime_format }}
+    {{ article.pub_date|datetime_format("%B %Y") }}
 
 Some decorators are available to tell Jinja to pass extra information to
 the filter. The object is passed as the first argument, making the value


davidism:
The typo would be in the use, not the registration. Always prefer separated words, so `|datetime_filter`. (Yes, "datetime" itself doesn't follow this, but it's the built-in name.)

davidism:
Be sure to make docs fixes against the latest feature branch (3.1.x) rather than main so that it gets into the current docs. You'll need to rebase and force push to fix that:

```
$ git rebase --onto origin/3.0.x origin/main
$ git push -f
```

vbuxbaum:
> Be sure to make docs fixes against the latest feature branch (3.1.x) 

Thanks @davidism ! I did not pay attention to that, sorry. Done ✅ 

> The typo would be in the use, not the registration. Always prefer separated words, so `|datetime_filter`. (Yes, "datetime" itself doesn't follow this, but it's the built-in name.)

I also prefer naming separating words with `_`, but I think it is better if the filter function and the register in `environment.filters` have different names. Someone (mainly beginners) may wonder 

- _"is it necessary to keep the same name?"_ 
- or even _"which should be used inside the template? the function's name or the string registered in `environment.filters`"_

What if we rename the `datetime_format` function to `datetime_format_filter`, and keep the filter's name as `datetime_format`?

davidism:
`datetime_format` everywhere please.

vbuxbaum:
> `datetime_format` everywhere please.

@davidism done ✅

vbuxbaum:
The `pre-commit.ci` failure has nothing to do with my changes, right? Do I need to do something about it?

PR #1915, opened 2023-12-13
Make ease of use update to template documentation

### LINKED ISSUE #1825: Docs update re triple quotes and block assignments
I think the docs could use a small update to this section https://jinja.palletsprojects.com/en/3.1.x/templates/#block-assignments to mention "triple quotes", to say that [they are not supported](https://github.com/pallets/jinja/issues/845#issuecomment-508951409), and that set / endset is the right method for multline strings and strings that mix single and double quotes that you don't want to escape.

I've seen people wonder how to do that multiple times, and they assume triple quotes work, and don't find it in the docs with a crtl-f. Literally adding the phrase 

"triple quotes (`'''` and `"""`)"

to that section seems like it would help people discover the right way to do this a lot.

ratchek:
Add the phrases 'multiline comment' and 'triple quotes' to docs in the templates/#block-assignments section. This allows for new users to find this alternative easily.  

Fixes #1825

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -1084,9 +1084,10 @@ Block Assignments
 
 Starting with Jinja 2.8, it's possible to also use block assignments to
 capture the contents of a block into a variable name.  This can be useful
-in some situations as an alternative for macros.  In that case, instead of
-using an equals sign and a value, you just write the variable name and then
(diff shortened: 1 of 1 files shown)


davidism:
continued in #2064

PR #1924, opened 2024-01-03
Eliminating undefined parameters entirely by `NeverUndefined`

### LINKED ISSUE #1923: Introducing `NeverUndefined`
This feature is done by introducing a new type of undefined class called `NeverUndefined`.

The problem is that when using the previously most restricted undefined policy `StrictUndefined` is not enough.

Take example on the following template:

```jinja2
{% macro test(a, b, c) %}
a macro with three parameters but not used
{%endmacro%}

{{test()}}
```

`NeverUndefined` will capture this error because in the normal case one shall pass three parameters to this macro in order to call it. `StrictUndefined` will not.

And this problem would be harder to debug, if one pass these undefined variables to another macro or function calls.

The implementation:

```python
class NeverUndefined(jinja2.StrictUndefined):
    def __init__(self, *args, **kwargs):
        # ARGS: ("parameter 'myvar2' was not provided",)
        # KWARGS: {'name': 'myvar2'}
        if len(args) == 1:
            info = args[0]
        elif "name" in kwargs.keys():
            info = f"Undefined variable '{kwargs['name']}"
        else:
            infoList = ["Not allowing any undefined variable."]
            infoList.append(f"ARGS: {args}")
            infoList.append(f"KWARGS: {kwargs}")
            info = "\n".join(infoList)

        raise Exception(info)
```

James4Ever0:
This pull request is done by changing two files in the source code directory. It integrates the `NeverUndefined` class and make it usable by:

```python
import jinja2
# access by: jinja2.NeverUndefined
```
This will raise exception for the following template, while `StrictUndefined` will not:

```jinja2
{% macro test(a, b, c) %}
a macro with three parameters but not used
{%endmacro%}

{{test()}}
```

- fixes #1923

Checklist:

### DIFF
--- a/src/jinja2/__init__.py
+++ b/src/jinja2/__init__.py
@@ -25,8 +25,8 @@ from .loaders import PrefixLoader as PrefixLoader
 from .runtime import ChainableUndefined as ChainableUndefined
 from .runtime import DebugUndefined as DebugUndefined
 from .runtime import make_logging_undefined as make_logging_undefined
-from .runtime import StrictUndefined as StrictUndefined
 from .runtime import NeverUndefined as NeverUndefined
+from .runtime import StrictUndefined as StrictUndefined
 from .runtime import Undefined as Undefined
 from .utils import clear_caches as clear_caches
 from .utils import is_undefined as is_undefined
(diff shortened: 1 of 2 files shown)


James4Ever0:
> did you create the PR using an LLM? the code and comments in there surely look like that's the case...
> 
> You also don't explain when your code would actually be useful...

This code is half a year old and I do this for fixing this long standing error.

I want to explain but my memory is fuzzy. It is useful.

davidism:
See issue for discussion. If you think this is generally useful to others, you can create an extension, but there's nothing requiring it to be in Jinja core.

PR #1935, opened 2024-02-04
tests: Tweak to work with markupsafe 2.1.4

rpurdie:
The recent markupsafe release 2.1.4 causes a regression in the jinja2 tests as a newline is present. Use strip to work around this change in behaviour to stop the test suite being affected.


rpurdie:
https://github.com/pallets/markupsafe/issues/417 means this is probably incorrect.

PR #1938, opened 2024-02-16
Slightly improve clarity of logical bool ops

sirosen:
The doc states that these "return true" but more precisely they're the "typical" short-circuiting boolean operations with Python-like semantics. ``and`` and ``or`` may not return booleans at all depending on their inputs.

---

I had a coworker express the expectation that ``x or y`` in a Jinja template would return a bool, in a context in which ``x`` and ``y`` are both ``str | None``, based on this section of the docs.
I don't read the doc in this way, but strictly speaking that's what it says (these don't say that they're Python-like boolean operations -- they say they "return true").

I think the clarification is possibly helpful and at worst harmless. It was written based off of the Python documentation for the boolean operators.[^1]

[^1]: https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -1412,27 +1412,31 @@ Comparisons
 Logic
 ~~~~~
 
-For ``if`` statements, ``for`` filtering, and ``if`` expressions, it can be useful to
-combine multiple expressions:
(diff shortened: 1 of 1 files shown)


davidism:
I added further clarification.

PR #1946, opened 2024-03-18
Fix: don't use current running loop

Jahongir-Qurbonov:
- fail run Template.generate when use in running event loop

Checklist:

### DIFF
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -1340,10 +1340,23 @@ class Template:
         if self.environment.is_async:
             import asyncio
 
+            close = False
+
+            try:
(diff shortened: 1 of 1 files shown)


davidism:
You call `generate_async` if you're already in a loop. The code you've shown is for when you're in sync code but the environment is async.

Jahongir-Qurbonov:
I can't call `template.stream(...)` in running loop, because it calls `generate` and raise error.

davidism:
Should be fine using `generate_async`. `stream` is just applying optional buffering (it's still unbuffered by default), which can be done pretty trivially with your own wrapper if you actually know it's needed.

PR #1982, opened 2024-05-19
include "examples" directory in the sdist again

dne:
When moving to flit from setuptools, the "examples" directory was lost from the source distribution. I'm assuming that this was unintentional, since `pyproject.toml` on the main branch still includes it.

Fixes regression introduced in #1968.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ name = "jinja2"
 [tool.flit.sdist]
 include = [
     "docs/",
+    "examples/",
     "requirements/",
     "tests/",
     "CHANGES.md",


davidism:
Thanks, but I don't think we need to include them. The examples aren't part of the code docs or tests (and those last two don't _technically_ need to be there either).

dne:
Sure, they're not strictly necessary, but they've been part of the sdist since version 2.0 up until 3.1.3. And, as I mentioned, this regression seems specific for the `3.1.x` branch.

FWIW, I discovered this when packaging 3.1.4 for Alpine Linux, which installs the examples as part of the "doc" sub-package (and [so does e.g. Debian](https://packages.debian.org/bookworm/all/python-jinja2-doc/filelist) apparently)

PR #1985, opened 2024-05-21
Refactor PackageLoader to use importlib_resources

### LINKED ISSUE #1978: refactor `PackageLoader` to use `importlib.resources`
I had to ignore a failing test when adding 3.13 to the test matrix #1977, as the zip importer no longer exposes a private `_files` list for `PackageLoader.list_templates` to use. Presumably I didn't use `importlib.resources` before because I had to support older Python versions? Either way, should be able to now. Consider the `importlib_resources` backport for 3.7 and 3.8 so we can use the `files` API in all versions.

danwuSBU:
- pip installed `importlib-resources` 
- added `importlib-resources` to `docs.in tests.in typing.in` and `pip-compile`d them
- refactored `PackageLoader` to use `importlib_resources`

fixes #1978


danwuSBU:
have to see why this is failing on Windows

PR #1986, opened 2024-05-22
Add python versions to pyproject.toml classifiers

labianchin:
Motivation: I realized these classifiers where missing, when looking at https://pyreadiness.org/3.12/


davidism:
See my comments on the pyreadiness issue tracker; https://werkzeug.palletsprojects.com/en/2.1.x/routing/

PR #1996, opened 2024-06-06
Fix error message for non-existing directory

### LINKED ISSUE #1995: PackageLoader raises misleading error when template directory does not exist.
When using `PackageLoader` and the template directory does not exist, it throws an error saying:

```
The 'foobar' package was not installed in a way that PackageLoader understands.
```

This is clearly incorrect. If the package is actually not there, we get an assertion error. I would say the error should be something like "Template directory was not found in package 'foobar'." or something along those lines.

Describe how to replicate the bug.
Just create a package without the template directory. Try to pass that name of the package to PackageLoader causes the incorrect error to be thrown.

I hope this is not considered too trivial. It did cost me some time to figure out, and only after reading the source code, so I think it's worth changing to something that makes the root cause more apparent.

Environment:

- Python version: 3.10.12
- Jinja version: 3.1.4

elektito:
As described in issue #1995, the error message thrown when the template directory does not exist was very misleading. This is a simple attempt to fix that.

fixes #1995

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ Unreleased
     ``Template.generate_async``. :pr:`1960`
 -   Avoid leaving async generators unclosed in blocks, includes and extends.
     :pr:`1960`
+-   Fix misleading error message when template does not exist.
+    :pr:`1995`
 
 
 Version 3.1.4
--- a/src/jinja2/loaders.py
+++ b/src/jinja2/loaders.py
@@ -347,10 +347,7 @@ class PackageLoader(BaseLoader):
                     break
 
         if template_root is None:
-            raise ValueError(
-                f"The {package_name!r} package was not installed in a"
-                " way that PackageLoader understands."
-            )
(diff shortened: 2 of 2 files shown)


davidism:
This seems even less clear. It doesn't say what package, or what directory name. I'm pretty sure this specific message can happen in other cases too, not just when the directory doesn't exist, so you'll need to split out detection of that and raise a specific error.

elektito:
If I understand correctly, in cases where the package/module does not exist, we get an assertion error. So this is left exclusively for cases where the template directory does not exist.

elektito:
I could add the package name and template directory to the error message if you think that's better. My reasoning for not including them was that in most cases these are included as strings, so you could see it right in the stack trace.

elektito:
I think if there is a case that we can find a package, but cannot find any path associated for it, this can happen too. Not quite sure if that's possible or not. But even in that case, the original error message doesn't make a lot of sense.

davidism:
duplicate of #1706

Issue #1316, opened 2020-11-27
Template "filename" using Template constructor / from_string


stevelittlefish:
When you create a template by passing a string into the Template constructor, if there is an exception raised by that template it appears in the stack trace like this:

`File "<template>", line 2, in top-level template code`

I would like to be able to pass a string into the Template constructor to identify the template in the stack trace like this:

`my_template = Template('<html>{{ invalid_variable }}</html>', identifier='<Product: ball-bearings>')`

Which would cause the exception to be rendered like this:

`File "<Product: ball-bearings>", line 2, in top-level template code`

This would make it a lot easier to find template issues when you have a lot of small templates rendered in a loop.

davidism:
```python
env = Environment()
code = env.compile("source", filename="<product>")
t = env.template_class.from_code(env, code, env.globals)
```

You can also do this by making your own loader, `Loader.load` returns a tuple of `code, filename, up_to_date_callback`. If you're creatinga  bunch of templates the way you're describing, that's a good idea regardless of wanting custom filenames, as it will be able to take advantage of the compilation cache.

davidism:
While it would be *possible* to add an argument to `from_string`, I think that might give users the impression that it's appropriate to use `from_string` as you've described, for lots of small templates. If a project is rendering so many small templates with errors that it becomes hard to tell them apart, that's an indication they should write a loader. It will give full control and greatly speed up rendering.

stevelittlefish:
Thanks @davidism - your suggestion looks like exactly what I was after.  I will update my code to use a loader.

Issue #1304, opened 2020-10-30
Ability to write Jinja loaders that use async/await


simonw:
I'm writing code an an async environment, using Jinja's excellent async mode.

I want to write a custom template loader that can load templates asynchronously from my database - but as far as I can tell the Jinja loader mechanisms only work with synchronous functions.

Borrowing an example [from here](https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.FunctionLoader) I'd love to be able to do something like this:
```python
async def load_template(name):
    row = await db.execute("select body from templates where name = ?", name)
    return row[0]

loader = FunctionLoader(load_template)
```
I initially implemented a workaround for this by loading templates asynchronously in my own code and compiling them for Jinja with `jinja_env.from_string(plugin_template_source)` - but then I realized that this doesn't work for templates loaded internally by Jinja due to `{% extends ... %}` and `{% include ... %}` blocks.

I'd be happy to help implement this if it's a feature that Jinja maintainers agree should be part of the library.

davidism:
I'd be fine adding it in, but I have no idea what the API/implementation would look like. Did you have something in mind?

simonw:
My ideal implementation as a user would be the one listed above - where the `FunctionLoader` loader can take an async function and magically call it in the right way, by internally using the `auto_await()` function (assuming Jinja is running in async mode, otherwise it would throw an error): https://github.com/pallets/jinja/blob/1eca37ba6d55c57f837a2126a5e240fd596f2e6d/src/jinja2/asyncsupport.py#L169-L172

The implementation looks tricky though, since I imagine getting `{% include %}` and `{% extends %}` working against an async loader would require some changes to the code that the Jinja compiler generates, which I'm not at all familiar with.

It may be more Jinja-like to have a new `Loader` class called `AsyncLoader` rather than reusing the existing `FunctionLoader` though.

grondman:
I'm using Tortoise ORM all over the place and being able to asynchronously load templates from a DB is something I'd definitely use 🙂

ssjunior:
I am using a custom loader from database in a Django async applications... Is there a way to make the loader async? Otherwise I can't load templates from database.. :(

bartfeenstra:
A workaround is to run your internal asynchronous code using `asyncio.to_thread()`.

hussam-almarzoq:
For reference, this package https://pypi.org/project/jinja2-async/ developed by @FatimahAlhumaidhi provides an async variant of the Jinja APIs

PR #2013, opened 2024-08-06
Allow `Environment.from_string()` to take a filename for the loaded template

bartfeenstra:
I currently use `Environment.from_string()` to dynamically load template files, and I would like to be able to set the correct file name on them so the generated tracebacks are much more helpful in identifying the source file (rather than the tracebacks saying the file was `<template>`). The reason I am not using a `Loader` is twofold: none of these files ever need to be rendered twice, and `Loader` is blocking (i.e. not async).

Please look at the PR as an example of the approach itself. If you think this is something that would be approved, I'll finish up the tests and documentation.

## To do
- Add tests that demonstrate the correct behavior of the change. Tests
- Add or update relevant docs, in the docs folder and in code.
- Add an entry in CHANGES.rst summarizing the change and linking to the issue.
- Add `.. versionchanged::` entries in any relevant code docs.

### DIFF
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -1093,6 +1093,7 @@ class Environment:
         source: t.Union[str, nodes.Template],
         globals: t.Optional[t.MutableMapping[str, t.Any]] = None,
         template_class: t.Optional[t.Type["Template"]] = None,
+        filename: t.Optional[str] = None,
     ) -> "Template":
         """Load a template from a source string without using
         :attr:`loader`.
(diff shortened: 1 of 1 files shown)


davidism:
As explained in https://github.com/pallets/jinja/issues/1316, we don't plan to add this.

bartfeenstra:
Thanks for clarifying this. I do think this is a needless restriction, especially considering https://github.com/pallets/jinja/issues/1304 has not been worked on (I get it, we're all busy, this is not a criticism). Switching to loaders, while possible, would add some unneeded blocking file I/O to my application, just to get a better hint in the traceback.

davidism:
There's another option offered in the issue I linked, which does allow setting the filename. I don't know your exact circumstance, but I wouldn't expect loading the template to be a significant length of blocked time. You could also use a thread executor if it was an issue: https://docs.python.org/3/library/asyncio-task.html#running-in-threads

> has not been worked on (I get it, we're all busy, this is not a criticism)

Thanks for understanding. I'm always happy to review a good design or PR! I don't personally use asyncio, so waiting for me to do something about it is going to take a lot longer, compared to someone who needs and understands asyncio contributing to the project.

bartfeenstra:
I admit I completely missed that one suggestion earlier, but I'm happy to report that I successfully implemented it in https://github.com/bartfeenstra/betty/pull/1861.

Thanks for your time and work on Jinja! 😊

PR #2022, opened 2024-09-03
Make compiled output deterministic for tuple unpacking in set tag

### LINKED ISSUE #2021: non-deterministic output from compile templates when using tuple unpacking
Ideally compile_templates will have deterministic output so we can use git diff etc

See also https://github.com/pallets/jinja/issues/1453 and https://github.com/pallets/jinja/issues/1452

Found another case where this occurs, e.g. I'm randomly getting these changes or their inverse:

```diff
-    context.vars.update({'triage_reports': l_0_triage_reports, 'open_group_index': l_0_open_group_index})
-    context.exported_vars.update(('triage_reports', 'open_group_index'))
+    context.vars.update({'open_group_index': l_0_open_group_index, 'triage_reports': l_0_triage_reports})
+    context.exported_vars.update(('open_group_index', 'triage_reports'))
```
or:
```diff
-                _loop_vars.update({'row_spans': l_2_row_spans, 'row': l_2_row})
+                _loop_vars.update({'row': l_2_row, 'row_spans': l_2_row_spans})
```

I can see that both of the cases above are a result of using tuple unpacking to set multiple new context vars in a single `set` call
e.g.
```jinja
{% set triage_reports, open_group_index = get_nav_list_items(TRIAGE_REPORTS_NAV, url_name, url_func=url) %}
```
```jinja
{% set row, row_spans = row %}
```

The output from compiling the templates should be stable if the underlying template hasn't changed, rather than varying randomly each time you run it.

Environment:

- Python version: 3.11.5
- Jinja version: 3.1.4

anentropic:
Makes compiled output deterministic, when doing tuple unpacking to set multiple vars in a single `{% set ... %}` call.

fixes #2021

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,8 @@ Unreleased
     async-aware filter. :issue:`1781`
 -   ``|int`` filter handles ``OverflowError`` from scientific notation.
     :issue:`1921`
+-   Make compiling deterministic for tuple unpacking in a ``{% set ... %}``
+    call. :issue:`2021`
 
 
 Version 3.1.4
--- a/src/jinja2/compiler.py
+++ b/src/jinja2/compiler.py
@@ -811,7 +811,7 @@ class CodeGenerator(NodeVisitor):
                 self.writeline("_block_vars.update({")
             else:
                 self.writeline("context.vars.update({")
-            for idx, name in enumerate(vars):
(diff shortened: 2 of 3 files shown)


davidism:
Please rename this with a descriptive title. The issue number is already linked in the body.

PR #2026, opened 2024-10-01
Preserve `__slots__` metadata on Undefined types

### LINKED ISSUE #2025: Undefined objects can't be copied or pickled by Python > 3.5
### Description
Due to [an ancient erroneous deletion of `__slots__` metadata](https://github.com/pallets/jinja/blob/3.1.4/src/jinja2/runtime.py#L1049-L1056), it's not possible to copy/deepcopy/pickle `Undefined` objects (all of which we occasionally need to do in Ansible) in Python > 3.5.

The comment above the deletion of `__slots__` implies a misunderstanding of how that metadata was/is used. The deletion was always probably technically a bug, but once Python started providing default implementations of `__getstate__` and `__reduce_ex__` on all objects (which consult that metadata at runtime), it broke the default copyability/pickle-ability of `Undefined` (and derived) objects.

We're working around this with a monkeypatch that restores the missing `__slots__` metadata. We've also submitted PR  #2026 to zap the problematic deletion and tests that verify copy/deepcopy/pickle functionality on all the built-in `Undefined` types.

### Repro

(on any Python > 3.5 with Jinja installed):
```
$ python -c 'import copy; import jinja2; copy.copy(jinja2.Undefined())'
  File "<string>", line 1, in <module>
  File "/usr/lib64/python3.11/copy.py", line 92, in copy
    rv = reductor(4)
         ^^^^^^^^^^^
TypeError: cannot pickle 'Undefined' object
```

### Expected
The default runtime-provided implementations of `__getstate__` and `__reduce_ex__` that back `copy.copy`, `copy.deepcopy` and `pickle` should work properly with all `Undefined` derived types.

### Environment

- Python version: >= 3.6
- Jinja version: anything from at least the past 16 years

nitzmahone:
Restores erroneous deletions of `__slots__` metadata from `Undefined` types that breaks various serialization/copy mechanisms on Python >= 3.6. Also:

* fix ChainedUndefined to raise AttributeError on unimplemented dunder methods to prevent breaking core Python protocols (including copy)
* add tests

fixes #2025

### DIFF
--- a/src/jinja2/runtime.py
+++ b/src/jinja2/runtime.py
@@ -860,7 +860,11 @@ class Undefined:
 
     @internalcode
     def __getattr__(self, name: str) -> t.Any:
-        if name[:2] == "__":
+        # Raise AttributeError on requests for names that appear to be unimplemented
+        # dunder methods to keep Python's internal protocol probing behaviors working
+        # properly in cases where another exception type could cause unexpected or
+        # difficult-to-diagnose failures.
+        if name[:2] == "__" and name[-2:] == "__":
             raise AttributeError(name)
 
         return self._fail_with_undefined_error()
@@ -984,10 +988,20 @@ class ChainableUndefined(Undefined):
     def __html__(self) -> str:
         return str(self)
 
-    def __getattr__(self, _: str) -> "ChainableUndefined":
+    def __getattr__(self, name: str) -> "ChainableUndefined":
+        # Raise AttributeError on requests for names that appear to be unimplemented
+        # dunder methods to avoid confusing Python with truthy non-method objects that
+        # do not implement the protocol being probed for. e.g., copy.copy(Undefined())
+        # fails spectacularly if getattr(Undefined(), '__setstate__') returns an
+        # Undefined object instead of raising AttributeError to signal that it does not
+        # support that style of object initialization.
+        if name[:2] == "__" and name[-2:] == "__":
+            raise AttributeError(name)
+
         return self
 
-    __getitem__ = __getattr__  # type: ignore
+    def __getitem__(self, _name: str) -> "ChainableUndefined":  # type: ignore[override]
+        return self
 
 
 class DebugUndefined(Undefined):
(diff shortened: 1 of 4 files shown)


src/jinja2/runtime.py
```diff
@@ -982,10 +982,15 @@ class ChainableUndefined(Undefined):
     def __html__(self) -> str:
         return str(self)
 
-    def __getattr__(self, _: str) -> "ChainableUndefined":
+    def __getattr__(self, name: str) -> "ChainableUndefined":
```
davidism, line 985: I don't understand this change. Can you make the comment clearer?

src/jinja2/runtime.py
```diff
@@ -982,10 +982,15 @@ class ChainableUndefined(Undefined):
...
+        # unimplemented dunder methods returning Undefined break core Python protocols
+        if name[:2] == "__":
+            raise AttributeError(name)
+
         return self
 
-    __getitem__ = __getattr__  # type: ignore
+    def __getitem__(self, _: str) -> "ChainableUndefined":  # type: ignore
```
davidism, line 992: Don't use `_` as an argument name. Use specific ignores, not a bare ignore.

src/jinja2/runtime.py
nitzmahone: This is the same (incomplete, BTW) heuristic exclusion [already used by `Undefined.__getattr__`](https://github.com/pallets/jinja/blob/3.1.4/src/jinja2/runtime.py#L861). `ChainableUndefined` needs it for the same reason, but there weren't previously any tests that exhibited the bugs caused by the current unconditional behavior.

There are a number of Python interaction protocols that sniff for the presence of a particular dunder method to support builtin behavior. This is one of the places where Jinja's getattr/getitem equivalency causes problems. In this case, `copy` uses the presence of a `__setstate__` method on the type to decide how it will create and populate the copy. Since `hasattr(ChainableUndefined, '__setstate__')` is true with the current impl, Python assumes it can call that to populate the empty storage for the copy, but blows up when it actually tries to invoke the `Undefined` object it receives instead of a bound method to fill in an empty object instance.

Happy to include a more detailed inline explanation along those lines in the code. but should probably either copy/paste to the original usage or actually share that logic between them.

Also happy to correct the heuristic in both places to only exclude `__XYZ__` instead of the current `__.*` - while probably unlikely, the current impl would incorrectly exclude, eg `__fooattr`.

src/jinja2/runtime.py
nitzmahone: lol, that's what PyCharm coughed up for the override signature- I assumed the base class did the same, but apparently it was just confused by the dynamic `_fail_with_undefined_error` aliasing. 

The base class is already doing the same blanket `type: ignore` on that method due to the aforementioned dynamic aliasing- I detest overly-broad ignores as well, so happy to figure out a more targeted ignore (and of course use a proper arg name).

davidism: Remove the ignore, run mypy, add an ignore for the actual finding. Yeah, there's still a bunch of bare ignores from when we first added typing, but I don't want to persist that going forward.

PR #2029, opened 2024-10-01
Fix pickle/copy support for the `missing` singleton

### LINKED ISSUE #2027: Cannot pickle `jinja2.utils.missing`
The `jinja2.utils.missing` singleton instance cannot be pickled since pickle cannot find `MissingType` due to the `jinja2.utils.MissingType` attribute not existing:

```python
import pickle
import jinja2.utils

pickle.dumps(jinja2.utils.missing)
```

Running the above code results in the following error:

```
  File "/tmp/bug.py", line 4, in <module>
    pickle.dumps(jinja2.utils.missing)
_pickle.PicklingError: Can't pickle <class 'jinja2.utils.MissingType'>: attribute lookup MissingType on jinja2.utils failed
```

Attempting to pickle (and then restore) the `jinja2.utils.missing` singleton instance should succeed. Supporting this would allow pickling of `jinja2.runtime.Undefined` instances which do not set `obj`, once https://github.com/pallets/jinja/issues/2025 is also resolved.

Environment:

- Python version: 3.11.1
- Jinja version: 3.1.4

mattclay:
Fix pickle/copy support for the `missing` singleton:

* Fix pickle support for the `missing` singleton.
* Add unit test for pickling the `missing` singleton.
* Add unit test for copying the `missing` singleton.

Fixes #2027

### DIFF
--- a/src/jinja2/utils.py
+++ b/src/jinja2/utils.py
@@ -18,8 +18,17 @@ if t.TYPE_CHECKING:
 
 F = t.TypeVar("F", bound=t.Callable[..., t.Any])
 
-# special singleton representing missing values for the runtime
-missing: t.Any = type("MissingType", (), {"__repr__": lambda x: "missing"})()
+
+class _MissingType:
+    def __repr__(self) -> str:
+        return "missing"
+
+    def __reduce__(self) -> str:
+        return "missing"
+
+
+missing: t.Any = _MissingType()
+"""Special singleton representing missing values for the runtime."""
 
 internal_code: t.MutableSet[CodeType] = set()
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,8 @@ Unreleased
     call. :issue:`2021`
 -   Fix dunder protocol (`copy`/`pickle`/etc) interaction with ``Undefined``
     objects. :issue:`2025`
+-   Fix `copy`/`pickle` support for the internal ``missing`` object.
+    :issue:`2027`
 
 
 Version 3.1.4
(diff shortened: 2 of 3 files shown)


mattclay:
Returning `"missing"` from `__reduce__` is a [documented solution](https://docs.python.org/3/library/pickle.html#object.__reduce__) for singleton types:

> If a string is returned, the string should be interpreted as the name of a global variable. It should be the object’s local name relative to its module; the pickle module searches the module namespace to determine the object’s module. **This behaviour is typically useful for singletons.**

src/jinja2/utils.py
```diff
@@ -19,7 +19,11 @@
 F = t.TypeVar("F", bound=t.Callable[..., t.Any])
 
 # special singleton representing missing values for the runtime
-missing: t.Any = type("MissingType", (), {"__repr__": lambda x: "missing"})()
+missing: t.Any = type(
+    "MissingType",
+    (),
+    {"__repr__": lambda x: "missing", "__reduce__": lambda x: "missing"},
+)()
```
mattclay, line 26: I used this form to minimize changes, but it could be replaced an actual `class` definition if desired.

davidism: Yeah, that makes more sense. I don't see a reason a "singleton" was done this way, it's enough to call the class `_MissingType` to indicate it's not public.

nitzmahone: Always the challenge when interacting with unfamiliar projects/maintainers around finding the balance of "minimize changes" vs "if you were building this today, it would look very different" :laughing:

PR #2031, opened 2024-10-02
Added linebreak option for indent filter.

### LINKED ISSUE #2016: indent filter uses \n always rather than following api newline_sequence
Environment:

- Python version: 3.8.10
- Jinja version: 3.1.4

mjeromin:
This change addresses the line break limitation with indent filter, which only supports line feed (LF). The change adds support for multiline template values that use carriage return line feed (CRLF) -- a standard for Windows text files. To enable this feature, a boolean flag is made available.

https://github.com/pallets/jinja/issues/2016

fixes #2016 

👋 This is my first time contributing, please advise if an entry to CHANGES.rst should be added and which version it should be associated (e.g. 3.2.0).

### DIFF
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -810,7 +810,11 @@ def do_urlize(
 
 
 def do_indent(
-    s: str, width: t.Union[int, str] = 4, first: bool = False, blank: bool = False
+    s: str,
+    width: t.Union[int, str] = 4,
+    first: bool = False,
+    blank: bool = False,
+    CRLF: bool = False,
 ) -> str:
     """Return a copy of the string with each line indented by 4 spaces. The
     first line and blank lines are not indented by default.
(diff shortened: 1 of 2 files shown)


davidism:
The issue points out that the environment already has a setting for the line ending. That should be used instead of adding a filter argument.

jmbraben:
As stated in #2016 , the changes addressed my use case. Was any search for any other instances of hardcoded use of '\n'?

davidism:
> Was any search for any other instances of hardcoded use of '\n' made?

I don't think so. If you want to look, and find other places that should use the env setting, that would be another good issue to open.

jmbraben:
There is nothing else in filters.py, the only thing in /src I'm questioning is:
https://github.com/pallets/jinja/blob/ada0a9a6fc265128b46949b5144d2eaa55e6df2c/src/jinja2/compiler.py#L457
But as in the compiler.py, thinking it is not an issue.

davidism:
You're right, that's not an issue.

jmbraben:
Is there any timeline to merge this?

mjeromin:
@jmbraben We confirmed the fix in the issue #2016, I'll go ahead and merge.

PR #2037, opened 2024-10-10
Add option to preserve comments when parsing templates

### LINKED ISSUE #1967: Option to preserve comments in the AST
Hello, thanks for this wonderful piece of software :slightly_smiling_face: 

I'm maintaining [mkdocstrings](https://github.com/mkdocstrings/mkdocstrings), which uses Jinja templates to render API documentation for different languages. Recently I had this idea of doing the same thing (collecting and rendering API docs) for... Jinja templates themselves :stuck_out_tongue: It's just an idea for now, but I'll be tracking my progress here: https://github.com/mkdocstrings/mkdocstrings/issues/661.

I see that Jinja provides the `Environment.parse` method, which builds an AST of the Jinja source: this is *fantastic*. I'll be able to visit such trees to build useful information and store it in structures that make it easy to render them in other Jinja templates.

**However...** Jinja comments are not preserved :cry: This https://github.com/pallets/jinja/issues/719#issuecomment-303933811 says that they're probably discarded by the lexer.

I can already extract lots of useful information from templates, but I had in mind to allow template writers to use Jinja comments to document variables, filters, blocks, macros, whatever, directly in the template, just like we document attributes, functions, classes, etc., directly within Python code.

For this I would need an option to tell `Environment.parse` to preserve the comments in the generated AST. Do you think that is feasible? I can definitely free up some of my time to work on this.

In any case, what do you think of this idea of auto-documentation for Jinja templates? I think that would be a wonderful tool in the Jinja ecosystem :smile: I personally have *lots* of Jinja templates that I would document this way :slightly_smiling_face:

pawamoy:
~~Add the `preserve_comments` parameter to `Environment.parse` to preserve comments in template ASTs.~~

Create a new `Comment` node, and always preserve those in ASTs obtained from `Environment.parse`.

Fixes #1967.

- [x] Add tests that demonstrate the correct behavior of the change. Tests
- [x] Add an entry in CHANGES.rst summarizing the change and linking to the issue.

### DIFF
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -600,6 +600,7 @@ class Environment:
         source: str,
         name: t.Optional[str] = None,
         filename: t.Optional[str] = None,
+        preserve_comments: bool = False,
     ) -> nodes.Template:
         """Parse the sourcecode and return the abstract syntax tree.  This
         tree of nodes is used by the compiler to convert the template into
@@ -610,15 +611,21 @@ class Environment:
         this gives you a good overview of the node tree generated.
         """
         try:
-            return self._parse(source, name, filename)
+            return self._parse(source, name, filename, preserve_comments)
         except TemplateSyntaxError:
             self.handle_exception(source=source)
 
     def _parse(
-        self, source: str, name: t.Optional[str], filename: t.Optional[str]
+        self,
+        source: str,
+        name: t.Optional[str],
+        filename: t.Optional[str],
+        preserve_comments: bool = False,
     ) -> nodes.Template:
         """Internal parsing function used by `parse` and `compile`."""
-        return Parser(self, source, name, filename).parse()
+        return Parser(
+            self, source, name, filename, preserve_comments=preserve_comments
+        ).parse()
 
     def lex(
         self,
@@ -663,12 +670,13 @@ class Environment:
         name: t.Optional[str],
         filename: t.Optional[str] = None,
         state: t.Optional[str] = None,
+        preserve_comments: bool = False,
     ) -> TokenStream:
         """Called by the parser to do the preprocessing and filtering
         for all the extensions.  Returns a :class:`~jinja2.lexer.TokenStream`.
         """
         source = self.preprocess(source, name, filename)
-        stream = self.lexer.tokenize(source, name, filename, state)
+        stream = self.lexer.tokenize(source, name, filename, state, preserve_comments)
 
         for ext in self.iter_extensions():
             stream = ext.filter_stream(stream)  # type: ignore
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -609,6 +609,9 @@ class Environment:
 
         If you are :ref:`developing Jinja extensions <writing-extensions>`
         this gives you a good overview of the node tree generated.
+
+        .. versionchanged:: 3.2
+            Added `preserve_comments` parameter.
         """
         try:
             return self._parse(source, name, filename, preserve_comments)
--- a/src/jinja2/environment.py
+++ b/src/jinja2/environment.py
@@ -600,7 +600,6 @@ class Environment:
         source: str,
         name: t.Optional[str] = None,
         filename: t.Optional[str] = None,
-        preserve_comments: bool = False,
     ) -> nodes.Template:
         """Parse the sourcecode and return the abstract syntax tree.  This
         tree of nodes is used by the compiler to convert the template into
@@ -609,12 +608,9 @@ class Environment:
 
         If you are :ref:`developing Jinja extensions <writing-extensions>`
         this gives you a good overview of the node tree generated.
-
-        .. versionchanged:: 3.2
-            Added `preserve_comments` parameter.
         """
         try:
-            return self._parse(source, name, filename, preserve_comments)
+            return self._parse(source, name, filename)
         except TemplateSyntaxError:
             self.handle_exception(source=source)
 
@@ -623,12 +619,9 @@ class Environment:
         source: str,
         name: t.Optional[str],
         filename: t.Optional[str],
-        preserve_comments: bool = False,
     ) -> nodes.Template:
         """Internal parsing function used by `parse` and `compile`."""
-        return Parser(
-            self, source, name, filename, preserve_comments=preserve_comments
-        ).parse()
+        return Parser(self, source, name, filename).parse()
 
     def lex(
         self,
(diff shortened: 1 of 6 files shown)


src/jinja2/environment.py
```diff
@@ -600,6 +600,7 @@ def parse(
         source: str,
         name: t.Optional[str] = None,
         filename: t.Optional[str] = None,
+        preserve_comments: bool = False,
```
pawamoy, line 603: suggestion: We could make it a keyword-only parameter for easier backward-compatibility maintenance in the future.

src/jinja2/lexer.py
```diff
@@ -146,17 +146,22 @@
...
         TOKEN_COMMENT_BEGIN,
         TOKEN_COMMENT,
         TOKEN_COMMENT_END,
-        TOKEN_WHITESPACE,
         TOKEN_LINECOMMENT_BEGIN,
         TOKEN_LINECOMMENT_END,
         TOKEN_LINECOMMENT,
     ]
 )
+ignored_tokens = frozenset(
```
pawamoy, line 159: note: `ignored_tokens` kept unchanged for backward-compatibility. `comment_tokens` is later subtracted from it depending on whether comments must be preserved or not.

src/jinja2/lexer.py
```diff
@@ -607,22 +612,37 @@ def tokenize(
         name: t.Optional[str] = None,
         filename: t.Optional[str] = None,
         state: t.Optional[str] = None,
+        preserve_comments: bool = False,
```
pawamoy, line 615: suggestion: We could make it a keyword-only parameter for easier backward-compatibility maintenance in the future.

src/jinja2/lexer.py
```diff
@@ -607,22 +612,37 @@ def tokenize(
...
+        return TokenStream(
+            self.wrap(stream, name, filename, preserve_comments), name, filename
+        )
 
     def wrap(
         self,
         stream: t.Iterable[t.Tuple[int, str, str]],
         name: t.Optional[str] = None,
         filename: t.Optional[str] = None,
+        preserve_comments: bool = False,
```
pawamoy, line 632: suggestion: We could make it a keyword-only parameter for easier backward-compatibility maintenance in the future.

src/jinja2/lexer.py
```diff
@@ -631,6 +651,10 @@ def wrap(
                 token = TOKEN_BLOCK_BEGIN
             elif token == TOKEN_LINESTATEMENT_END:
                 token = TOKEN_BLOCK_END
+            elif token == TOKEN_LINECOMMENT_BEGIN:
+                token = TOKEN_COMMENT_BEGIN
+            elif token == TOKEN_LINECOMMENT_END:
+                token = TOKEN_COMMENT_END
```
pawamoy, line 657: note: Possible (small?) performance hit. Not sure how to avoid it.

src/jinja2/parser.py
```diff
@@ -57,9 +57,12 @@ def __init__(
         name: t.Optional[str] = None,
         filename: t.Optional[str] = None,
         state: t.Optional[str] = None,
+        preserve_comments: bool = False,
```
pawamoy, line 60: suggestion: We could make it a keyword-only parameter for easier backward-compatibility maintenance in the future.

src/jinja2/parser.py
```diff
@@ -57,9 +57,12 @@ def __init__(
...
     ) -> None:
         self.environment = environment
-        self.stream = environment._tokenize(source, name, filename, state)
+        self.stream = environment._tokenize(
+            source, name, filename, state, preserve_comments
+        )
```
pawamoy, line 65: note: We only pass the `preserve_comments` value to the lexer, and do not store it as an instance attribute. Previously, I had `subparse` check `self.preserve_comments`, but I think  it's more robust to simply rely on what the lexer yields.

src/jinja2/parser.py
```diff
@@ -1025,6 +1028,11 @@ def flush_data() -> None:
                     else:
                         body.append(rv)
                     self.stream.expect("block_end")
+                elif token.type == "comment_begin":
```
pawamoy, line 1031: note: As mentioned in previous note, we don't check for a possible `self.preserve_comments` instance attribute value. Instead we rely on what the lexer yields.

pawamoy:
Not sure to understand the readthedocs build failure.

src/jinja2/nodes.py
```diff
@@ -715,6 +715,13 @@ def as_const(self, eval_ctx: t.Optional[EvalContext] = None) -> t.Any:
         return self.expr2.as_const(eval_ctx)
 
 
+class Comment(Stmt):
+    """A template comment."""
```
pawamoy, line 719: docs: Could maybe add `.. versionadded:: 3.2`, is that a valid directive?

davidism:
The docs failure is unrelated and something I'm actively working on, you can ignore it.

What if we just added these nodes to the AST unconditionally? I'm not sure that would be a breaking change, as an existing walker would skip those nodes.

pawamoy:
> What if we just added these nodes to the AST unconditionally? I'm not sure that would be a breaking change, as an existing walker would skip those nodes. I'd rather not have more configuration if possible.

Totally fine by me! I definitely understand not wanting more configuration. I'll run a few code searches on GitHub to see if this would break anything, and then do the change.

pawamoy:
This would break jinja2schema: https://github.com/aromanovich/jinja2schema/tree/master.

```pytb
  File "/media/data/dev/jinja/.venv/lib/python3.12/site-packages/jinja2schema/core.py", line 49, in infer_from_ast
    rv = visit(ast, {}, config)
         ^^^^^^^^^^^^^^^^^^^^^^
  File "/media/data/dev/jinja/.venv/lib/python3.12/site-packages/jinja2schema/visitors/util.py", line 20, in visit
    structure = visit_many(node.body, macroses, config)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/media/data/dev/jinja/.venv/lib/python3.12/site-packages/jinja2schema/visitors/util.py", line 37, in visit_many
    structure = visit(node, macroses, config, predicted_struct_cls, return_struct_cls)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/media/data/dev/jinja/.venv/lib/python3.12/site-packages/jinja2schema/visitors/util.py", line 14, in visit
    structure = visit_stmt(node, macroses, config)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/media/data/dev/jinja/.venv/lib/python3.12/site-packages/jinja2schema/visitors/stmt.py", line 53, in visit_stmt
    raise Exception('stmt visitor for {0} is not found'.format(type(ast)))
Exception: stmt visitor for <class 'jinja2.nodes.Comment'> is not found
```

pawamoy:
Other than that, almost all uses of `env.parse` I find on GitHub are just to pass the AST to `meta.find_undeclared_variables`.

pawamoy:
I've pushed a fixup commit that removes the new parameter and changes the code so that comments are always preserved in ASTs. Makes the overall change much smaller :slightly_smiling_face:

PR #2038, opened 2024-10-17
Update pyproject.toml Don't use the newest vesion of MarkUpSafe

BjorklundBenjamin:
MarkUpSafe was updated to have a version 3 it's still a bit unstable and we should wait to use it


ThiefMaster:
No. Pin it on your side - libraries should never pin upper bounds of dependencies unless they are completely incompatible.

PR #2040, opened 2024-10-23
Fix typo in templates docs

kkirsche:
This fixes a minor typo in the parameters to ngettext in the templates docs.

I left out changes.rst from this as it's just a single word, and I don't think it necessarily justifies a changelog entry


davidism:
duplicate of #1881

PR #2058, opened 2024-12-11
Update templates.rst to fix "per default blocks" typo

b9chris:


kevin-brown (approved):

davidism:
"per default" is another way to say "by default". This PR changes the meaning rather than only the wording. "default content" is not what's meant here, and doesn't really make sense. I'll make a separate PR making the wording more appropriate for this page.

PR #2061, opened 2024-12-19
fix default for `Environment.overlay(enable_async)`

SamyCookie:
Got a wrong behaviour with `Environment.overlay` and seems to me an obvious error code.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -425,3 +425,11 @@ class TestLowLevel:
         env = CustomEnvironment()
         tmpl = env.from_string("{{ foo }}")
         assert tmpl.render() == "resolve-foo"
+
+
+def test_overlay_enable_async(env):
+    assert not env.is_async
+    assert not env.overlay().is_async
+    env_async = env.overlay(enable_async=True)
+    assert env_async.is_async
+    assert not env_async.overlay(enable_async=False).is_async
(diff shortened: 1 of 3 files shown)


kevin-brown:
Do you have a test case that happens to reproduce the issue that this fixes?

SamyCookie:
Very simple : 
```python
env = Environment(enable_async=True)
assert env.is_async == env.overlay().is_async, "overlay async state differs with parent"
```

davidism:
@SamyCookie I appreciate that you've submitted a fix, but please follow our contributing guide next time. You didn't open an issue, or explain the PR what the issue was and what the fix does. Your PR doesn't contain a test. While brevity is appreciated, you need to provide _some_ detail rather than just dumping something on us and saying "it's obvious".

davidism:
Rebased to target bug fix at stable instead of main. Added test. Added changelog.

tests/test_api.py
```diff
@@ -425,3 +425,11 @@ class CustomEnvironment(Environment):
         env = CustomEnvironment()
         tmpl = env.from_string("{{ foo }}")
         assert tmpl.render() == "resolve-foo"
+
+
+def test_overlay_enable_async(env):
+    assert not env.is_async
+    assert not env.overlay().is_async
+    env_async = env.overlay(enable_async=True)
+    assert env_async.is_async
```
SamyCookie, line 434: Missing test case of the concerning issue with `overlay` without any `enable_async` parameter , that is : 
```python
    assert env_async.overlay().is_async
 ```

PR #2076, opened 2025-03-05
Support both os.path-based separators when traversing looking for templates

magul:
This PR ensures proper handling of file paths when passing os.path.join(path, file) to the template loader, specifically improving support for Windows environments.


davidism:
Please review many previous discussions explaining that template names are not filesystem paths. They happen to look like paths when using the filesystem loader, but are normalized to use forward slashes.

magul:
Oops, I struggled a bit, but I finally found it: https://github.com/pallets/jinja/pull/1579.
Thanks!

PR #2093, opened 2025-05-20
Updated dependency management to uv

aenglander:
Updated dependency management to uv

uv required update to Python 3.9, so I updated the projectt to 3.9+. After moving to 3.9+, style check required that primitives be used rather than their typing aliases.

All projects are being migrated to uv for development dependency management.


davidism:
Needs readthedocs and pre-commit-config copied over as well.

PR #2098, opened 2025-05-28
deprecate `__version__`

davidism:
As has been done with the other Pallets projects.

The `__version__` attribute is an old pattern from early in Python packaging. Setuptools eventually made it easier to use the pattern by allowing reading the value from the attribute at build time, and some other build backends have done the same.

However, there's no reason to expose this directly in code anymore. It's usually better to use feature detection (`hasattr`, `try/except`) instead. `importlib.metadata.version("jinja2")` can be used to get the version at runtime in a standard way, if it's really needed.


davidism:
Fixed project name in 9c11dd6

PR #2104, opened 2025-06-13
Use the currently handled exception as UndefinedError.__context__

### LINKED ISSUE #2103: Make {{ obj.bad_attr }} preserve the AttributeError's traceback
Environment.getattr() and Environment.getitem() discard errors
from (`TypeError`, `LookupError`, `AttributeError`) and return Undefined instead.

When the Undefined raises `UndefinedError`, the original exception,
and its traceback, is lost. This makes debugging hard.

## My use case

The CPython [buildbot release dashboard](https://buildbot.python.org/#/release_status)
uses lazily computed attributes (with [`cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property)).
In other words, “business logic” is driven by the template rendering.
This avoids additional logic to pre-determine which data the template will need.
Unfortunately, any `AttributeError` that occurs deep in application code bubbles
up and is replaced by a traceback-less `UndefinedError`.

## Possible workarounds outside Jinja

A hacky workaround is [a decorator that replaces `AttributeError` by a different kind of error](https://github.com/python/buildmaster-config/blob/123e57d5d84f8815d52bfa4950b305c342da3e65/master/custom/release_dashboard.py#L35-L49).
(The linked code doesn't use it -- I apply it only when debugging.)

It is also possible to subclass `Undefined` to preserve the exception:

```py
class StrictUndefinedWithCause(jinja2.StrictUndefined):
    __slots__ = ['_original_exception']
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._original_exception = sys.exception()

    def _fail_with_undefined_error(self, *args, **kwargs):
        try:
            super()._fail_with_undefined_error(*args, **kwargs)
        except BaseException as exc:
            raise exc from self._original_exception
```

This would work well if `Environment.getitem` and `Environment.getattr`
always called their `self.undefined(...)` inside `except` blocks.
That's not always the case -- and that's the main reason I can't find
a good solution without changes in Jinja.

I suggest that Jinja itself adds the relevant tracebacks when it raises
`UndefinedError`, and adds tests to ensure it does that.

encukou:
*At this point, please treat this PR as a “feasibility study” for issue #2103. There are other ways to fix this.*

---

Add ``Undefined._undefined_cause`` attribute, which is used as the default ``__context__`` for exceptions raised by that ``Undefined``.
When an ``Undefined`` is created, automatically set this attribute to the currently handled exception (or ``None``). This mirrors Python's own exception chaining.

Also, restructure `Environment.getitem` and `Environment.getattr` to use exception chaning to preserve the exceptions they encounter.

Fixes: #2103

- [x] TODO: add appropriate `.. versionadded::`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,10 @@ Unreleased
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`1793`
 -   Use ``flit_core`` instead of ``setuptools`` as build backend.
+-   When ``Undefined`` is created in an ``except`` block, the handled
+    exception is stored in a new  attribute, ``_undefined_context``,
+    and used as the default ``__context__`` for errors raised by
+    ``_fail_with_undefined_error``.
 
 
 Version 3.1.6
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,7 @@ Unreleased
     exception is stored in a new  attribute, ``_undefined_context``,
     and used as the default ``__context__`` for errors raised by
     ``_fail_with_undefined_error``.
+    :issue:`2103`
 
 
 Version 3.1.6
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -289,6 +289,9 @@ others fail.
 The closest to regular Python behavior is the :class:`StrictUndefined` which
 disallows all operations beside testing if it's an undefined object.
 
+When :class:`Undefined` is created in an ``except`` or ``finally`` clause,
+
+
 .. autoclass:: jinja2.Undefined()
 
     .. attribute:: _undefined_hint
@@ -310,6 +313,15 @@ disallows all operations beside testing if it's an undefined object.
 
         The exception that the undefined object wants to raise.  This
         is usually one of :exc:`UndefinedError` or :exc:`SecurityError`.
+
+    .. attribute:: _undefined_context
+
+        The default ``__context__`` for exceptions raised when operations
+        on this undefined value fail.
+
+        When :class:`Undefined` is created while an exception is being
+        handled (for example, inside an ``except``  clause),
+        ``_undefined_context`` is automatically set to the handled exception.
 
     .. method:: _fail_with_undefined_error(\*args, \**kwargs)
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -289,7 +289,6 @@ others fail.
 The closest to regular Python behavior is the :class:`StrictUndefined` which
 disallows all operations beside testing if it's an undefined object.
 
-
 .. autoclass:: jinja2.Undefined()
 
     .. attribute:: _undefined_hint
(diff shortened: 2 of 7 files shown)


docs/api.rst
```diff
@@ -289,6 +289,9 @@ others fail.
 The closest to regular Python behavior is the :class:`StrictUndefined` which
 disallows all operations beside testing if it's an undefined object.
 
+When :class:`Undefined` is created in an ``except`` or ``finally`` clause,
```
davidism, line 292: Did you mean to finish this line, or did you move it further down and forget to remove it here?

CHANGES.rst
```diff
@@ -13,6 +13,10 @@ Unreleased
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`1793`
 -   Use ``flit_core`` instead of ``setuptools`` as build backend.
+-   When ``Undefined`` is created in an ``except`` block, the handled
+    exception is stored in a new  attribute, ``_undefined_context``,
+    and used as the default ``__context__`` for errors raised by
+    ``_fail_with_undefined_error``.
```
davidism, line 19: Link to the issue you opened using `` :issue:`...` ``.

src/jinja2/runtime.py
```diff
@@ -856,7 +861,10 @@ def _fail_with_undefined_error(
         """Raise an :exc:`UndefinedError` when operations are performed
         on the undefined value.
         """
-        raise self._undefined_exception(self._undefined_message)
+        exception = self._undefined_exception(self._undefined_message)
+        if exception.__context__ is None:
```
davidism, line 865: Do we need to consider `__cause__` as well?

davidism:
You might need to modify `SandboxedEnvironment` or `NativeEnvironment` `getattr`/`getitem` as well.

docs/api.rst
encukou: The latter. Thanks for noticing!

src/jinja2/runtime.py
encukou: No.
Setting `__cause__` will hide `__context__` in the default (“linear”) traceback display, but both are available for more in-depth debugging. (See [docs](https://docs.python.org/3/library/exceptions.html#exception-context)).

encukou (commented):
> This all makes sense to me. Were you planning to propose another idea as well?

Thanks!
If the issue looks fix-worthy, there are a few questions best answered by the project maintainer.

Should this be an explicit keyword argument for `Undefined`? (Or should there be an *additional* argument for setting `__cause__` -- that one is set using explicit syntax in Python?)

Some projects treat tested behaviour as promises. Should there be some disclaimer around `raises_cause_chain` that jijnja doesn't promise these exact chains -- we only want to ensure the tracebacks get shown to the user?

Rowlando13:
@davidism Talked with Petr at Pycon. Would be nice to get this merged if possible. Petr has gotten the pr as good as he can with the knowledge that he has for jinja. I can cut a release if that would be helpful.

PR #2115, opened 2025-09-11
Changed the static annotation of default_for_string and default parameter of the select_autoescape method in utils.py file

subhajitsaha01:
Changed the static type annotations of the default_for_string and default parameters of the select_autoescape method of the utils.py file. The static type annotation for default_for_string parameter has been changed from `bool = True` to `bool | str = True` and for the default parameter has been changed from `bool = False` to `bool | str = False`. The change has been made because string values has been passed to both the parameters in test cases file i.e. test_utils.py. The references for the codes has been given below:

https://github.com/pallets/jinja/blob/5ef70112a1ff19c05324ff889dd30405b1002044/src/jinja2/utils.py#L581
https://github.com/pallets/jinja/blob/5ef70112a1ff19c05324ff889dd30405b1002044/tests/test_utils.py#L119

### DIFF
--- a/src/jinja2/utils.py
+++ b/src/jinja2/utils.py
@@ -581,8 +581,8 @@ class LRUCache:
 def select_autoescape(
     enabled_extensions: t.Collection[str] = ("html", "htm", "xml"),
     disabled_extensions: t.Collection[str] = (),
-    default_for_string: bool = True,
-    default: bool = False,
+    default_for_string: bool | str = True,
+    default: bool | str = False,
 ) -> t.Callable[[str | None], bool]:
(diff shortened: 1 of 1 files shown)


davidism:
This doesn't appear to be correct. It seems like the test is just using those values to ensure the default is passed through rather than producing a boolean in some other way. Is that useful for this test? Probably not, but it doesn't mean the type is wrong.

Please stop using AI to try to find/fix issues unasked for. Whatever tool you're using is clearly giving false positives, and you're clearly not thinking about the changes it's making. You also ignored the feedback from Flask to make such changes to the stable branch.

PR #2135, opened 2026-02-11
Add type annotation for FILTERS dictionary

### LINKED ISSUE #2120: Missing type annotation for FILTERS
When using Jinja2 with filters, Pylance is screaming at me because `Environment.filters` is partially unknown.

Environment.filters is copied from here: https://github.com/pallets/jinja/blob/5ef70112a1ff19c05324ff889dd30405b1002044/src/jinja2/filters.py#L1812

`FILTERS` does not have a type annotation.

You already defined a TypeVar for the expected signature here:

https://github.com/pallets/jinja/blob/5ef70112a1ff19c05324ff889dd30405b1002044/src/jinja2/filters.py#L44

Is there a specific reason to leave out the annotation? I'd be happy to submit a PR.

```python
# Suggested code
FILTERS: dict[str, F] = {
```

Best,
Gab

Environment:

- Python version: 3.12
- Jinja version: 3.1.6
- I searched for other issues with the keywords `type` and `filter`
- I tested the suggested changes locally by modifying jinja in my venv

Teylersf:
This PR adds an explicit type annotation t.Dict[str, t.Callable[..., t.Any]] to the FILTERS dictionary to improve type checking and IDE support.

Fixes #2120

### DIFF
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -1,4 +1,4 @@
-"""Built-in template filters used with the ``|`` operator."""
+﻿"""Built-in template filters used with the ``|`` operator."""
 
 import math
 import random
@@ -1808,8 +1808,7 @@ async def async_select_or_reject(
             if func(item):
                 yield item
 
-
-FILTERS = {
+FILTERS: t.Dict[str, t.Callable[..., t.Any]] = {
     "abs": abs,
     "attr": do_attr,
     "batch": do_batch,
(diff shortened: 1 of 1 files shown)


src/jinja2/filters.py
```diff
@@ -1,4 +1,4 @@
-"""Built-in template filters used with the ``|`` operator."""
+﻿"""Built-in template filters used with the ``|`` operator."""
```
davidism, line 1: Why did you change this line? What is the change?

Teylersf:
@davidism Thanks for catching that! That ? was added accidentally — it wasn't intentional. I'll revert that line back to the original docstring. Appreciate the sharp eye during review!

Teylersf:
@davidism The stray character has been fixed by pre-commit.ci in commit

davidism:
What "?" are you referring to. I see that pre-commit reverted something, but it looks exactly the same in both diffs.

PR #2137, opened 2026-02-20
Fix DeprecationWarning for invalid escape sequences in template strings

### LINKED ISSUE #1156: invalid escape sequence DeprecationWarning from template syntax itself
## mcve

(this is extracted from a larger example which was rendering actual `.j2` files -- however this was the shortest example I could come up with)

```python
from jinja2 import Template

print(Template(r'{{ "\d" }}').render())
```

## Expected Behavior

I'm unclear on the expected behaviour, there's kinda 3 options that I see as reasonable (unranked since I don't really have an opinion here)
- warning: produce a warning, but with better indication on the user's fix (especially: point at the actual filename where the problem is)
- error: make invalid escape sequences in templates an error
- (silence) accept that string tokens in jinja source can have escape sequences that do not comply with python's rules

## Actual Behavior

```console
$ python3 -Wonce t.py
/tmp/x/venv/lib/python3.6/site-packages/jinja2/lexer.py:606: DeprecationWarning: invalid escape sequence '\d'
  .decode("unicode-escape")
\d
```

## Full Traceback

This can be converted into an error with `-Werror`

```console
$ python3 -Werror t.py
DeprecationWarning: invalid escape sequence '\d'

The above exception was the direct cause of the following exception:

  File "/tmp/x/venv/lib/python3.6/site-packages/jinja2/lexer.py", line 606, in wrap
    .decode("unicode-escape")
DeprecationWarning: decoding with 'unicode-escape' codec failed (DeprecationWarning: invalid escape sequence '\d')

During handling of the above exception, another exception occurred:

  File "t.py", line 3, in <module>
    print(Template(r'{{ "\d" }}').render())
  File "/tmp/x/venv/lib/python3.6/site-packages/jinja2/environment.py", line 986, in __new__
    return env.from_string(source, template_class=cls)
  File "/tmp/x/venv/lib/python3.6/site-packages/jinja2/environment.py", line 896, in from_string
    return cls.from_code(self, self.compile(source), globals, None)
  File "/tmp/x/venv/lib/python3.6/site-packages/jinja2/environment.py", line 625, in compile
    self.handle_exception(source=source_hint)
  File "/tmp/x/venv/lib/python3.6/site-packages/jinja2/environment.py", line 787, in handle_exception
  File "<unknown>", line 1, in template
jinja2.exceptions.TemplateSyntaxError: invalid escape sequence '\d')
```

## Your Environment

```console
$ python --version --version
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
$ pip freeze --all
Jinja2==3.0.0a1
MarkupSafe==1.1.1
pip==20.0.2
setuptools==45.2.0
wheel==0.34.2
```

___

The "correct" fix for those hitting the same issue:

making a raw string is not valid syntax in jinja so that is not an option:

```diff
$ diff -u t.py t.py.new
--- t.py	2020-02-18 09:53:02.892139615 -0800
+++ t.py.new	2020-02-18 09:52:57.860139615 -0800
@@ -1,3 +1,3 @@
 from jinja2 import Template
 
-print(Template(r'{{ "\d" }}').render())
+print(Template(r'{{ r"\d" }}').render())
```

```console
$ python3 t.py.new
...
jinja2.exceptions.TemplateSyntaxError: expected token 'end of print statement', got 'string'
```

the "fix" I needed was to manually escape the backslashes:

```diff
$ diff -u t.py t.py.new
--- t.py	2020-02-18 09:53:02.892139615 -0800
+++ t.py.new	2020-02-18 09:53:57.068139615 -0800
@@ -1,3 +1,3 @@
 from jinja2 import Template
 
-print(Template(r'{{ "\d" }}').render())
+print(Template(r'{{ "\\d" }}').render())
```

```console
$ python3 t.py.new
\d
```

danielalanbates:
## Summary

Fixes #1156 

When a Jinja2 template string contains an escape sequence that is not a valid Python escape sequence (e.g., `\d`, `\p`, `\s`), the lexer currently triggers a `DeprecationWarning` from Python's `unicode-escape` codec. On newer Python versions this warning is becoming an error, breaking templates that use regex patterns or other backslash sequences in string literals.

This PR replaces the `.encode("ascii", "backslashreplace").decode("unicode-escape")` approach with a custom `_decode_escapes()` function that:

- Handles all standard escape sequences: `\\`, `\'`, `\"`, `\a`, `\b`, `\f`, `\n`, `\r`, `\t`, `\v`, `\0`, `\xNN`, `\uNNNN`, `\UNNNNNNNN`, `\N{name}`
- Keeps unrecognized escape sequences (like `\d`) as literal backslash + character, matching user expectations for template strings
- Still raises `TemplateSyntaxError` for malformed recognized sequences (e.g., unknown Unicode character names)

### Before (with `-Werror`):
```python
>>> Template(r'{{ "\d" }}').render()
TemplateSyntaxError: invalid escape sequence '\d'
```

### After:
```python
>>> Template(r'{{ "\d" }}').render()
'\\d'
```

## Test plan

- [x] All 911 existing tests pass
- [x] The `test_string_escapes` test continues to pass, verifying `\0`, `\u2668`, `\xe4`, `\t`, `\r`, `\n`, and `\N{HOT SPRINGS}` all work correctly
- [x] Verified manually with `-Werror` that `\d` no longer triggers any warning
- [x] Tested on Python 3.14 where invalid escape sequences in Python source are already errors

*This PR was created with the assistance of Claude Opus 4.6 by Anthropic. Happy to make any adjustments! Reviewed and submitted by a human.*

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,11 @@ Version 3.2.0
 
 Unreleased
 
+-   Replace the ``unicode-escape`` codec with a custom escape sequence
+    decoder for template strings. Unrecognized escape sequences such as
+    ``\d`` no longer produce ``DeprecationWarning`` (or errors on newer
+    Python versions), and instead keep the backslash literally.
+    :issue:`1156`
 -   Drop support for Python 3.7, 3.8, and 3.9.
 -   Update minimum MarkupSafe version to >= 3.0.
 -   Update minimum Babel version to >= 2.17.
--- a/src/jinja2/lexer.py
+++ b/src/jinja2/lexer.py
@@ -711,9 +711,7 @@ class Lexer:
             elif token == TOKEN_STRING:
                 # try to unescape string
                 try:
-                    value = _decode_escapes(
-                        self._normalize_newlines(value_str[1:-1])
-                    )
+                    value = _decode_escapes(self._normalize_newlines(value_str[1:-1]))
                 except Exception as e:
                     msg = str(e).split(":")[-1].strip()
                     raise TemplateSyntaxError(msg, lineno, name, filename) from e
--- a/src/jinja2/lexer.py
+++ b/src/jinja2/lexer.py
@@ -6,6 +6,7 @@ template code and python code in expressions.
 
 import re
 import typing as t
+import unicodedata
 from ast import literal_eval
 from collections import deque
 from sys import intern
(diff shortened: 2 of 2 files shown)


danielalanbates:
Thanks for taking a look. Let me know if there's anything I can improve for a future submission.

PR #2141, opened 2026-02-25
Add type annotation for filters

### LINKED ISSUE #2120: Missing type annotation for FILTERS

Gabgobie:
A minimal PR adding a type annotation to `FILTERS` as discussed in #2120

The change was tested in my venv and fixed my project's issues.
You should be aware that it causes the following Pylance warning in the jinja codebase:

> Type variable "F" has no meaning in this context Pylance([reportGeneralTypeIssues](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportGeneralTypeIssues.md))

Looking into this a little more, my understanding is that `TypeVar`s should really only be used for generic `function` and `class` declarations. https://github.com/microsoft/pyright/discussions/9826#discussioncomment-12077223

A type alias may be more appropriate in line 44. https://github.com/pallets/jinja/blob/5ef70112a1ff19c05324ff889dd30405b1002044/src/jinja2/filters.py#L44

```python
FilterFunction = t.Callable[..., t.Any]
```

Additionally, I was unable to find any use of `F` from line 44 and ran `pytest` with the line commented out. I am fairly certain that it was dead code until now.

---

#2135
#2139

The above PRs tried to address #2120 in a different way.

fixes #2120

> Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.

I don't think I can test this without adding a Pylance action, which would fail anyways since jinja is not written with type checkers in mind.

> Add or update relevant docs, in the docs folder and in code.

We could be more explicit about the requirements in [Custom Filters](https://jinja.palletsprojects.com/en/stable/api/#custom-filters) but this doesn't really change anything about them. Python type annotations _usually_ serve as documentation only.

## Other considerations

Aside from annotating `FILTERS`, this same annotation could be applied to https://github.com/pallets/jinja/blob/5ef70112a1ff19c05324ff889dd30405b1002044/src/jinja2/environment.py#L352

During my search for use of line 44, I noticed that the same definition of `F` is declared five times. There may be a semantic difference between the otherwise identical declarations.

### DIFF
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -41,7 +41,7 @@ if t.TYPE_CHECKING:
             pass
 
 
-F = t.TypeVar("F", bound=t.Callable[..., t.Any])
+FilterFunction = t.Callable[..., t.Any]
 K = t.TypeVar("K")
 V = t.TypeVar("V")
 
@@ -1815,7 +1815,7 @@ async def async_select_or_reject(
                 yield item
 
 
-FILTERS: t.Dict[str, F] = {
+FILTERS: t.Dict[str, FilterFunction] = {
     "abs": abs,
     "attr": do_attr,
     "batch": do_batch,
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -1815,7 +1815,7 @@ async def async_select_or_reject(
                 yield item
 
 
-FILTERS = {
+FILTERS: t.Dict[str, F] = {
     "abs": abs,
     "attr": do_attr,
     "batch": do_batch,
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -41,7 +41,7 @@ if t.TYPE_CHECKING:
             pass
 
 
-F = t.TypeVar("F", bound=t.Callable[..., t.Any])
+FilterFunction = t.Callable[..., t.Any]
 K = t.TypeVar("K")
 V = t.TypeVar("V")
 
@@ -1815,7 +1815,7 @@ async def async_select_or_reject(
                 yield item
 
 
-FILTERS = {
+FILTERS: t.Dict[str, FilterFunction] = {
     "abs": abs,
     "attr": do_attr,
     "batch": do_batch,
--- a/src/jinja2/tests.py
+++ b/src/jinja2/tests.py
@@ -12,6 +12,9 @@ if t.TYPE_CHECKING:
     from .environment import Environment
 
 
+TestFunction = t.Callable[..., t.Any]
+
+
 def test_odd(value: int) -> bool:
     """Return true if the variable is odd."""
     return value % 2 == 1
@@ -213,7 +216,7 @@ def test_in(value: t.Any, seq: t.Container[t.Any]) -> bool:
     return value in seq
 
 
-TESTS = {
+TESTS: t.Dict[str, TestFunction] = {
     "odd": test_odd,
     "even": test_even,
     "divisibleby": test_divisibleby,
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,14 @@
 .. currentmodule:: jinja2
 
+Version 3.1.7
+-------------
+
+Unreleased
+
+-   Add type annotation to ``FILTERS`` and ``TESTS``
+    :issue:`2120`, :pr:`2141`
+
+
 Version 3.1.6
 -------------
 
--- a/src/jinja2/__init__.py
+++ b/src/jinja2/__init__.py
@@ -35,4 +35,4 @@ from .utils import pass_environment as pass_environment
 from .utils import pass_eval_context as pass_eval_context
 from .utils import select_autoescape as select_autoescape
 
-__version__ = "3.1.6"
+__version__ = "3.1.7"


Gabgobie:
The error from `Tests / 3.7 (pull_request)` seems unrelated to me:

> Error: The version '3.7' with architecture 'x64' was not found for Ubuntu 24.04.

I'll use the older `t.Dict[str, F]` annotation to appease `typing` and `3.8`.

Gabgobie:
I think the typing error is basically what I discussed above about the use of `TypeVar`.

The 3.7 error I really don't get. Looking at the manifest, versions 3.7.1-17 are available.

davidism:
Don't worry about < 3.10.

You don't need to use the type var `F`, you're right that it doesn't make sense there. The other PRs were closed because they were LLM junk, not because they were using a particular solution vs your suggestion. I'm fine with whatever you end up figuring out the answer is.

davidism:
Also note that `TESTS` in `tests.py` needs the same fix.

src/jinja2/filters.py
```diff
@@ -41,7 +41,7 @@ def __html__(self) -> str:
             pass
 
 
-F = t.TypeVar("F", bound=t.Callable[..., t.Any])
+FilterFunction = t.Callable[..., t.Any]
```
felixfontein: Shouldn't filters accept at least one positional argument? If yes, I think it would be better to adjust this as follows:
```suggestion
P = t.ParamSpec("P")
FilterFunction = t.Callable[t.Concatenate[t.Any, P], t.Any]
```
(https://docs.python.org/3/library/typing.html#typing.ParamSpec, https://docs.python.org/3/library/typing.html#typing.Concatenate)

src/jinja2/tests.py
```diff
@@ -12,6 +12,9 @@
     from .environment import Environment
 
 
+TestFunction = t.Callable[..., t.Any]
```
felixfontein: Similarly here.

src/jinja2/filters.py
felixfontein: This requires Python 3.10+; I hope I didn't interpret "Don't worry about < 3.10" wrongly...

Gabgobie: Thanks for giving this a review and linking the docs.

I haven't used `typing.ParamSpec` and `typing.Concatenate` yet but skimming the docs and reading [PEP 612's Abstract](https://peps.python.org/pep-0612/#abstract), it seems they are useful for decorators. The functions we take here are not decorated by Jinja but just called on any given input from/to the templates.

I feel like using anything other than the ellipses would add unnecessary complexity and "risk" adding restrictions were there weren't any previously. Please feel free to correct me if I'm mistaken.

Best,

Edit: I think I get your intention now. Ensure that the function takes at least one argument. I'll get back to you when I find the time to give this some thought. I'm not sure if this would be a breaking change since I didn't verify if this is currently a hard requirement or if this would just be communicating the status quo.

davidism: I'd be interested if this worked and actually showed an error on zero arg functions. Definitely try type checking an example to make sure it works.

felixfontein: > I haven't used `typing.ParamSpec` and `typing.Concatenate` yet but skimming the docs and reading [PEP 612's Abstract](https://peps.python.org/pep-0612/#abstract), it seems they are useful for decorators. The functions we take here are not decorated by Jinja but just called on any given input from/to the templates.

It definitely *is* useful for decorators, but there are also other places - like here, if you want to describe a function type whose first parameter has a specific type (even if it's `t.Any`, which simply means it must have at least one positional argument), and can have an arbitrary number of other (positional and keyword) parameters.

> I'd be interested if this worked and actually showed an error on zero arg functions.

That's what it did in my tests with mypy.

It's Python 3.10+ though, so if Jinja2 wants to keep supporting Python < 3.10 also for type checking, it would need to go into a `if sys.version_info >= (3, 10):`, with some less strict definition for Python < 3.10.

(I personally use `from __future__ import annotations`, `if t.TYPE_CHECKING:` and restrict type checking to newer Python versions in my projects. I guess jinja2 can't just introduce that as it would be a breaking change, so `if sys.version_info >= (3, 10):` is probably a better way to use this for newer Pythons, and keep some less strict definition for older ones.)

Gabgobie: @davidism It was my original goal to write this in a way that can be merged on stable so anything >=3.7 can use `filters` (and `tests`) without using `#noqa` or type checkers screaming at the consumer. Paying <3.10 no mind as you suggested would to my understanding require a rebase against main. I don't use these old versions anyways so I don't mind either way.

@felixfontein I checked your suggestion in a minimal example. `Concatenate[Any, ...]` seems to be the way to go. Using `ParamSpec` triggers the `no meaning in this context` warning.

> ```python
> FilterFunction = t.Callable[t.Concatenate[t.Any, ...], t.Any]
> ```

That said, I'd need to know if we

- should keep this as is for stable,
- try to use the `if sys.version_info >= (3, 10):` route to get this on stable and fall back to the simple version or
- just exclusively use the new syntax and target main.

<details><summary>Demo tested against <code>pylance</code></summary>

```Python
from typing import (
    Any,
    Callable,
    Concatenate,
    List,
)

functions: List[Callable[Concatenate[Any, ...], Any]] = []

def invalid() -> Any: ...

def also_invalid(*, param1: Any) -> Any: ...

def valid(param1: Any) -> Any: ...

def also_valid(param1: Any, param2: Any) -> Any: ...

def kw_valid(param1: Any, *, param2: Any) -> Any: ...

functions.append(invalid)
# Argument of type "() -> Any" cannot be assigned to parameter "object" of type "(Any, ...) -> Any" in function "append"
#   Type "() -> Any" is not assignable to type "(Any, ...) -> Any"
#     Function accepts too many positional parameters; expected 0 but received 2

functions.append(also_invalid)
# Argument of type "(*, param1: Any) -> Any" cannot be assigned to parameter "object" of type "(Any, ...) -> Any" in function "append"
#   Type "(*, param1: Any) -> Any" is not assignable to type "(Any, ...) -> Any"
#     Function accepts too many positional parameters; expected 0 but received 2

functions.append(valid)

functions.append(also_valid)

functions.append(kw_valid)
```

</details>

Curiously, the `pylance` error message seems backwards to me but I may just be tired.

davidism: I don't care if tests < 3.10 fail. Or rebase on main.

Gabgobie: I had to touch the `pyproject.toml` to make `t.Concatenate` available at all. -> lift version requirements

- 3.7-3.9 failed as expected
- tox broke on everything else, maybe because I didn't erase 7-9 throughout the entire project but just in the toml.
- `MyPy` seems to have an entirely different interpretation of the annotation from `Pylance`. It seems to be happy with functions that only define protocols for their input but is looking for an exact signature match instead of recognizing `(t.Any, ...) -> t.Any` being a superset of whichever signature the value has on explicit definitions.

Sample error:

```
Dict entry 41 has incompatible type "str": "Callable[[str | HasHTML], str]"; expected "str": "Callable[[Any, VarArg(Any), KwArg(Any)], Any]"
```

How `"str": "Callable[[str | HasHTML], str]"` &NotSubsetEqual; `"str": "Callable[[Any, VarArg(Any), KwArg(Any)], Any]"` is beyond me. Maybe this is for some reason desirable but it's getting late for me.

Unless anyone has a good idea for what is going on with MyPy and knows how to fix that I'd suggest to roll back my last commit and say it's good enough with just `Callable[..., t.Any]`.

Just to be sure, I also branched off of main [here](https://github.com/Gabgobie/jinja/tree/annotations) but that's facing the same MyPy issue.

Edit: On second thought, MyPy probably does recognize the superset relationship but expects the value definition to be a superset of the dict annotation. Either way, this probably won't work.

felixfontein: The problem is that right now, a quite old mypy version (that still supports Python 3.8) is used. With the latest mypy it works fine. Since you updated the version ranges in pyproject.toml, you should also run `tox -e update-requirements` I guess. With that I get an entirely different set of mypy error messages (all being `Unused "type: ignore" comment`).

Gabgobie: > Just to be sure, I also branched off of main [here](https://github.com/Gabgobie/jinja/tree/annotations) but that's facing the same MyPy issue.

I switched to the annotations branch based on main and ran

1. `uv sync`
1. `tox -e update-requirements`
1. `mypy --version` (`mypy 1.15.0 (compiled: yes)`)
1. `mypy src/jinja2/filters.py` got the same mypy errors about incompatible types [dict-item]

After that, I went into `pyproject.toml` and forcibly upgraded MyPy to version `1.20.2` and everything checked out with zero errors. `1.20.2` [requires python >= 3.10](https://github.com/python/mypy/blob/v1.20.2/pyproject.toml#L50) so that should be fine.

I'll take a quick look into porting that over here. If it's more effort than it's worth, I'd close this PR in favor of a fresh one against main since going by the GH docs, I can't just switch out the base branch.

Edit: tox doing nothing was a main branch thing. On stable it upgraded to 1.20.2

davidism:
This PR is getting to be too unwieldy for a tiny typing change. Please don't update every single requirement here.

Make a PR branching off main, and just make the annotation change. I'll deal with the rest.

src/jinja2/filters.py
Gabgobie: So to report on the tests:

- version checks still choke on the tox command synax
- In PyPy, pytest fails to import conftest and blames the ellipsis in Concatenate: `E   TypeError: The last parameter to Concatenate should be a ParamSpec variable.` - We can't use `ParamSpec` because pyright would reject it as meaningless in this context. That makes me fairly sure that a PR against main wouldn't be a fix. `pytest` runs without issue for me locally. I think this is a 3.10 and maybe 3.11 issue because the 3.12 source allows the use of an ellipsis.

<details><summary><code>cpython-3.12.11-windows-x86_64-none/Lib/typing.py</code> Concatenate source</summary>

```Python
@_SpecialForm
def Concatenate(self, parameters):
    """Special form for annotating higher-order functions.

    ``Concatenate`` can be used in conjunction with ``ParamSpec`` and
    ``Callable`` to represent a higher-order function which adds, removes or
    transforms the parameters of a callable.

    For example::

        Callable[Concatenate[int, P], int]

    See PEP 612 for detailed information.
    """
    if parameters == ():
        raise TypeError("Cannot take a Concatenate of no types.")
    if not isinstance(parameters, tuple):
        parameters = (parameters,)
    if not (parameters[-1] is ... or isinstance(parameters[-1], ParamSpec)):
        raise TypeError("The last parameter to Concatenate should be a "
                        "ParamSpec variable or ellipsis.")
    msg = "Concatenate[arg, ...]: each arg must be a type."
    parameters = (*(_type_check(p, msg) for p in parameters[:-1]), parameters[-1])
    return _ConcatenateGenericAlias(self, parameters)
```

</details>

Ultimately I'd have loved to use Concatenate and indicate the param requirements to consumers but think we need to wait for Jinja to drop support up to 3.11 or 3.12. My goal was to _fix_ Jinja as a dependency causing type checking issues in consumers. Trying to get this in caused this PR to change files that it realistically has no business touching which makes the review unnecessarily complex and merging more difficult.

IMO a PR for a type hint on two constants should not upgrade dependencies, lift version requirements and change lines outside of the annotations.

Gabgobie:
Agreed though main wouldn't help, see https://github.com/pallets/jinja/pull/2141#discussion_r3196539021

Let's roll the latest changes back and go with `Callable[..., t.Any]`

PR #2144, opened 2026-03-18
Add handling logic for none to jinja-filters.default

trouver:
Add handling logic for `None` to `jinja-filters.default`. Add a parameter for the `None` type to control whether this rule applies when the value is `None`.


davidism:
Seems redundant with `boolean`.

trouver:
The `boolean` parameter treats all falsy values as `false`, but this can cause issues when `0` has a specific meaning or when only `None` should be handled exclusively.

Issue #759, opened 2017-08-22
Jinja2 fails to render template with more than 98 if/elif statements


dseira:
## Expected Behavior

No errors on rendering a template.

## Actual Behavior

Having the following jinja template (``this.jinja2`` ) with 99 ``if/elif`` statements lines; when the template is rendered it shows the following error:
```
IndentationError: too many levels of indentation
```
On the other hand, if one of the ``elif`` statements line is removed (keeping 98 ``elif``) the template is rendered correctly without any error.

It seems that there is a hard limit on the number of ``if/elif`` statements.

## Template Code
**this.jinja2**
```jinja
{% if name == 'server0' %}
{% elif name == 'server1' %}
{% elif name == 'server2' %}
{% elif name == 'server3' %}
....
{% elif name == 'server96' %}
{% elif name == 'server97' %}
{% elif name == 'server98' %}
{% endif %}
```
**this.py**
```
import os
import jinja2
def render(tpl_path, context):
    path, filename = os.path.split(tpl_path)
    return jinja2.Environment(
        loader=jinja2.FileSystemLoader(path or './')
    ).get_template(filename).render(context)

print(render('./this.jinja2', {'name': 'test'}))
```
## Full Traceback
```pytb
Traceback (most recent call last):
  File "this.py", line 9, in <module>
    print(render('./this.jinja2', {'name': 'test'}))
  File "this.py", line 7, in render
    ).get_template(filename).render(context)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 791, in get_template
    return self._load_template(name, self.make_globals(globals))
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 765, in _load_template
    template = self.loader.load(self, name, globals)
  File "/usr/lib/python2.7/dist-packages/jinja2/loaders.py", line 125, in load
    code = environment.compile(source, name, filename)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 551, in compile
    return self._compile(source, filename)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 512, in _compile
    return compile(source, filename, 'exec')
  File "./this.jinja2", line 499
    if 0: yield None
                   ^
IndentationError: too many levels of indentation
```

## Your Environment
* Python version: 2.7.6
* Jinja version: 2.7.2-2

ThiefMaster:
[...]

PR #2146, opened 2026-03-18
tests: lower test_elif_deep depth from 1000 to 100

### LINKED ISSUE #2145: `test_elif_deep` fails on s390x with `RecursionError`
`test_elif_deep` generates 999 nested elif branches, which exceeds CPython's `Py_C_RECURSION_LIMIT` of 800 on s390x (set lower due to the z/Architecture ABI's 160-byte-per-frame register save area). The flat elif codegen from #759 is unconditional, so the test depth can be lowered without reducing coverage.

Environment:

- Python version: 3.13
- Jinja version: latest (3.1.6)

amaanq:
The flat `elif` codegen (#759) is unconditional, so even a handful of branches exercises the codepath. On s390x, CPython sets `Py_C_RECURSION_LIMIT` to just 800 because the z/Architecture ABI requires a 160-byte register save area per call frame, roughly 10x what x86_64 needs, which makes 1000 nested branches overflow `compile()` with a `RecursionError`. Lowering to 100 still thoroughly exercises the flat elif codepath while staying within all platform limits.

- Fixes #2145

### DIFF
--- a/tests/test_core_tags.py
+++ b/tests/test_core_tags.py
@@ -309,11 +309,12 @@ class TestIfCondition:
         assert tmpl.render() == "..."
 
     def test_elif_deep(self, env):
-        elifs = "\n".join(f"{{% elif a == {i} %}}{i}" for i in range(1, 1000))
+        n = 100
+        elifs = "\n".join(f"{{% elif a == {i} %}}{i}" for i in range(1, n))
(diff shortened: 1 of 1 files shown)


amaanq:
@davidism friendly ping, any chance this can get looked at?

davidism:
Looks good, but not a lot of free time right now.

PR #2149, opened 2026-03-22
Add MAX_STRING_SIZE limit to sandbox string multiplication

kodareef5:
`SandboxedEnvironment` limits `range()` via `MAX_RANGE` to prevent DoS from large sequences, but does not limit string multiplication. A template expression like `{{ "A" * 10**9 }}` allocates 1GB of memory instantly in any application rendering untrusted templates in a sandbox.

This adds `MAX_STRING_SIZE` (default 1,000,000) and a `safe_mul` function that checks result size before performing string repetition. The `*` operator is added to `intercepted_binops` by default so the check applies automatically.

Unaffected operations:
- Normal arithmetic: `{{ 6 * 7 }}` → `42`
- Small strings: `{{ "ab" * 5 }}` → `ababababab`
- Lists: `{{ [1,2] * 3 }}` → `[1, 2, 1, 2, 1, 2]`
- Floats: `{{ 3.14 * 2 }}` → `6.28`

Blocked:
- `{{ "A" * 10000000 }}` → `OverflowError`

All 911 existing tests pass.

### DIFF
--- a/src/jinja2/sandbox.py
+++ b/src/jinja2/sandbox.py
@@ -24,6 +24,9 @@ F = t.TypeVar("F", bound=t.Callable[..., t.Any])
 #: maximum number of items a range may produce
 MAX_RANGE = 100000
 
+#: maximum size of a string that can be produced by the ``*`` operator
+MAX_STRING_SIZE = 1000000
+
 #: Unsafe function attributes.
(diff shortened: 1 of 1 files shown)


davidism:
Set resource limits in the os, docker container, etc. Trying to limit the resource use from within the process is an impossible surface to cover. 

Please review our security policy to learn how to responsibly report issues.

PR #2167, opened 2026-05-11
Limit GitHub Actions token permissions

xsourabhsharma:
This limits the default `GITHUB_TOKEN` permissions for CI workflows that only need to read repository contents.

- Add `contents: read` to the tests and pre-commit workflows.
- Add `contents: read` to the publish workflow's build job, while preserving the existing release and PyPI job permissions.

This follows the principle of least privilege for GitHub Actions and reduces the impact if an action or job step is compromised.

Verification:

- `uv run --locked --group pre-commit pre-commit run --files .github/workflows/tests.yaml .github/workflows/pre-commit.yaml .github/workflows/publish.yaml`
- `git diff --check`


ThiefMaster:
We already configured this on the GH org level, so the default permissions are read-only

xsourabhsharma:
okay

PR #2178, opened 2026-05-29
update documentation section on Whitespace control for correctness an…

uneasyalliance:
Fixing the documentation related to whitespace control. Clarified that lstrip_blocks and trim_blocks does not apply to expressions. Remove section indicating '+' before block end disables trim_blocks (actually an error).


uneasyalliance:
Closed with no comment? I don't get it.

Issue #2176, opened 2026-05-27
`indent` filter ignores `blank`=`False` when `first`=`True` on first line


abstract-official:
Currently, `indent` filter indents the first line no matter whether it's empty if `first` is `True`, even if `blank` is `False` (which is the default).

```python
from jinja2 import Environment
Environment().from_string("{% filter indent(4, first=True) %}{% endfilter %}").render()
```

Expected result is `""`, but actual result is `"    "`.

This is important when rendering something like this (with `lstrip_blocks` and `trim_blocks` on):
```html
<head>
  <style>
    /* some default styles, such as: */
    :root {
      color-scheme: light dark;
    }
    {% filter indent(4, first=True) %}
    {% block extra_style %}{% endblock %}
    {% endfilter %}
  </style>
</head>
```
which becomes:
```html
<head>
  <style>
    /* some default styles, such as: */
    :root {
      color-scheme: light dark;
    }
      </style>
</head>
```
if `extra_style` is not overwritten by subtemplates. (There's extra four spaces before `  </style>`.)
However, if subtemplates overwrites this block, then everything is properly indented.

Wrap all styles in a block and let subtemplates use `{{ super() }}`? Then subtemplates have to care about indentation in base templates, which breaks separation of concerns.

Or compromise with uglily indented things like:
```html
<head>
  <style>
{% filter indent(4, first=True) %}
/* some default styles, such as: */
:root {
  color-scheme: light dark;
}
{% block extra_style %}{% endblock %}
{% endfilter %}
  </style>
</head>
```

Or, hack with:
```html
<head>
  <style>
    /* some default styles, such as: */
    :root {
      color-scheme: light dark;
    {# Yes, begin the filter before common part ends. More difficulties when editing default styles. #}
    {% filter indent(4) %}
    }
    {% block extra_style %}{% endblock %}
    {% endfilter %}
  </style>
</head>
```

This is even more significant when rendering e.g. Python or YAML.

Environment:

- Python version: 3.13.13
- Jinja version: 3.1.6

devtechedge:
Taking a look at this. Plan is to make `indent(..., first=True)` still honor `blank=False` for an empty first line (including fully empty input), then add a regression test around the empty/`{% endfilter %}` case from the report.

noobiieez1:
Proposed fix on fork (tests for empty `first=True` / `blank=False` case):

https://github.com/noobiieez1/jinja/tree/fix/indent-blank-first-empty-line

[...]

PR #2181, opened 2026-06-02
Fix indent filter ignoring blank=False for empty first line

sapirbaruch:
Fixes #2176.

When `first=True` is passed to the `indent` filter, the code was
unconditionally prepending the indentation string to the output, without
checking whether `blank=False` should suppress it when the first line is
empty.

**Before:**

```python
>>> env.from_string('{% filter indent(4, first=true) %}{% endfilter %}').render()
'    '  # four spaces added to an empty string
```

**After:**

```python
>>> env.from_string('{% filter indent(4, first=true) %}{% endfilter %}').render()
''  # blank=False (default) suppresses indentation of the empty first line
```

The fix is a one-liner in `do_indent`: instead of `if first:`, check
`if first and (blank or rv.split("\n", 1)[0]):` so that an empty first
line is only indented when `blank=True` is explicitly requested.

A new test (`test_indent_first_blank`) covers the three cases:
1. Empty string with `first=True` → no indentation
2. Leading blank line with `first=True` → blank line left alone, rest indented
3. Empty string with `first=True, blank=True` → indentation applied

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,8 @@ Unreleased
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`1793`
 -   Use ``flit_core`` instead of ``setuptools`` as build backend.
+-   Fix ``indent`` filter not respecting ``blank=False`` for the first line
+    when ``first=True`` is set. :issue:`2176`
 
 
 Version 3.1.6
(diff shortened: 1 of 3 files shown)


sapirbaruch:
The PyPy failure is a pre-existing `trio` incompatibility that affects other open PRs too — not caused by this change. All other checks pass.

PR #2193, opened 2026-06-20
AI spam

### LINKED ISSUE #1755: [doc] Add notes to explain about associativity of, and precedence between operators and filters
The doc of jinja lacks of descriptions about the followings, if I'm not mistaken.

- associative properties of operators and filters
- precedence between operators and filters

Therefore, there are some cases users may not be able to expect how expressions will be evaluated beforehand, like the following examples, IMHO.

- 1 + 2 * 3
- 1 + "2" | int
- xs | length > 0  (where xs is a list, e.g. [1, 2])
- xs | length > 0 is true

To resolve this, I would like to suggest to enhance the doc (the section "Template Designer Documentation" should be suitable for this purpose, I think), to add some notes or a section to explain about associativity of, and precedence between operators and filters like the followings.

- Filters and operators are left associative, that is, grouped from left to right. So, for example,
  - {{ 1 + 2 - 3 }} will be evaluated as {{ (1 + 2) - 3 }}
  - {{ 3 ** 2 ** 3 }} will be evaluated as {{ (3 ** 2) ** 3 }} unlike python evaluates it as {{ 3 ** (2 ** 3) }}
  - {{ xs | list | length }} will be evaluated as {{ (xs | list) | length }}
- Filters have higher precedence than operators. So, for example,
  - {{ 1 + "2" | int }} will be evaluated as {{ 1 + ("2" | int) }} and return 3
  - {{ [3] + [2, 1] | sort }} will be evaluated to {{ [3] + ([2, 1] | sort) }} and return [3, 1, 2]
- Operators have precedence same as them in python and left associative. So, for example,
  - {{ 1 + 2 * 3 }} will be evaluated as {{ 1 + (2 * 3) }} and return 7

BTW, this is related to #379 and may close it.

MakenRosa:
﻿## Summary

Adds an operator precedence section to the template expressions documentation.

The section lists precedence from lowest to highest, calls out Jinja's left-associative `**` behavior, and includes examples for arithmetic, filters, and comparisons.

## Validation

- Rendered the documented examples with the local `src/` package using `PYTHONPATH=src`
- `git diff --check`

Fixes #1755

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -1472,6 +1472,40 @@ two categories:
     Get an attribute of an object.  (See :ref:`variables`)
 
 
+Operator Precedence
+~~~~~~~~~~~~~~~~~~~
+
+Expressions are grouped by operator precedence. Operators with higher
+precedence are evaluated before operators with lower precedence. Parentheses
+can always be used to group an expression explicitly.
+
+The following list shows precedence from lowest to highest:
+
+-   Inline if expressions: ``a if condition else b``
+-   ``or``
+-   ``and``
+-   ``not``
+-   Comparisons: ``==``, ``!=``, ``>``, ``>=``, ``<``, ``<=``, ``in``,
+    ``not in``
+-   Addition and subtraction: ``+``, ``-``
+-   String concatenation: ``~``
+-   Multiplication, division, floor division, and modulo: ``*``, ``/``,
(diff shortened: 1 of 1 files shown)


ThiefMaster:
seems AI-generated and full of unrelated (whitespace) changes...

PR #2209, opened 2026-07-04
AI junk

zayeem06:
The code generator formats the template name into the "no else section" message for a failing inline if-expression by writing self.position(node) straight into a double-quoted string literal in the generated module, so a name that contains a double quote breaks out of that literal. When such a template renders an inline if without an else branch and the condition is false, compiling the generated code raises a SyntaxError, or runs code injected through the name (a name of the form `a", <python>, "b` reaches the generated module verbatim). This is the same shape as the FromImport case that was already escaped for GHSA-gmj6-6f8f-6699; here I build the message and write it through repr so quotes and other characters are escaped consistently. Added a regression test in tests/test_compile.py using a quoted template name, which fails with a SyntaxError before the change.


zayeem06:
any update?

PR #2222, opened 2026-07-23
AI junk

MohammedAnasNathani:
## Summary
Fixes #2176.

When `indent` is used with `first=True` and the default `blank=False`, empty input previously produced a pure-whitespace result (e.g. four spaces) because `first` always forced a leading indent — even when the first line was empty.

That breaks real templates such as:

```jinja
{% filter indent(4, first=True) %}
{% block extra_style %}{% endblock %}
{% endfilter %}
```

when the block is empty (extra spaces before the following markup).

### Behavior
| Input | `first` | `blank` | Result |
|-------|---------|---------|--------|
| `""` | True | False | `""` (was `"    "`) |
| `""` | True | True | `"    "` |
| `"jinja"` | True | False | `"    jinja"` |
| multiline starting with blank line | True | False | unchanged (still indents the block as before) |

### Tests
- New regression cases for empty filter body / empty string
- Existing indent tests still pass

## Checklist
- [x] Tests added
- [x] Local `pytest tests/test_filters.py -k indent` green


ThiefMaster:
You used an LLM to generate this PR, didn't you?

MohammedAnasNathani:
Yes — I used an AI coding assistant while working on this, and I should have said so up front. Sorry about that.

I did verify the bug locally before/after:

```python
env = Environment()
# empty filter body with first=True
env.from_string("{% filter indent(4, first=True) %}{% endfilter %}").render()
# was "    ", expected ""
```

The intent of the change is narrow: `blank=False` should still apply when the first line is empty, so `first=True` does not force a pure-whitespace indent on empty input. Existing multiline indent tests were left green.

If you would prefer a human-only rewrite or a different approach (e.g. only special-casing fully empty input), I’m happy to rework this by hand. No hard feelings if you want to leave it closed.

ThiefMaster:
Yes, please see our [AI policy](https://palletsprojects.com/contributing/llm-ai). You will also learn much more when working on something yourself.

Issue #2118, opened 2025-10-01
Slice returns one extra item when slice count is a divisor of iterable length and fill_with not none


matiasilva:
The slice filter returns one extra item when the slice count can divide the iterable length completely and a fill_with value is used.

Example:

```
{% set a = [1, 2, 3, 4] %}
{{ a|slice(4, 'foo')|list }}
```

Output:

```[[1, 'foo'], [2, 'foo'], [3, 'foo'], [4, 'foo']]```

Expected output:

```[[1], [2], [3], [4]]```

Environment:

- Python version: 3.10.12
- Jinja version: 3.1.6

PR #2223, opened 2026-07-23
AI junk

MohammedAnasNathani:
## Summary
Fixes #2118.

`slice(n, fill_with)` incorrectly appended a filler to **every** column when the sequence length was evenly divisible by `n`:

```jinja
{{ [1, 2, 3, 4]|slice(4, 'foo')|list }}
```

- **Before:** `[[1, 'foo'], [2, 'foo'], [3, 'foo'], [4, 'foo']]`
- **After:** `[[1], [2], [3], [4]]`

### Root cause
Padding used `slice_number >= slices_with_extra`. When the remainder is `0`, that is true for **all** slices, so every full column got one extra filler.

### Fix
Pad only when `len(tmp) < target`, where `target` is the tallest column height. Uneven cases still pad shorter columns as before.

### Tests
- Regression for even divisor + `fill_with`
- Existing slice cases still pass


MohammedAnasNathani:
Note on process: I used an AI coding assistant for drafting/tests, and reviewed the root cause myself.

The bug is mechanical: when `length % slices == 0`, the old condition `slice_number >= slices_with_extra` is true for every column, so `fill_with` was appended to already-full slices. The fix pads only when `len(tmp) < target`. Happy to adjust the wording or tests if maintainers want a different style.

PR #2225, opened 2026-07-26
AI junk

deprrous:
This fixes two small docs mistakes I hit while checking the current examples.

The Mako-style environment example used `commend_end_string`, which isn't a real Jinja setting. It should be `comment_end_string`.

The gettext docs also pointed readers at `env.install_translations`, but the current API is `env.install_gettext_translations`.

I only changed the docs text.

Validation:
- `uv run pytest -q tests/test_ext.py`


davidism:

PR #2227, opened 2026-07-29
AI junk

cyforkk:
## Description

The indent filter with first=True was indenting the first line even
when it was empty and blank=False. This caused issues when using
the filter with empty blocks.

## Changes
- Added a check to skip indenting the first line when it's empty and blank=False

Fixes pallets/jinja#2176


davidism:

PR #2228, opened 2026-07-29
AI junk

tomatotomata:
## Summary

- respect blank=False when first=True by leaving an empty first line unindented
- preserve the existing behavior for non-empty first lines and for blank=True
- add regression coverage for an empty value and leading blank lines

## Validation

- python -m pytest tests/test_filters.py -q (132 passed)
- python -m pytest tests/test_filters.py -k indent -q (3 passed)
- python -m ruff check src/jinja2/filters.py tests/test_filters.py
- python -m compileall -q src/jinja2/filters.py tests/test_filters.py
- git diff --check

Closes #2176


davidism:

PR #2229, opened 2026-08-02
AI junk

fyrsta7:
## Summary

This simplifies `_PassArg.from_obj` to read `jinja_pass_arg` with one `getattr(..., None)` call instead of checking with `hasattr` and then reading the attribute again.

## Why this can improve performance

`hasattr` performs an attribute lookup internally. The previous code could therefore perform two lookups for objects that define `jinja_pass_arg`. A single `getattr` keeps the same `None` fallback while avoiding the duplicate lookup.

## Validation

- `git diff --check`
- `python3 -m compileall -q src/jinja2/utils.py`
- Behavior check for objects with and without `jinja_pass_arg`

## Benchmark

Current machine, CPython microbenchmark, 9 samples, median:

| Benchmark | Base | This change | Delta |
|---|---:|---:|---:|
| Repeated `_PassArg.from_obj` lookups | 49,928.194 ns/op | 48,589.975 ns/op | 2.68% faster |


davidism:

PR #2230, opened 2026-08-05
Fixes #2176

daleselaji-dev:
Fixes #2176

The indent filter currently applies first=True to an empty first line even when blank=False, producing spaces for empty input and unwanted whitespace before later content. Preserve the original first-line value and only apply first-line indentation when that line is non-empty, unless blank=True explicitly requests blank-line indentation.

The behavior is documented in the 3.2 changelog and filter docstring.

Tests: PYTHONPATH=src python -m pytest tests/test_filters.py -q (132 passed); PYTHONPATH=src python -m ruff check src/jinja2/filters.py tests/test_filters.py; python -m compileall -q src/jinja2/filters.py tests/test_filters.py.


davidism:

PR #2236, opened 2026-08-13
AI junk

rawsun007:
fixes #2176

`do_indent` ends with an unconditional prepend:

```python
if first:
    rv = indention + rv
```

So `first=True` indents the first line even when it is empty, right after `blank=False` finished skipping every other empty line. The result is trailing whitespace on a line with nothing on it.

```pycon
>>> Environment().from_string("{% filter indent(4, first=True) %}{% endfilter %}").render()
'    '
```

In the reporter's template that shows up as an empty `{% block %}` inside `{% filter indent(4, first=True) %}` leaving four spaces that shift the following line, so the base template only renders correctly when a child happens to override the block.

## Why `blank` wins

The two arguments look like they conflict, but they answer different questions:

- `first` — is the first line special? (`True`: no, treat it like the rest.)
- `blank` — are empty lines indented? (`False`: no.)

An empty first line is both, so once `first=True` says "not special", it falls under the ordinary empty-line rule and `blank` decides. Reading it the other way makes `first=True` mean "treat the first line like the others, except give it the indentation the others would not get".

```python
if first and (blank or rv.partition(newline)[0]):
    rv = indention + rv
```

## Heads-up: this changes an assertion

`_test_indent_multiline_template` feeds text beginning with a newline and asserts:

```python
t = env.from_string("{{ foo|indent(2, true, false) }}")
assert t.render(foo=text) == '  \n  foo bar\n  "baz"\n'
```

That leading `'  '` is the bug, so the assertion is updated rather than preserved. Flagging it explicitly because it means someone could be relying on the old output: anyone passing `first=True, blank=False` with a leading newline loses two spaces on line one. Adding a separate opt-in flag seemed worse than fixing `blank` to mean what it says, but that is a maintainer call and I am happy to redo it that way.

`blank=True` is untouched: an empty first line is still indented there, which is what opting into blank-line indentation should do.

## Tests

New `test_indent_blank_first_line_not_indented_without_blank` covers the four cases that pin the rule:

| Input | Args | Expected |
| --- | --- | --- |
| `""` (empty filter body) | `first=True` | `""` |
| `"\nsecond"` | `first=True` | `"\n  second"` |
| `"\nsecond"` | `first=True, blank=True` | `"  \n  second"` |
| `"first\nsecond"` | `first=True` | `"  first\n  second"` |

The third and fourth are there so the fix cannot be over-applied: it would be easy to drop the leading indent for every `first=True` case, or to break `blank=True`, and both would pass with only the first two.

```
$ pytest -q
912 passed in 0.76s

$ ruff check src tests    All checks passed!
$ ruff format --check     49 files already formatted
```

Docstring for `blank` now states the interaction, and there is a `CHANGES.rst` entry under 3.2.0 noting the behaviour change.


davidism:

PR #2237, opened 2026-08-13
Do not report names assigned by every if branch as undeclared

### LINKED ISSUE #1253: What's wrong with this Jinja template when using {% elif %}

### LINKED ISSUE #2069: meta.find_undeclared_variables returning variables set in if blocks as of 3.1.5
I've observed a change in behaviour between 3.1.4 and 3.1.5 where `meta.find_undeclared_variables` is now returning variables that are `set` in all branches of an `if` statement.

For example with the following:

```python
from jinja2 import Environment, meta

env = Environment()

template = """\
{% if control == 'something' %}
    {% set output = 1 %}
{% elif control == 'something else' %}
    {% set output = 2 %}
{% else %}
    {% set output = 3 %}
{% endif %}
{{ output }}
"""
result = meta.find_undeclared_variables(env.parse(template))
assert result == {'control'}
```

With jinja2 version 3.1.4, this runs without error. With jinja2 version 3.1.5, this errors becasue `result` contains both "control" and "output". I believe the 3.1.4 behaviour is correct in this instance.

Reading the changelog, I suspect that the fix for #1253 (PR #1665) looks likely to be the cause of this change.

Environment:

- Python version: 3.10.12
- Jinja version: 3.1.5

rulfox:
`meta.find_undeclared_variables` reports a name that is `{% set %}` in every branch of an `{% if %}` chain ending in `{% else %}`, even though such a name is always defined afterwards and is never read from the context:

```python
env.parse("{% if a %}{% set x = 1 %}{% else %}{% set x = 2 %}{% endif %}{{ x }}")
# reports {"a", "x"}; "x" cannot come from the context
```

### Why it regressed

`Symbols.branch_update` used to skip the context load for a name stored by every branch. #1665 removed that optimisation to fix #1253, because it also skipped the load when a branch *read* the name before assigning it.

### This change

Restores the optimisation behind two guards:

1. Every branch must introduce the name as a fresh local, that is, its load is `undefined` rather than `resolve`/`alias`. This is the case #1253 needed.
2. The name must not be read inside a scope `FrameSymbolVisitor` does not enter. Those reads are invisible to the frame's symbols but still read the frame's variables, so `FrameSymbolVisitor` now records them and `visit_If` consults them. Every field the compiler visits with a child frame is tracked: `For.body/else_/test`, `Macro.body/defaults`, `CallBlock.body/defaults`, `FilterBlock.body/filter`, `With.body`, `AssignBlock.body/filter`, `Scope.body`, `Block.body`, `OverlayScope.context/body`.

`visit_If` also tracks each `{% elif %}` body as its own branch, so plain `{% if %}`/`{% else %}` is recognised too. That case was reported in #2069 as failing on 3.1.4 as well.

The scan of nested scopes is lazy, so a frame with no exhaustive `{% if %}` chain does no extra work.

### Verification

- The three snippets in #2069 (with `elif`, without `elif`, and the no-`else` control case) all behave as the reporter expected.
- Differential run over ~1260 generated templates x 8 contexts covering `for`, `for recursive`, `macro`, `call`, `filter`, `with`, `set`/`endset`, `block`, `extends`, `include`, `import`, namespaces, tuple targets and nested chains: no rendered output changes, only the intended reduction in `find_undeclared_variables` results.
- Compile time measured at roughly +2% on a 200-unit page template.

### Tests

- `tests/test_api.py` — parametrized `find_undeclared_variables` cases, including the negative ones: no `else`, a branch that does not assign, read-before-assign, read in an `elif` test, read in a nested scope.
- `tests/test_regression.py` — rendering tests beside the existing #1253 test, covering a read in each child-frame field.
- `tests/test_idtracking.py` — symbol-level assertions for an exhaustive chain and for `OverlayScope.context`.

Full suite: 936 passed. `ruff` clean; `mypy` reports the same pre-existing errors as `main`.

fixes #2069


davidism:

PR #2238, opened 2026-08-15
AI junk

Swastikbhat-lab:
Fixes #2118.

## Bug

When the slice count divides the iterable length evenly, the `slice` filter still pads every chunk with `fill_with`:

```
{{ [1, 2, 3, 4]|slice(4, 'foo')|list }}
```

Actual: `[[1, 'foo'], [2, 'foo'], [3, 'foo'], [4, 'foo']]`
Expected: `[[1], [2], [3], [4]]`

## Cause

In `sync_do_slice`, the padding condition is `slice_number >= slices_with_extra`. When `length % slices == 0`, `slices_with_extra` is 0, so the condition is true for every chunk and each one gets `fill_with` appended — even though no chunk is short.

## Fix

Only pad when there are short chunks to pad, i.e. when `slices_with_extra` is non-zero. The non-divisor behavior (padding the chunks after the extras, e.g. `range(10)|slice(3, 'X')` → `[[0, 1, 2, 3], [4, 5, 6, 'X'], [7, 8, 9, 'X']]`) is unchanged.

## Tests

Added a regression test for the divisor case. Full suite:

```
912 passed in 2.49s
```


davidism:

PR #2239, opened 2026-08-22
AI junk

Mehaksaluja:
Fixes #2176

## Summary

The `indent` filter in the current implementation mistakenly indents the first line irrespective of whether it's empty when `first=True`, even if `blank=False`. The filter should check if the first line is blank and skip indenting it if `blank=False`. This behavior is causing the first line to be indented unnecessarily in specific use cases.

## Validation

- New test passes with the fix: True
- New test fails without the fix: True
- Existing suite still passes: True
- Linter passes: True

## Pre-submit review

The proposed patch correctly modifies the `indent` filter logic to avoid indenting the first line if it's empty and `blank=False`. The new tests provided sufficiently validate that the changes address the original issue and would fail on the unmodified code.

---

This pull request was opened by BOUNTY, an AI-assisted contributor. A human reviewed the diff before it was submitted.


davidism:

PR #2242, opened 2026-08-24
fix: prevent code injection via unsafe __repr__ in Const and TemplateData nodes

truongsontung:
Fix code injection vulnerability in compiler.py where unsafe repr() output was injected into generated Python code without validation.


davidism:
https://palletsprojects.com/security

Issue #2109, opened 2025-07-22
poor performance when parsing unclosed string with many escape characters


wsparks-vc:
The lexer has poor performance on unclosed strings with many escape characters. It seems to be due to the `string_re` regular expression. It should probably be possible to make this run in linear or near-linear time instead by adjusting the regex or lexer.

```python
import time


def timer(func):
    def wrapper(*args, **kwargs):
        start_time = time.time()
        result = func(*args, **kwargs)
        end_time = time.time()
        elapsed_time = end_time - start_time
        print(f"Function '{func.__name__}' executed in {elapsed_time:.4f}s")
        return result

    return wrapper


def sizer(func):
    def wrapper(*args, **kwargs):
        result = func(*args, **kwargs)
        print(f"Payload length: {args[0]}\nPayload size: {len(result.encode()) / (1024 ** 2)} MB")
        return result

    return wrapper


from jinja2 import Environment
from jinja2.lexer import get_lexer


@sizer
def create_payload(char_length: int):
    # The slow pattern: r"('([^'\\]*(?:\\.[^'\\]*)*)'" r'|"([^"\\]*(?:\\.[^"\\]*)*)")'
    # This is slow on a string that starts with a quote, has many escape sequences,
    # and ends with characters that will cause backtracking
    payload = "'" + ("\\a" + "b" * char_length) * char_length + "c"
    return payload


@timer
def runner(char_length: int):
    try:
        # Create the payload
        payload = create_payload(char_length)

        # Get the lexer from jinja2
        env = Environment()
        lexer = get_lexer(env)

        # Tokenize the payload - this is slow
        tokens = list(lexer.tokenize(payload))

        return tokens
    except Exception as e:
        print(f"Exception occurred: {e}")
        return None


if __name__ == '__main__':
    runner(100)
    runner(1000)
    runner(2000)
    runner(4000)
    runner(8000)
    runne(10000)
```

Environment:

- Python version: 3.13.5
- Jinja version: 3.1.6

ThiefMaster:
Especially in shared environments, proper rlimits, timeouts, etc. should be set.

Also, unless you use the sandboxed environment - templates must never come from untrusted sources anyway (otherwise you can do RCE via malicious templates).

In any case, I think we'd be open to a PR to fix this.

davidism:
I've updated this to directly talk about the performance issue rather than attacks and DoS, as described in our security policy.

davidism:
[...]

PR #2245, opened 2026-08-28
Reduce lexer backtracking on unclosed strings with many escapes

thejdubb02:
Fixes the performance issue reported in #2109: the lexer does a lot of redundant work on an unclosed string that contains many escape sequences.

Root cause:
`string_re` matches the body of a quoted string with the unrolled pattern `[^'\\]*(?:\\.[^'\\]*)*` and then requires the closing quote. When the closing quote is missing, the regex engine backtracks through the already matched body, retrying at each position for a quote that is not there. For the input in the issue (a quote, then many `\a` plus a run of `b`, repeated, with no closing quote), that backtracking is expensive.

Fix:
Match the body atomically so the engine does not backtrack into it. Python only gained atomic groups and possessive quantifiers in 3.11, and Jinja supports 3.10, so this uses the lookahead plus backreference form `(?=(...))\N`, which is atomic on 3.10 and later. The lexer only consumes the whole match (`m.group()`), not the inner capture groups, so restructuring the internal groups is safe.

Verification:
- The full test suite passes on Python 3.10, plus a new regression test that lexes the issue's input and confirms it raises `TemplateSyntaxError`.
- The old and new patterns produce identical matches on valid and edge-case strings: empty strings, escaped quotes, escaped backslashes, newlines under `re.S`, adjacent strings, a trailing backslash, and unterminated strings (both reject).
- On the issue's input the new pattern is about 5 times faster. On ordinary valid strings the extra lookahead pass is negligible (about a microsecond per match over 300k matches).

Per the maintainers' note in the issue, this is a performance fix, not a security report, so no CVE is requested.


davidism:

PR #2246, opened 2026-08-28
AI junk

afonsojanu:
Fixes #2176.

`do_indent()` indented the first line unconditionally whenever `first=True`, ignoring `blank` for that one line even though every other line already respects it. In practice this only bit when the whole input collapsed to that single blank first line — e.g. an empty `{% filter indent(first=True) %}{% endfilter %}` block — turning "nothing" into a few bytes of bare trailing whitespace.

A blank first line followed by real content is left as-is: the existing multiline test (`indent(2, true, false)` over text starting with a blank line) already expects that line to be indented, matching `first`'s documented "don't skip indenting the first line" contract. So the fix is scoped to the case where there's no other line at all — `first and (blank or rv or lines)` instead of an unconditional `first`.

Added a regression test for both the filter-block and plain-filter forms of the empty-input case, and a changelog entry. Full test suite passes locally (912 passed).


davidism:

PR #2247, opened 2026-08-28
AI junk

### LINKED ISSUE #2165: `[{}] | map(attribute = "foo", default = None)` still fails
According to the documentation, when using `attribute` in `map` instead of a filter, if the value of `attribute` is not a key of an input element, `default` can be specified to supply a value instead. In the past, that would fail if that value could be cast to `False` (https://github.com/pallets/jinja/issues/1331). I have no idea if that is some sort of regression, or if `None` was never tested against.

This seems indeed to work for most of those values, except for `None`. I have no jinja executable and no idea how to invoke jinja, so here is an example with ansible, with an error message being identical as if `default` had not been specified at all, as seen in the last meaningful line:

```
./venv/latest/bin/ansible -m debug -a 'msg={{[{}]|map(attribute="foo",default=none)}}' localhost
[ERROR]: Task failed: Finalization of task args for 'ansible.builtin.debug' failed: Error while resolving value for 'msg': object of type 'dict' has no attribute 'foo'
 
Task failed.
Origin: <adhoc 'debug' task>
 
{'action': 'debug', 'args': {'msg': '{{[{}]|map(attribute="foo",default=none)}}'}, 'timeout': 0, 'async_val': 0, [...]
 
<<< caused by >>>
 
Finalization of task args for 'ansible.builtin.debug' failed.
Origin: <CLI option '-m'>
 
debug
 
<<< caused by >>>
 
Error while resolving value for 'msg': object of type 'dict' has no attribute 'foo'
Origin: <CLI option '-a'>
 
{{[{}]|map(attribute="foo",default=none)}}
 
localhost | FAILED! => {
    "changed": false,
    "msg": "Task failed: Finalization of task args for 'ansible.builtin.debug' failed: Error while resolving value for 'msg': object of type 'dict' has no attribute 'foo'"
}
```
Environment:

- Python version: 3.13.5
- Jinja version: 3.1.6

I believe this results from how `prepare_map()` handles `kwargs`.

afonsojanu:
make_attrgetter() only substituted an Undefined attribute with `default` when `default is not None`, and prepare_map() popped `default` from kwargs with a plain `None` fallback -- so map(attribute=..., default=None) was indistinguishable from not passing default at all, and always left the attribute Undefined instead of substituting None.

Introduce a private _no_default sentinel and thread it through make_attrgetter()'s own default parameter, prepare_map()'s kwargs.pop, and groupby's default parameter (which shared the exact same bug via the same helper, and would otherwise have regressed once make_attrgetter's own default changed).

Fixes #2165


davidism:

PR #2254, opened 2026-08-31
[rejected AI] indent filter ignoring blank=False when first=True

sanyamk23:
Fixes #2176

When `blank=False` (the default), the `indent` filter should not indent empty/blank lines. However, when `first=True` and the input is an empty string, the filter was incorrectly indenting the empty result.

This caused issues when rendering Jinja templates with empty blocks -- the closing tag would get extra indentation even when the block content was empty.

**Before:** `{% filter indent(4, first=True) %}{% endfilter %}` returned `'    '` instead of `''`
**After:** Returns `''` as expected

### DIFF
--- a/src/jinja2/filters.py
+++ b/src/jinja2/filters.py
@@ -860,7 +860,7 @@ def do_indent(
                 indention + line if line else line for line in lines
             )
 
-    if first:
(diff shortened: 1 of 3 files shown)


davidism:

PR #2260, opened 2026-09-01
[rejected AI] do not append fill_with in slice when slice count evenly divides iterable

pylrn:
Fixes #2118

### Problem
When `slices` evenly divides `len(seq)` (e.g. `[1, 2, 3, 4]|slice(4, 'foo')`), `slices_with_extra = length % slices` is `0`.
Because the check was `slice_number >= slices_with_extra`, every slice satisfied `slice_number >= 0` and had `fill_with` appended, adding an extra unnecessary element to every sub-list even though all slices were already full and of equal length.

### Fix
- Updated the `fill_with` check in `sync_do_slice` in `src/jinja2/filters.py` to require `slices_with_extra > 0` in addition to `slice_number >= slices_with_extra`.
- Added test assertions to `tests/test_filters.py` and `tests/test_async_filters.py`.


davidism:

PR #2261, opened 2026-09-04
[rejected AI] Do not append fill_with in slice filter when length is divisible

kishansaaai:
.


davidism:

PR #2262, opened 2026-09-04
[rejected AI] preserve default=None in map filter (#2165)

### LINKED ISSUE #2165: `[{}] | map(attribute = "foo", default = None)` still fails

eminsk:
When using `map(attribute="...", default=None)`, `make_attrgetter` previously checked `if default is not None:`, treating `default=None` the same as no default provided. Missing attributes therefore remained `Undefined` instead of defaulting to `None`, causing `UndefinedError` in strict environments (such as Ansible or when using `StrictUndefined`).

This PR:
- Uses the `missing` singleton from `jinja2.utils` as the default sentinel in `make_attrgetter` and `prepare_map`.
- Ensures explicitly passed `default=None` evaluates to `None`.
- Preserves backward compatibility in `groupby` (avoiding the PyPy sorting regression in stale PR #2175).
- Adds unit test coverage in `tests/test_filters.py` verifying `default=None` under standard and `StrictUndefined` environments.

Fixes #2165 (Supersedes #2175)


davidism:

PR #2265, opened 2026-09-05
[rejected AI] Preserve empty strings in indent(first=True)

Nikhi00718:
Fixes #2176.

The `indent` filter now leaves an empty string unchanged when `first=True`, while preserving existing indentation behavior for non-empty and blank lines. Added a regression test covering an empty filtered block.

Tests: `python -m pytest tests/test_filters.py -q` (132 passed).


davidism:

PR #2266, opened 2026-09-05
[rejected AI] tojson: escape U+2028 and U+2029 (JavaScript line/paragraph separators)

### LINKED ISSUE #1882: Escape newlines for tojson filter as Django
`tojson` filter escapes  ``<``,  ``>``, ``&`` and ``'``. To avoid javascript injection in HTML `<script>` tags and `data-*` attributes, Django and other (e.g. Odoo) seems escaping also `\u2028` and `\u2029` (treated as newlines by some javascript engines, which may allow an attacker to begin a new javascript instruction).
More info at [https://code.djangoproject.com/ticket/17419#comment:27](https://code.djangoproject.com/ticket/17419#comment:27)
See [issue and commit in djangoproject](https://github.com/django/django/pull/9235).

Nowadays symbols are some more I think [https://github.com/jonashaag/django/blob/master/django/utils/html.py#L54](https://github.com/jonashaag/django/blob/master/django/utils/html.py#L54):
```
_js_escapes = {
    ord('\\'): '\\u005C',
    ord('\''): '\\u0027',
    ord('"'): '\\u0022',
    ord('>'): '\\u003E',
    ord('<'): '\\u003C',
    ord('&'): '\\u0026',
    ord('='): '\\u003D',
    ord('-'): '\\u002D',
    ord(';'): '\\u003B',
    ord('\u2028'): '\\u2028',
    ord('\u2029'): '\\u2029'
}
```

I saw only [`htmlsafe_json_dumps`](https://github.com/pallets/jinja/blob/85e5ad85d10eaf119dc0f17f245118c9f78bb864/src/jinja2/utils.py#L628) without knowing all Jinja codebase, however: sorry if already addressed or evaluated issue.

Thanks,
Pp

IMGillusion:
## Problem

`htmlsafe_json_dumps` (the `|tojson` filter) escapes `<`, `>`, `&` and `'` so its output is safe to embed in HTML and `<script>` tags. It does **not** escape the Unicode line separator `U+2028` and paragraph separator `U+2029`.

JavaScript treats those two characters as line terminators, exactly like `\n`. When `json.dumps` is told to emit raw Unicode (`ensure_ascii=False`, a very common config), the separators are emitted unescaped, so a crafted string can break out of a `<script>` block:

```python
env.policies['json.dumps_kwargs'] = {'ensure_ascii': False}
template = "{{ x|tojson }}"
# x = '""; alert(1); //\u2028document.location = \"http://evil\"'
```

## Fix

Escape `U+2028` / `U+2029` to `\\u2028` / `\\u2029`, matching what Django's `json_script` filter does. On the default path (`ensure_ascii=True`) `json.dumps` already escapes them as non-ASCII, so the new `.replace()` calls are harmless no-ops there.

Closes #1882


davidism:

PR #2268, opened 2026-09-09
[Rejected AI junk] indent: honor blank=False for empty first line when first=True

devtechedge:
## Summary

`indent(..., first=True)` was always prefixing the indent width, even when the first line was empty and `blank` was left at its default `False`. That made empty input (and an empty first line) render as spaces instead of staying empty.

Example from #2176:

```python
Environment().from_string("{% filter indent(4, first=True) %}{% endfilter %}").render()
# was "    ", now ""
```

`first=True` still indents a non-empty first line. With `blank=True`, an empty first line is still indented.

## Changes

- Only prepend indent when `first=True` if `blank=True` or the first line is non-empty
- Regression coverage for empty filter body, empty string, and multiline empty-first-line cases
- `CHANGES.rst` + `versionchanged` note for 3.2

fixes #2176

## Test plan

- [x] `pytest tests/test_filters.py -k indent` (4 passed)
- [x] full `tests/test_filters.py` (133 passed)


davidism:

PR #2270, opened 2026-09-10
[rejected AI] Fix indent filter handling of empty first lines

LouisDeconinck:
Fixes #2176

The `indent` filter currently prepends the indentation string when `first=True`, even if the first line is empty and `blank=False`. This produces whitespace for an empty filtered block.

Use the first line's content when deciding whether `first` should indent, while preserving `blank=True` behavior. The existing filter tests now cover the regression.

Checks:
- `git diff --check`
- Four direct focused render assertions (the repository test suite could not run because Trio is not installed in the environment).

The change is limited to the filter, its regression test, and `CHANGES.rst`.


ThiefMaster:

PR #2272, opened 2026-09-10
[rejected AI] do not append fill_with in slice filter when slice count divides length

GruffElixir:
When the iterable length is evenly divisible by \slices\ (i.e. \slices_with_extra == 0\), all slices already have equal length (\items_per_slice\). Previously, \slice_number >= slices_with_extra\ evaluated to \True\ for every slice when \slices_with_extra\ was 0, which caused \ill_with\ to be erroneously appended to every slice instead of only filling missing values on shorter slices.

This change adds \slices_with_extra > 0\ to the condition in \sync_do_slice\ so that \ill_with\ is only appended when there are unequal slices that need padding.

fixes #2118


davidism:

